PR #1, opened 2018-04-03
Adding an initial protobuf v3 template

fkautz:
Signed-off-by: Frederick F. Kautz IV <fkautz@alumni.cmu.edu>

An initial PR to initiate discussion on what should go into a network service mesh.

A few notes:

* We should determine whether grpc's error mechanism will be sufficient, or whether we need to separate mesh errors from transport errors by adding in error handling explicitly into the proto definitions.

* We will need to work out how expressive we want service discovery queries to be.

* grpc is bi-directional. This initial commit deals only with the mesh itself. We should also set up rpc services that may be present on the service requester and service provider.


hagbard5235:
A few other notes :)

* We should consider what kind of metadata goes with a session request.  I suspect we could associate labels with sessions, and in the future use those for the Network Service Mesh equivalent of [Route Rules](https://istio.io/docs/concepts/traffic-management/rules-configuration.html#route-rules).

fkautz:
I explicitly left the labels and metadata off, but also think that they
explicitly belong in. Just, in what form? I was thinking of
map[string]string. Also, do we want to append some required labels? Or have
reserved labels that can only be set by certain actors?

Let's start with the most basic, and iterate.

hagbard5235:
@fkautz Totally agree let's start basic and go from there :)

Question: How do you see the ConcealChannelRequest working?  What would you see as a lifecycle here (you've been a bit more complete than I was in my proposal :) ).

Ed

fkautz:
My view on Conceal was to have a way to say this channel is going away, but without forcing the channel to be closed on existing connections. We could use Cordon or Drain as well. I want to get the terminology here right, so if you have a better word for this, definitely add it in.

hagbard5235:
Ah, good idea on the Conceal thing.

Would it be fair to describe the lifecycle of a session sort of like this:

* Create (Pod -> Network Service Manager (NSM) )
* Conceal/Cordon/Drain (Pod -> NSM)
* Destroy (Pod -> NSM)
 
Do we need a way for the NSM to indicate that a Session is being torn down or redirected?

fkautz:
I added some labels to the requests. They are expressed as a string->string map at this point.

fkautz:
I added a string id that is returned in each request that generates a new resource. I chose string for now to facilitate human readable representations of a given id. 

Simultaneously, I think uint64 is insufficient due to the birthday paradox regarding the likeliness of a collision while selecting random numbers.

fkautz:
@hagbard5235 would it make sense to return labels assigned? E.g.

```
message PublishServiceRequest {
    map<string, string> labels = 1;
}
message PublishServiceResponse {
    string service_id = 1;
    **map<String, string> labels = 2;**
}
```

I'm thinking that it is possible that the SDN may assign additional labels and it may be useful to return the final resolved set.

fkautz:
Errors are set as a tuple: (code, errormsg)

We are able to set our own status codes and error messages. These codes do not appear to be enumerable in the specification itself, so we will need to be consistent with our implementations.

There are a default set of status codes you can read about here:

https://github.com/grpc/grpc-go/blob/master/codes/codes.go

My recommendation is we define our own set of error codes to supplement the default codes. We may need to add machine readable error messages if more context is required.

PR #3, opened 2018-04-17
Start hacking on some go for NSM

hagbard5235:


deviceplugin/deviceplugin.go
```diff
@@ -0,0 +1,144 @@
+// Package deviceplugin provides a basic implementation of a Kubernetes DevicePlugin without content.
+// It is intended to be used to build other DevicePlugins.
+package deviceplugin
+
+import (
+	"log"
+	"net"
+	"os"
+	//      "strings"
```
fkautz: Best to just remove these, let your favorite editor re-add as necessary.

deviceplugin/deviceplugin.go
```diff
@@ -0,0 +1,144 @@
...
+	"golang.org/x/net/context"
+	"google.golang.org/grpc"
+	pluginapi "k8s.io/kubernetes/pkg/kubelet/apis/deviceplugin/v1beta1"
+	"path"
+	"time"
+)
+
+type DevicePlugin struct {
```
fkautz: It is better to expose an interface and create a struct to implement.

e.g.

```go
type DevicePlugin interface {
    Foo() uint64
}

type devicePlugin struct {}

func (plugin *devicePlugin) Foo() uint64 {
    return 0
}

func NewDevicePlugin() DevicePlugin {
    return &devicePlugin{}
}
```
This allows us to easily mock the interface for testing and allows our code to be generic and work with other implementations that implement the interface.

deviceplugin/deviceplugin.go
```diff
@@ -0,0 +1,144 @@
...
+		unixSocketPath,
+		grpc.WithInsecure(), grpc.WithBlock(),
+		grpc.WithDialer(func(addr string, timeout time.Duration) (net.Conn, error) {
+			return net.DialTimeout("unix", addr, timeout)
+		}),
+	)
+}
+
+func (d *DevicePlugin) cleanup() error {
+	if err := os.Remove(d.socket); err != nil && !os.IsNotExist(err) {
```
fkautz: Is there any reason to suppress other errors?

deviceplugin/deviceplugin.go
```diff
@@ -0,0 +1,144 @@
...
+	go d.server.Serve(sock)
+
+	ctx := context.TODO()
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
+	conn, err := dial(ctx, d.socket)
+	if err != nil {
+		return err
+	}
+	conn.Close()
```
fkautz: We should ```defer conn.Close()``` so that we can defensively close the connection in case something else added in the future returns an error.

deviceplugin/deviceplugin.go
```diff
@@ -0,0 +1,144 @@
...
+
+func (d *DevicePlugin) Register(kubeletEndpoint, resourceName string) error {
+	ctx := context.TODO()
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
+	conn, err := dial(ctx, kubeletEndpoint)
+	if err != nil {
+		return err
+	}
+	defer conn.Close()
```
fkautz: Defer is a stack, so we should make sure we defer conn.Close() before we defer cancel()

deviceplugin/deviceplugin.go
```diff
@@ -0,0 +1,144 @@
...
+	client := pluginapi.NewRegistrationClient(conn)
+	request := &pluginapi.RegisterRequest{
+		Version:      pluginapi.Version,
+		Endpoint:     path.Base(d.socket),
+		ResourceName: resourceName,
+	}
+	_, err = client.Register(context.Background(), request)
```
fkautz: Do we want to add a timeout to this context?

nsm/main.go
```diff
@@ -0,0 +1,37 @@
...
+func main() {
+	log.Println("Loading NSM")
+	log.Println("Starting FS watcher.")
+	log.Println(pluginapi.DevicePluginPath)
+	watcher, err := newFSWatcher(pluginapi.DevicePluginPath)
+	if err != nil {
+		log.Println("Failed to created FS watcher.")
+		os.Exit(1)
+	}
+	defer watcher.Close()
```
fkautz: Does newFSWatcher block? If not, this function will exit, close the connection, and kill the application.

Take a look at https://golang.org/pkg/sync/#WaitGroup

```
var wg sync.WaitGroup
wg.Add(1)
go run(wg)
wg.Wait()
```

run calls ```wg.Done()``` when it is complete.

nsm/server.go
```diff
@@ -0,0 +1,140 @@
...
+	"path"
+	"time"
+)
+
+const (
+	resourceName = "nsm.ligato.io"
+	serverSock   = pluginapi.DevicePluginPath + "nsm.sock"
+)
+
+type NetworkServiceManagerDevicePlugin struct {
```
fkautz: This should probably be an interface

nsm/server.go
```diff
@@ -0,0 +1,140 @@
...
+	nsm.server = grpc.NewServer([]grpc.ServerOption{}...)
+	pluginapi.RegisterDevicePluginServer(nsm.server, nsm)
+
+	go nsm.server.Serve(sock)
+
+	conn, err := dial(context.TODO(), nsm.socket, 5*time.Second)
+	if err != nil {
+		return err
+	}
+	conn.Close()
```
fkautz: Add defer

nsmdp/nsmdp.go
```diff
@@ -0,0 +1,48 @@
+package nsmdp
+
+import (
+	"github.com/ligato/networkservicemesh/deviceplugin"
+	pluginapi "k8s.io/kubernetes/pkg/kubelet/apis/deviceplugin/v1beta1"
+)
+
+type NSMDevicePlugin struct {
```
fkautz: Should probably be an interface

nsmdp/nsmdp.go
```diff
@@ -0,0 +1,48 @@
...
+	*deviceplugin.DevicePlugin
+}
+
+type NSMDevice struct {
```
fkautz: Same as above, interface

fkautz:
I didn't look at the structure in this PR. Also, I noticed much of the code is the same between different packages. We should probably find a way to abstract some of this code into a consumable package.

deviceplugin/deviceplugin.go
hagbard5235: So if I were actually doing an interface, definitely.  But I am also trying to create the moral equivalent of an abstract base, where a lot of the repetitive work is abstracted out of being the concern of the author of various Device Plugins.  I.e.  You will need to handle your own logic around Allocate et al, but Start(), Stop() etc can be done in a single standard way.  That was the thought anyway, I'm open to other approaches :)

deviceplugin/deviceplugin.go
hagbard5235: I'm not finding where I defer cancel...

hagbard5235:
I'm closing this PR to break it up into smaller bits and submit those :)

PR #4, opened 2018-04-17
Add some notes on prerequisites to build on a Mac

mestery:


mestery:
Good idea! I'll update this PR with that change.

PR #5, opened 2018-04-18
Add dependencies using dep

mestery:
This commits adds dependencies using the golang dep dependency
management tool. This commit was generated by installing dep
and then running `dep init`. Please see the discussion here [1]
for details on the advantages of checking in the `vendor/`
directory vs. not checking it in.

[1] https://github.com/golang/dep/blob/master/docs/FAQ.md#should-i-commit-my-vendor-directory


hagbard5235:
LOL... I did this (along with to many other things) here: https://github.com/ligato/networkservicemesh/pull/3

Given that my PR is going to take some sorting out, I'd recommend merging this one to get us to dep faster :)

@fkautz thoughts?  Do you want to do the honors?

mestery:
Ha! I had a sense, sorry for the overlap. But I agree, lets merge this one to get to dependency tracking now. I'll take a peek at #3 next to make sure the next work I do doesn't overlap as well. ;)

hagbard5235:
I deeply appreciate your patch.  I literally was thinking of splitting it out myself, and couldn't be more delighted that you beat me to the punch this morning :)

Particularly at this stage, please don't block on overlap.  My PR is basically an abundance of transparency, it's not at all baked yet.  I pushed it *because* failing to push it had the tendency to lead to a game of 'follow the bouncing repo' which I wanted to avoid.

That said, it's probably a good place to converse a bit about direction.  Honestly, I have no ego invested there, and as long as we move forward would be happy to withdrawal the PR when the time comes.

PR #8, opened 2018-04-18
Add Build Status and License headers

mestery:


mestery:
This won't work until someone from the ligato GitHub org enables travis on the ligato/networkservicemesh repository, FYI.

fkautz:
Ed will need to do that, I don't appear to have access

PR #12, opened 2018-04-19
Proposal for reorging to better align with how k8s handles .proto APIs

hagbard5235:
This pull request is a proposal to attempt to bring us closer into alignment with how k8s handles pull requests.


Makefile
```diff
@@ -1,11 +1,11 @@
...
-PROTOS = $(wildcard networkservice/*.proto)
+PROTOS = $(wildcard pkg/*/apis/**/*.proto)
 PBS = $(PROTOS:%.proto=%.pb)
 
 all: protos
 protos: ${PBS}
 
 %.pb: %.proto
-	PATH=~/bin:$(PATH):$(GOBIN) protoc -I networkservice $*.proto --go_out=plugins=grpc:go
+	PATH=~/bin:$${PATH}:$(GOBIN) protoc -I $(dir $*.proto) $*.proto --go_out=plugins=grpc:$(dir $*.proto)
```
fkautz: Can you double check this? The last flag should probably be --go_out=plugins=grpc:go

hagbard5235:
Hold off on merge for a bit...

fkautz:
Nevermind about the --go-out comment. We should verify that it is in the correct. However, you should delete api.pb.pb.go and replace it with the generated api.pb.go.

* pkg/nsm/apis/pod2nsm/api.pb.pb.go is present
* pkg/nsm/apis/pod2nsm/api.pb.go is missing

hagbard5235:
Yeah... on the --go-out comment I was breaking it into two pieces in my head:

1)  Does go-out work the way I think it works
2)  Is that really were we want the generated go

Both of which are valid in my mind.  I totally take your point about multiple .pb.pb, working it :)

PR #13, opened 2018-04-19
Simple device plugin 'abstract' implementation.

hagbard5235:
The NSM is going to need to provide a 'deviceplugin' for
    a resource
    
    nsm.ligato.io
    
    so that Pods can request NSM as a resource.
    
    The NSM is also going to provide a 'deviceplugin' for
    resources:
    
    nsm.ligato.io/${service}/${channel}
    
    so that the Pod may request an actual hardware NIC providing
    that Network Service.
    
    In otherwords, the NSM will need to provide multiple 'deviceplugins'.
    
    For this reason, I'm attemtping to abstract he housekeeping parts
    of the 'deviceplugin' into a deviceplugin.go and we can then use
    Go composition to construct easily the *specific* device plugins we
    need for those resources.
    
    References:
    "Why Go Structs are superior to class based inheritance":
    https://medium.com/@simplyianm/why-gos-structs-are-superior-to-class-based-inheritance-b661ba897c67
    
    K8s DevicePlugin documentation:
    https://kubernetes.io/docs/concepts/cluster-administration/device-plugins/


fabiand:
Just FYI - We were also creating a few DPs and we also have an abstraction: DPM https://github.com/kubevirt/device-plugin-manager/

Side note IIRC there was some issue with more than one slash in resource names, i.e. `
nsm.ligato.io/${service}/${channel}` might not work

/cc @phoracek

deviceplugin/deviceplugin.go
```diff
@@ -0,0 +1,143 @@
...
+	return &DevicePlugin{
+		socket: serversock,
+		stop: make(chan interface{}),
+		resourceName: resourcename,
+	}
+}
+
+func dial(unixSocketPath string, timeout time.Duration) (*grpc.ClientConn, error) {
+	c, err := grpc.Dial(unixSocketPath, grpc.WithInsecure(),grpc.WithBlock(),
+		grpc.WithTimeout(timeout),
```
fkautz: This is deprecated, we should use context.WithTimeout instead.

fkautz:
LGTM

PR #20, opened 2018-04-22
Added gofmt check to .travis.yml

hagbard5235:


hagbard5235:
It looks like I broke the Travis CI with this... back to the drawing board...

PR #21, opened 2018-04-22
Add go fmt check to .travisci

hagbard5235:
Note: the first patch will fail because the code does not clear the fmt check.

A follow up patch will fix it :)


hagbard5235:
@fkautz I put in the ${RH copyright string} to remind us to get it right :)

hagbard5235:
I believe we now have the correct RedHat copyright.

PR #25, opened 2018-04-24
Use a goroutine to block for the signal

mestery:
Experimenting with a more go-friendly way of handling signals.

### DIFF
--- a/cmd/nsm/main.go
+++ b/cmd/nsm/main.go
@@ -5,13 +5,13 @@ import (
 	"log"
 	"os"
 	"os/signal"
-	"syscall"
 	"sync"
+	"syscall"
 )
 
-var wg sync.WaitGroup
-
 func main() {
+	var wg sync.WaitGroup
+
 	log.Println("Starting NSM")
 
 	dp := nsmdp.NewNSMDevicePlugin()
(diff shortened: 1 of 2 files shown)


cmd/nsm/main.go
```diff
@@ -20,10 +20,9 @@ func main() {
 		syscall.SIGINT,
 		syscall.SIGTERM,
 		syscall.SIGQUIT)
-	select {
-	case s := <-sigChan:
+	go func() {
```
edwarnicke: Is this still producing a race condition?

fkautz: What are we trying to accomplish here?

fkautz: Try using this to block main from exiting.

https://golang.org/pkg/sync/#WaitGroup

mestery: OK, I've got an update which adds WaitGroup support, and when coupled with #29, this now works as expected! New commit pushing in 30 seconds!

nsmdp/nsmdp.go
```diff
@@ -36,6 +36,7 @@ func NewNSMDevicePlugin() *NSMDevicePlugin {
 		DevicePlugin: deviceplugin.NewDevicePlugin(serverSock, resourceName),
 		devs:         make(map[string]*NSMDevice),
 		stop:         make(chan interface{}),
+		updatedevs:   make(chan *NSMDevice, 10),
```
mestery: This is the change in #29, if we merge these two together, I'll just close #29. Or, I can remove this and we can merge them separately.

PR #26, opened 2018-04-24
Add a build.sh script

mestery:
This commit moves the build logic out of the Dockerfile and into
a script, thus allowing a user to run the build without having
to copy and paste from the Dockerfile.


scripts/build.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+#
+# This script builds the networkservicemesh
+#
+
```
edwarnicke: I think we want to retain:

test -z $(go fmt ./...) || (echo "Run go fmt ./... and recommit your code";exit 1)

To do the format checking.

scripts/build.sh
```diff
@@ -0,0 +1,10 @@
...
+go get -u github.com/golang/protobuf/protoc-gen-go
+go generate ./...
+go install ./...
+go test ./...
```
fkautz: Can you add -race to this?

Race doesn't provide false positives, so it should be safe to add.

scripts/build.sh
mestery: I had thought I left that in, but I will add it back, as it does appear to be gone.

scripts/build.sh
mestery: Will do!

scripts/build.sh
mestery: Actually, the format check remains in the Dockerfile, see line 5 above. Let me know if you'd prefer it moved here.

edwarnicke: Ah... I missed that.  I could see either way, so long as we are deliberate about it :)

scripts/build.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+#
+# This script builds the networkservicemesh
+#
+
+go get -u github.com/golang/protobuf/protoc-gen-go
+go generate ./...
+go install ./...
+go test -race ./...
```
mestery: So it turns out using "go test -race" with alpine linux is a non-starter [1]. I am going to remove this for now.

[1] https://github.com/golang/go/issues/14481

scripts/build.sh
mestery: @fkautz See here as to why "go test -race" won't work with alpine linux:

https://github.com/golang/go/issues/14481

scripts/build.sh
mestery: If that's the case, let me move it, as it may make more sense to have it in one place.

PR #29, opened 2018-04-25
Initialize channel as buffered

mestery:
If this channel is not buffered, the program hangs waiting
for something to read from it after we add the first item
into the channel.


mestery:
This was integrated into #25 and merged there.

PR #32, opened 2018-04-25
Adding race test to travis

fkautz:


scripts/race-test.sh
```diff
@@ -0,0 +1 @@
+go test -race ./...
```
mestery: I am of two opinions here: 1) We add this into build.sh and pass arguments into build.sh to trigger running this (and other future things). 2) We rename race-test.sh to test.sh and move all tests into it, again with arguments for this test. Thoughts?

build/nsm/docker/Test.Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM golang:1 as build
```
mestery: I had started this earlier, but never got around to finishing it nor pushing out a PR, happy to see you did this! :)

scripts/race-test.sh
fkautz: I think tests should be a part of build. There is no reason to waste upstream resources if it doesn't pass tests locally. Tests are fast at this point. If they ever get too slow, we can reconsider.

build/nsm/docker/Test.Dockerfile
fkautz: And now we always use the latest stable go :)

build/nsm/docker/Test.Dockerfile
```diff
@@ -0,0 +1,6 @@
...
+ENV PACKAGEPATH=github.com/ligato/networkservicemesh/
+RUN apk add --update protobuf git bash
```
mestery: The golang base doesn't have apk, thus causing travis to be sad and fail. I believe the golang base is Debian based and you may want a "apt-get install protobuf git bash" here. Note I may have the package names wrong, but you can figure those out ;).

PR #33, opened 2018-04-26
First pass at model

mestery:
This is a first pass at modeling NetworkService and NetworkServiceEndpoint.


mestery:
Looking for some feedback on this, I tried to read Ed's mind from the slides and put down what is in there for NetworkService and NetworkServiceEndpoint.

netmesh/model/netmesh/netmesh.proto
```diff
@@ -0,0 +1,33 @@
+syntax = "proto3";
+
+package netmesh;
+
+enum NetmeshApiVersion {
```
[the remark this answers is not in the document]
mestery: Ack

netmesh/model/netmesh/netmesh.proto
```diff
@@ -0,0 +1,33 @@
...
+    NETMESH_API_VERSION_1 = 0;
+};
+
+enum NetmeshPayload {
```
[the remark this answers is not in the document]
mestery: Ack

netmesh/model/netmesh/netmesh.proto
```diff
@@ -0,0 +1,33 @@
+syntax = "proto3";
+
+package netmesh;
+
+enum ApiVersion {
+    NETMESH_API_VERSION_1 = 0;
```
fkautz: Do we want to use an integer representation here? I would prefer to see some form of semver for api versions. It's a bit more noisy, but I think it would be worth it.

netmesh/model/netmesh/netmesh.proto
```diff
@@ -0,0 +1,33 @@
...
+};
+
+enum PayloadType {
```
fkautz: Do we want to dictate payloads types as an enumerable here? There are other payloads that we may see, e.g. Geneve, GRE, LISP. There may be others we don't know about. We could extend this list over time, but it may be better to leave this as a fillable label with the schema for that label managed outside of the proto.

netmesh/model/netmesh/netmesh.proto
```diff
@@ -0,0 +1,33 @@
...
+    string name = 1;
+    string uuid = 2;
+    string selector = 3;
+
+    message NetmeshChannel {
+        string name = 1;
+        PayloadType payload = 2;
+    };
+    repeated NetmeshChannel channels = 4;
+    ApiVersion version = 5;
```
fkautz: If we separate out the api version from this call, we can do some initial handshake to determine whether the client and service are capable of communicating with each other, rather than adding that to the data path. This would allow us to fail fast on initialization rather than on service request.

netmesh/model/netmesh/netmesh.proto
mestery: A quick search leads me to this:

https://github.com/blang/semver

I think we can pull that in, and remove versions from the protobuf here completely, making it a part of the initial handshake, as you mention below.

netmesh/model/netmesh/netmesh.proto
mestery: I'll change the payload value to a string for now.

netmesh/model/netmesh/netmesh.proto
mestery: See above.

PR #36, opened 2018-05-01
Protobuf 1.1.0

mestery:
This moves our protobuf dependency to version 1.1.0, the latest release.


vendor/github.com/ligato/networkservicemesh/LICENSE
```diff
@@ -0,0 +1,202 @@
+
```
edwarnicke: Why are we vendoring our own code here?

vendor/github.com/ligato/networkservicemesh/deviceplugin/deviceplugin.go
```diff
@@ -0,0 +1,142 @@
+// Package deviceplugin provides a basic implementation of a Kubernetes DevicePlugin without content.
```
edwarnicke: Why are we vendoring our own code here?

vendor/github.com/ligato/networkservicemesh/nsmdp/nsmdp.go
```diff
@@ -0,0 +1,106 @@
+package nsmdp
```
edwarnicke: Why are we vendoring our own code here?

PR #42, opened 2018-05-02
Moving runtime image from alpine to busybox:glibc

fkautz:


build/nsm/docker/Dockerfile
```diff
@@ -1,10 +1,15 @@
-FROM golang:1-alpine as build
+FROM golang:1 as build
```
mestery: Just to be clear, these are Debian images? The apt commands below lead me to believe that, but perhaps busybox also uses apt.

build/nsm/docker/Dockerfile
```diff
@@ -1,10 +1,15 @@
...
 ENV PACKAGEPATH=github.com/ligato/networkservicemesh/
-RUN apk add --update protobuf git bash
+
+RUN apt update
+RUN apt install -y git bash unzip
+RUN wget https://github.com/google/protobuf/releases/download/v3.5.1/protoc-3.5.1-linux-x86_64.zip
```
mestery: I like this implicit install better than whatever version apk was giving us above.

edwarnicke:
Overall I'm positively inclined towards this change.  Questions though:

1)  Do the glibc versions match between the golang image we are using to build and the busybox image we are using to run?
2)  Do we care?

build/nsm/docker/Dockerfile
fkautz: The build is on Debian stretch. The runtime is on busybox with the compiled version copied over.

mestery: Thanks for clarifying! Now I have the same question as @edwarnicke :)

PR #43, opened 2018-05-02
Statically compiling nsm

fkautz:


scripts/build.sh
```diff
@@ -5,22 +5,9 @@
...
-	shift
-done
-
 [ -d vendor/github.com/ligato/networkservicemesh/ ] && (echo "Run: rm -rf vendor/github.com/ligato/networkservicemesh;dep ensure";exit 1)
 test -z $(go fmt ./...) || (echo "Run go fmt ./... and recommit your code";exit 1)
 go get -u github.com/golang/protobuf/protoc-gen-go
 go generate ./...
-go install ./...
-go test $RACE_ENABLED ./...
+CGO_ENABLED=0 go install -ldflags '-w -extldflags "-static"' ./...
```
mestery: I'm ok with this for now, we can change the static linking later if we need to.

Issue #52, opened 2018-05-18
When running and then stopping networkservicemesh as a daemonset, the Close() plugin function isn't called


mestery:
We need to understand what signal is generated to the containers when a daemonset is stopped. Currently, our Close() function is not called when deleting a daemonset containing networkservicemesh. This may require changes in the Ligato framework should it need to handle a signal other than SIGINT [2].

[1] https://github.com/ligato/networkservicemesh/blob/master/plugins/netmesh/plugin_impl_netmesh.go#L91
[2] https://github.com/ligato/cn-infra/blob/master/core/event_loop.go#L25

mestery:
Closed since #55 merged.

PR #55, opened 2018-05-22
Update to latest version of cn-infra

mestery:
This pulls in the following fix [1]. This allows EventLoopWithInterrupt to
handle SIGTERM, meaning when NSM is run as a daemonset it will handle
SIGTERM when the daemonset is deleted.

[1] https://github.com/ligato/cn-infra/pull/268


mestery:
This closes #52.

PR #54, opened 2018-05-21
Add a custom resource definition plugin

mestery:
This adds a CRD plugin into networkservicemesh. This takes care of adding
custom resources for networkserviceendpoints, networkservicechannels, and
networkservices. This includes validators for the resources as well, and
makes use of the existing generated go code from our protobufs as the
payload.

In practice, this appears as the following:

```
localadmin@kmestery1604:~$ kubectl get crd
NAME                                          AGE
networkeservicendpoints.networkservice.mesh   4m
networkservicechannels.networkservice.mesh    4m
networkservices.networkservice.mesh           4m
localadmin@kmestery1604:~$ kubectl get NetworkService -o yaml
apiVersion: v1
items:
- apiVersion: networkservice.mesh/v1
  kind: NetworkService
  metadata:
    clusterName: ""
    creationTimestamp: 2018-05-21T18:16:56Z
    name: networkservice-radio
    namespace: default
    resourceVersion: "8781"
    selfLink: /apis/networkservice.mesh/v1/namespaces/default/networkservices/networkservice-radio
    uid: 25014149-5d23-11e8-a84b-0800275bf7f3
  spec:
    message: NetworkService Radio Network
    name: Hog*Invalid*Name
    selector: purple
    thisshouldfail: yes-it-should
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
localadmin@kmestery1604:~$
```

Signed-off-by: Kyle Mestery <mestery@mestery.com>

Issue #48, opened 2018-05-18
Implement Customer Resource Definitions


mestery:
We have a protobuf file containing NetworkServices and NetworkServiceEndpoints, we need to implement CRD support to enable these resources to exist in the Kubernetes DB and be manipulated with kubectl.

mestery:
#54 merged so marking this as closed now, any further work done around CRDs will be via new issues.

PR #56, opened 2018-05-25
Add k8s.io/code-generator dependency via dep

mestery:


mestery:
The idea is to merge the generators as this separate patch, then I can rebase #54 and make that one a bit smaller.

mestery:
I should note this PR is ultimately a part of #48.

Gopkg.toml
```diff
@@ -55,10 +61,20 @@ required = ["github.com/golang/protobuf/protoc-gen-go"]
...
   version = "3.3.5"
 
 [prune]
   go-tests = true
   unused-packages = true
+
+  [[prune.project]]
+    name = "k8s.io/code-generator"
+    unused-packages = false
+    non-go = false
```
fkautz: non-go = false is redundant and should be removed.

mestery: I was wondering about this, thanks!

mestery:
Verified k8s.io/gengo and k8s.io/code-generator are both licensed under Apache 2. golang.org/x/tools uses the following license:

https://github.com/golang/tools/blob/master/LICENSE

Issue #50, opened 2018-05-18
Integration test coverage


mestery:
We need to settle on some way to test NSM functionally.

mestery:
Closing this now, the initial integration tests have merged, we'll open another PR if we need to add more.

PR #60, opened 2018-05-31
Add sample configuration and scripts to test CRDs

mestery:
This commit adds some sample yaml configuration files to create all the CRDs
in use by Network Service Mesh. It also adds some simple shell scripts to
test creation and deletion of these resources.


mestery:
I should note these scripts could be the start of some functional testing for issue #50.

conf/sample/teardown.sh
```diff
@@ -0,0 +1,17 @@
...
+kubectl delete -f conf/sample/networkservice.yaml
+
+# Delete the endpoints
+kubectl delete -f conf/sample/networkservice-endpoint.yaml
+
+# Delete the channel
+kubectl delete -f conf/sample/networkservice-channel.yaml
+
+# Now delete the daemonset
+kubectl delete -f conf/sample/networkservice-daemonset.yaml
```
prateekgogia: LGTM
nit: this will not clean up CRD objects in Kubernetes. Its not needed now, but may be later for cleaning up.

prateekgogia: Ok, I see the point.

Issue #59, opened 2018-05-31
Deletion of CRDs isn't working as expected


mestery:
For some reason, on a CRD delete, this is triggered, meaning the factory can't find the object. An example of the error I get is as follows:

```
time="2018-05-30 18:51:03.41696" level=info msg="Read item 'default/yogi-bear' off workqueue. Processing..." loc="crd/crd_queue.go(67)" logger=netmeshcrd
time="2018-05-30 18:51:03.41759" level=info msg="Dequeuing interface of type networkservices" loc="crd/crd_queue.go(70)" logger=netmeshcrd
ERROR: logging before flag.Parse: E0530 18:51:03.417636       1 crd_queue.go:74] error getting object 'default/yogi-bear' from api: networkservice.networkservicemesh.io "yogi-bear" not found
```

PR #61, opened 2018-05-31
Fix informer factory usage

mestery:
While I'm in here, add in per-CRD channels as well.


mestery:
Note this does not fix #59, but it does fix some issues uncovered by #60.

Issue #53, opened 2018-05-18
Investigate in-pod authentication


mestery:
The review for #46 indicated a desire to explore in-pod authentication. This issue tracks that investigation.

prateekgogia:
@mestery  Can we can close this?

mestery:
Was this implemented as part of #67? If so, yes, lets close it.

prateekgogia:
Yes, #67 adds support for in-cluster authentication.

PR #67, opened 2018-06-02
Allows pod to access API using in-cluster config, adding RBAC rule

prateekgogia:
K8S API access can be restricted more based on API being used. This provides a starting point to deploy NSM in k8s cluster. Without this API access keeps failing and no CRDs can be created.

Signed-off-by: Prateek Gogia <prateekgogia@hotmail.com>

### DIFF
--- a/conf/sample/networkservice-daemonset.yaml
+++ b/conf/sample/networkservice-daemonset.yaml
@@ -10,7 +10,38 @@ spec:
     spec:
       nodeSelector:
         app: networkservice-node
+      serviceAccountName: networkservicemesh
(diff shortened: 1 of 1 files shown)


prateekgogia:
Fixes #53

fkautz:
There are some conflicts that need to be resolved before we can merge this in.

Also, we should tighten up what API calls that the network service mesh has access to. We can merge this in for now. Created #69 to track this concern.

prateekgogia:
@fkautz PTAL, I resolved the conflict.

PR #107, opened 2018-06-22
Client Connectivity API

fkautz:


pkg/nsm/apis/nsmconnect/nsmconnect.proto
```diff
@@ -0,0 +1,90 @@
...
+    FIRST = 1;
+    SECOND = 2;
+    THIRD = 3;
+    FORTH = 4;
+    FIFTH = 5;
+}
+
+// ConnectionRequest is sent by a client or NSEP to build a connection.
+message ConnectionRequest {
+// Since connection request will trigger certain actions
```
mestery: This is a pretty pedantic nit, but should we shift these comments to match the indentation below them?

PR #110, opened 2018-06-23
Moving away from deprecated Version to Versions in CRD definition

fkautz:


mestery:
Looks like the CI is using an older version of Kubernetes which is expecting `Version` instead of `Versions`. I think you'll want to change the version of Kubernetes installed by updating the version in the integration test script here:

https://github.com/ligato/networkservicemesh/blob/master/scripts/cluster-up-minikube.sh#L30

PR #113, opened 2018-06-23
Utilities to make idempotent plugins easy

edwarnicke:
Embedding
idempotent.Impl

In a plugin and writing the Plugin Init() and Close() as:

    func (plugin *Plugin) Init() error {
         return plugin.IdempotentInit(plugin.init)
    }

    func (plugin *Plugin) Close() error {
         return plugin.IdempotentClose(plugin.close)
    }

will make sure that the plugin Init() and Close()
are idempotent.  Additionally, refCounting is performed
so that the Plugin isn't *actually* closed until
Close() has been called as many times as Init()

This allows safe sharing of a Plugin.


edwarnicke:
Note: 100% unit test coverage

PR #122, opened 2018-06-28
Add support for listing CRD objects

mestery:
The shared informer we use is a cache of API objects. Currently, we simply
make use of the cache for Create/Update/Delete operations, but we can use
a generated Lister() to traverse the cache. This patch exposes this
functionality as an API from the CRD Plugin.

Note that I need to convert the CRD plugin into an idempotent plugin for
this to be useful, as there is a circular dependency at the moment: The
CRD plugin uses the handler plugin API, and the handler would need to
call the CRD list APIs.


mestery:
Closing, after discussing with @sbezverk we want to use the Informer cache to simply cache API objects, and we'll use our objectstore cache to cache things locally for NSM clients.

PR #123, opened 2018-06-28
Breakout object types in the handler plugin

mestery:
We pass all objects into the handler APIs as interface objects, meaning
we don't pass metadata. We can rely on go's built-in type checking
ability to verify the type of object passed in.


mestery:
Closing in lieu of pushing object abstractions into the objectstore code.

PR #142, opened 2018-07-06
Idempotent logger and config plugins

edwarnicke:
This commit adds idempotent logger and config plugins.

The config is based on Viper, and makes use of cobra

The logger is based on logrus

logger has 100% code coverage

config is *almost* 100%, but has an err check for a place
where an err is returned, but the underlying source can't
possibly return a non nil err.  But since it could in the
future, the err is checked, but literally can't be tested.


edwarnicke:
@sbezverk My next PR (or possibly two) will start migrating to using it.

As to line numbers, logrus supports 'Hooks' (which are awesome) and I've got a set of [defaultHooks](https://github.com/edwarnicke/networkservicemesh/blob/e1db0c67ad07a0b5b7f2b8097da23a5e6e6dd9fc/plugins/logger/options.go#L99) that I [add](https://github.com/edwarnicke/networkservicemesh/blob/e1db0c67ad07a0b5b7f2b8097da23a5e6e6dd9fc/plugins/logger/options.go#L119).

plugins/config/config_test.go
```diff
@@ -0,0 +1,150 @@
...
+	cmd := &cobra.Command{Use: "command"}
+	command.SetRootCmd(cmd)
+	plugin := config.NewPlugin()
+	Expect(plugin).NotTo(BeNil())
+	Expect(plugin.Deps.Name).To(Equal(config.DefaultName))
+	Expect(plugin.Deps.DefaultConfig).ToNot(BeNil())
+	Expect(plugin.Deps.Cmd).To(Equal(cmd))
+	err := plugin.Init()
+	Expect(err).To(BeNil())
+	config := plugin.LoadConfig()
```
fkautz: We're importing a config package and creating a config variable. We should either rename this variable or alias the config package under a new name.

edwarnicke:
fkautz: I fixed your complaint about config shadowing. (and also rebased on master)

mestery:
Merging since @fkautz's comment was addressed and we can then focus on #145 and the ObjectStore discussion.

PR #143, opened 2018-07-06
Updates for the files that help new starters along

iawells:
README -> BUILD (link fixed) and MINIKUBE (added)
MINIKUBE -> corrections, added missing steps, pointer to startup.sh
startup and teardown scripts made more resilient


fkautz:
@iawells Are you able to resolve these conflicts? The cause of the conflict is that shellcheck was introduced and issues fixed based on its advice.

mestery:
I have similar comments as @fkautz here. And also, the direction you're moving setup.sh into is good, but slowly encroaches on what the developer integration tests [1] do. Just pointing out the changes you've made to wait for CRDs to be created are good, for example, nothing for you to do. Thanks for the patch!

[1] https://github.com/ligato/networkservicemesh/blob/master/scripts/dev-integration-tests.sh

iawells:
Haven't checked the shell lint stuff - any clues as to how to do that?

PR #146, opened 2018-07-08
Stop shadowing the delete command

fkautz:


fkautz:
Some background on this, delete is a reserved keyword used to delete elements from a map.

PR #151, opened 2018-07-09
Ensure shellcheck is checking sample scripts

mestery:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ before_script:
 - sudo mount --make-rshared /
 
 script:
-- shellcheck scripts/*.sh conf/sample/*.sh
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -13,7 +13,7 @@ before_script:
 - sudo mount --make-rshared /
 
 script:
-- shellcheck scripts/*.sh
+- shellcheck scripts/*.sh conf/sample/*.sh
```
fkautz: Can we switch to this?

find . -name "*.sh" -not -path "./vendor/*" | xargs shellcheck

mestery: Or even this:

` shellcheck $(find . -name "*.sh" -not -path "./vendor/*")`

I'll re-spin.

PR #170, opened 2018-07-15
Add CheckDeps, InitDeps, CloseDeps helper

edwarnicke:
The Deps pattern in Plugins is a powerful way to allow
users to wire in alternate implementations for things the Plugin
uses. In addition to the flexibility it offers, it also makes
it much much easier to Mock and test.

Unfortunately, there are several mistakes its easy to make with Deps:

- Making a Deps field private (lowercase) which precludes wiring in
  an alternate implementation
- Failing to set a Default value for Deps in options.go

CheckDeps checks for all of these mistakes. It is suitable to
be run in the NewPlugin(...) in options.go as a final check before
returning a new Plugin to insure its ready to Go :)

*****************

InitDeps - Runs Init() on all Deps implemented with Plugins

Deps should be interfaces, they do not have to be Plugins
InitDeps loops through all of the fields in Deps, and if they
are Plugins, it Init()'s them.

*****************

CloseDeps - Runs Close() on all Deps implemented with Plugins

Closes all Deps of the given plugin if they themselves
are idempotent.Plugins

### DIFF
--- a/utils/helper/deptools/check.go
+++ b/utils/helper/deptools/check.go
@@ -60,7 +60,7 @@ func check(structField *reflect.StructField, value *reflect.Value) error {
 			field := v.Interface()
 			zeroValue := reflect.Zero(v.Type())
 			if field == nil || field == zeroValue {
-				return fmt.Errorf("field in Deps struct  named %s is zero valued, and lacks tag `optional:\"true\"`.  Please either set it to a non-zero value or add the `optional:\"true\"` to indicate its acceptable for it to be zero valued", sf.Name)
+				return fmt.Errorf("field in Deps struct named %s is zero valued, and lacks tag `optional:\"true\"`.  Please either set it to a non-zero value or add the `optional:\"true\"` to indicate its acceptable for it to be zero valued", sf.Name)
 			}
 		}
 	}


edwarnicke:
Note to reviewer: #166 should be merged first as this is based on it.

utils/helper/check_deps.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	if sf == nil {
+		return nil
+	}
+	return checkDeps(sf, vdeps)
+}
+
+func getDeps(p idempotent.PluginAPI) (*reflect.StructField, *reflect.Value, error) {
+	if p == nil {
+		return nil, nil, fmt.Errorf("Cannot call CheckDeps on nil")
```
fkautz: Cannot should be lower case.

utils/helper/check_deps.go
```diff
@@ -0,0 +1,99 @@
...
+		if sf.PkgPath != "" {
+			return fmt.Errorf("%s fieldname \"%s\" cannot be lowercase. %s cannot have private fields.  All fieldnames must be Uppercase", structField.Name, structField.Name, sf.Name)
+		}
+		v := value.Field(i)
+		tag, ok := sf.Tag.Lookup(optionalTagName)
+		if (!ok || tag != optionalTagValue) && v.CanInterface() {
+			field := v.Interface()
+			zeroValue := reflect.Zero(v.Type())
+			if field == nil || field == zeroValue {
+				return fmt.Errorf("field in % struct  named %s is zero valued, and lacks tag `optional:\"true\"`.  Please either set it to a non-zero value or add the `optional:\"true\"` to indicate its acceptable for it to be zero valued", structField.Name, sf.Name)
```
fkautz: Looks like the first % is missing its parameter. E.g. %s, %d.

utils/helper/check_deps_test.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/ligato/networkservicemesh/utils/helper"
+	id1 "github.com/ligato/networkservicemesh/utils/idempotent"
+	. "github.com/onsi/gomega"
+)
+
+type API interface {
+	Running() bool
+}
+
+type PluginAPI interface {
```
fkautz: This exists in a test file and is unused. Should we remove it?

utils/helper/check_deps_test.go
```diff
@@ -0,0 +1,141 @@
...
+	Deps string
+}
+
+func TestCheckDepsNonStructDeps(t *testing.T) {
+	RegisterTestingT(t)
+	p := &PluginStringDeps{Deps: "foo"}
+	Expect(helper.CheckDeps(p)).ToNot(Succeed())
+}
+
+type PluginNonInterfaceDeps struct {
```
fkautz: This type is not currently used and exists in a test. Should we remove?

utils/helper/close_deps_test.go
```diff
@@ -0,0 +1,79 @@
...
+	p := &PluginWithDeps{
+		Deps: NonOptionalDeps{
+			One: &Runable{},
+		},
+	}
+	Expect(helper.CloseDeps(p)).To(Succeed())
+	Expect(p.Deps.One.Running()).To(BeFalse())
+}
+
+type PluginErrorOnClose struct {
```
fkautz: This is not being used.

utils/helper/init_deps.go
```diff
@@ -0,0 +1,55 @@
...
+	// Note: Any error from getDeps would have been hit previously
+	// by CheckDeps
+	sf, vdeps, _ := getDeps(p)
+	if sf == nil {
+		return nil
+	}
+	return initDeps(sf, vdeps)
+}
+
+func initDeps(structField *reflect.StructField, value *reflect.Value) error {
```
fkautz: structField is not being used. I assume this is by design. Can we rename this variable to _ so that it is clear it is unused?

fkautz:
Considering the size of check_deps, would it make sense to make a checkdeps package within helpers? If there is only going to be one helper, perhaps we can rename helper to checkdeps?

utils/helper/check_deps.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func getDeps(p idempotent.PluginAPI) (*reflect.StructField, *reflect.Value, error) {
+	if p == nil {
+		return nil, nil, fmt.Errorf("Cannot call CheckDeps on nil")
+	}
+	t := dereferenceType(reflect.TypeOf(p))
+	v := dereferenceValue(reflect.ValueOf(p))
+	if t.Kind() != reflect.Struct {
+		return nil, nil, fmt.Errorf("CheckDeps requires a plugin that is a struct,%s is a %s", t, t.Kind())
```
fkautz: add a space after struct in the comment for readability

utils/helper/check_deps.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+	sf, ok := t.FieldByName(depsFieldname)
+	if !ok {
```
fkautz: Is it ever possible that ok would be false at this point? If not, we should probably log this as this indicates an error.

utils/idempotent/idempotent_impl.go
```diff
@@ -42,39 +58,61 @@ func (i *Impl) IsIdempotent() bool {
...
+func (i *Impl) Close() error { return nil }
+
 // IdempotentInit increments the refCount and calls init precisely once
 func (i *Impl) IdempotentInit(init func() error) error {
-	i.refCountMutex.Lock()
-	defer i.refCountMutex.Unlock()
-	if i.refCount < 0 { // i.refCount < 0 means we are terminally closed and no longer Init-able
+	i.Lock()
+	defer i.Unlock()
+	if i.state == CLOSED { // We are closed an no longer initable
```
fkautz: replace "an" with "and"

edwarnicke:
@fkautz in response to:

```
Considering the size of check_deps, would it make sense to make a checkdeps package within helpers? If there is only going to be one helper, perhaps we can rename helper to checkdeps?
```

I'm quite warming to the idea of doing separate package for CheckDeps, InitDeps, CloseDeps, and GetDeps (I split out GitDeps in a subsequent PR to avoid a dependency loop).  Lets hold this patch while I split these up.

edwarnicke:
OK, I think this is good to go now.

utils/helper/deptools/check.go
```diff
@@ -0,0 +1,68 @@
...
+		if sf.PkgPath != "" {
+			return fmt.Errorf("Deps fieldname \"%s\" cannot be lowercase. Deps cannot have private fields.  All fieldnames must be Uppercase", sf.Name)
+		}
+		v := value.Field(i)
+		tag, ok := sf.Tag.Lookup(optionalTagName)
+		if (!ok || tag != optionalTagValue) && v.CanInterface() {
+			field := v.Interface()
+			zeroValue := reflect.Zero(v.Type())
+			if field == nil || field == zeroValue {
+				return fmt.Errorf("field in Deps struct  named %s is zero valued, and lacks tag `optional:\"true\"`.  Please either set it to a non-zero value or add the `optional:\"true\"` to indicate its acceptable for it to be zero valued", sf.Name)
```
fkautz: Missed this earlier, struct has two spaces after it instead of one

PR #173, opened 2018-07-16
Add a Plugin Registry

edwarnicke:


utils/registry/registry.go
```diff
@@ -0,0 +1,88 @@
...
+// must be implemented by each plugin.
+
+package registry
+
+import (
+	"reflect"
+	"sync"
+
+	"github.com/ligato/networkservicemesh/utils/helper"
+
```
fkautz: Delete this line and resort the deps.

utils/registry/registry.go
```diff
@@ -0,0 +1,88 @@
...
+// Delete - delete p from registry
+func (r *RegistryImpl) Delete(p idempotent.PluginAPI) {
+	r.Lock()
+	defer r.Unlock()
+	for i, value := range r.plugins {
+		vdeps, _ := helper.GetDeps(value)
+		pdeps, _ := helper.GetDeps(p)
+		if reflect.TypeOf(value) == reflect.TypeOf(p) && reflect.DeepEqual(pdeps, vdeps) {
+			r.plugins = append(r.plugins[:i], r.plugins[i+1:]...)
+			return
```
fkautz: Return is not necessary here, should be removed.

utils/registry/registry.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+	}
+}
+
+// Clear is intended only for testing
```
fkautz: Is there an interface this is implementing? If not, perhaps we can make a test struct that extends RegistryImpl that knows how to clear the plugins slice.

edwarnicke: fkautz: That is a very good idea!

PR #174, opened 2018-07-16
Add proxy support to docker build commands in the Makefile

mestery:
When trying to use the Makefile to build on a machine behind a proxy, it
was observed the docker build commands all failed. This commit adds support
for using a proxy with the docker build commands in the Makefile. It also
fixes an explicit usage of "wget" in the nsm Test.Dockerfile to use a proxy.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,11 +40,6 @@ endif
 
 DOCKERBUILD=docker build ${HTTPBUILD} ${HTTPSBUILD}
 
-# Test target to debug proxy issues
-checkproxy:
-	echo "HTTPBUILD=${HTTPBUILD} HTTPSBUILD=${HTTPSBUILD}"
-	echo "DOCKERBUILD=${DOCKERBUILD}"
-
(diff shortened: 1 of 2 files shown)


mestery:
I have verified `make docker-build` works correctly on a machine behind a proxy as well as on one which is not behind a proxy.

Makefile
```diff
@@ -26,6 +26,25 @@ GOVETTARGETS=cmd \
...
+ifeq ($(HTTPS_PROXY),)
+HTTPSBUILD=
+else
+HTTPSBUILD=--build-arg HTTPS_PROXY=$(HTTPS_PROXY)
+endif
+
+DOCKERBUILD=docker build ${HTTPBUILD} ${HTTPSBUILD}
+
+# Test target to debug proxy issues
+checkproxy:
```
fkautz: reorder checkproxy to be after the default all target

mestery: D'oh! Fixing now.

PR #202, opened 2018-07-24
Ensure veth interfaces are in the UP state

mestery:
I have noticed in my own testing environment the interfaces the simple-dataplane
creates do not show as "UP".

Note that even with this patch ping's from the NSE to the client still do not
work for me.


mestery:
Let me re-run without this patch. But for me, running `scripts/dev-integration-test.sh` fails on both minikube (v0.28.1) on a Linux Ubuntu host and with Docker for Mac.

mestery:
Closing in lieu of #203.

Issue #186, opened 2018-07-20
Publish images to docker hub


fkautz:
NSM images should be published to the docker repo on every successful merge in master.

fkautz:
If there is no easy way to do this from Travis, we can trigger a build in docker hub.

mestery:
I'll take a peek at this one, I think I can get something pushed out relatively quickly to make this work.

lixuna:
7/27/2018: ETA early next week

mestery:
Actually, I have another plan to revive the travis method after discussing with @sbezverk so I'll propose that patch later this afternoon.

PR #208, opened 2018-07-28
Build all musl based artifacts in a single pass and set docker image names

fkautz:


fkautz:
Previous builds were around 11m50s, this build passed in 9m35s. Eliminates roughly 20% of the build time.

Also, prepares the images to be ready to push. Next step (tomorrow) is pushing these images up to github once this patch lands.

mestery:
Let me push my version which does this and other cleanup, and also does the `docker push` and we can like combine the two. Sound good @fkautz? I was working on these as a part of #186 .

PR #222, opened 2018-08-02
Update dependencies to the latest

mestery:


mestery:
I should note that we just merged one of these, but after that one merged and I went back to look at updating to cn-infra v1.4.0, these popped up after running `dep ensure -update`.

PR #236, opened 2018-08-06
Push images to Docker Hub on each merge to master

mestery:
For now, we should push images to Docker Hub each time we merge to master.
Eventually we'll want to do this based on tags or branches, but given the
velocity of development, I think it's prudent to push with each merge.


mestery:
Merging to move this forward, we can change this back to pushing to Docker Hub on tags later.

PR #241, opened 2018-08-07
Fix embarrassing spelling error in Makefile targets

mestery:
This was preventing `docker push` from working for all repositories
because the Makefile targets it was incorrectly trying to run had
.PHONY entries but no actual entries.

*sigh*


fkautz:
If only we had a bot that could automerge... go go gadget auto merge bot!

Barring that, feel free to merge when this passes.

PR #247, opened 2018-08-09
Guidelines for extending NSM

doonhammer:
Signed-off-by: John McDowall <jmcdowall@paloaltonetworks.com>

Draft guidelines for extending NSM and suggestions how to re-organize code layout to make it easier for developers to keep all extensions in a single branch of tree.

### DIFF
--- a/docs/extending-nsm.md
+++ b/docs/extending-nsm.md
@@ -24,12 +24,12 @@ docker-push: add "docker-push-cnf-dataplane"
 
 ## Dockerfile
 
-A new dockerfile is added: networkservicemesh/build/Dockerfile-cnf.dataplane
+A new dockerfile is added: networkservicemesh/build/Dockerfile.cnf-dataplane
 
 
 ## API Directory and Interface Definition
 
-Create a new directory for the cnf-dataplane daemonset CRD interface definition.
+Create a new directory for the cnf-dataplane daemonset API interface definition.
 
 The new directory is: networkservicemesh/pkg/nsm/apis/cnfdataplane
 
@@ -61,7 +61,7 @@ A new file is created in the directory: networkservicemesh/cmd/cnf-dataplane/cnd
 ### DaemonSet YAML File
 To activate the NSM plugin end users need to create a YAML file to deploy the DaemonSet. 
 
-This file should be created: networkservicemesh/conf/sample/vnf-dataplane.yaml
+This file should be created: networkservicemesh/conf/sample/cnf-dataplane.yaml
 
 This will define the DaemonSet to be run in each Pod.
 
--- a/docs/extending-nsm.md
+++ b/docs/extending-nsm.md
@@ -1,0 +1,129 @@
+# Extending Network Service Mesh (NSM)
+
+
+# Introduction
+
+This document is initially aimed at developers looking to extend the NSM with new plugins to provide users with new functionality. It is based on the version of NSM available August 2018 in the [Network Service Mesh Github](https://github.com/ligato/networkservicemesh) Repository. It documents the steps that were used to create the simple-dataplane extension. 
+
+All steps assume the top level directory is networkservicemesh from github.
+
+
+# Development Steps
+
+The development steps are to add a new data plane called cnf-dataplane.
+
+
+## Edit Makefile
+
+The top level Makefile [networkservicemesh/Makefile] to add to the following targets
+
+docker-build: add "docker-build-cnf-dataplane"
+
+docker-push: add "docker-push-cnf-dataplane"
+
+
+## Dockerfile
+
+A new dockerfile is added: networkservicemesh/build/Dockerfile-cnf.dataplane
+
+
+## API Directory and Interface Definition
+
+Create a new directory for the cnf-dataplane daemonset CRD interface definition.
+
+The new directory is: networkservicemesh/pkg/nsm/apis/cnfdataplane
+
+The interface definition file written in Google Protobufs is added to this directory. 
+
+The file will be:  networkservicemesh/pkg/nsm/apis/cnfdataplane/cnfdataplane.proto
+
+The interface definitions must match the go code created for the daemonset.
+
+
+## Documentation
+
+A documentation file should be added to the documentation directory: networkservicemesh/docs.
+
+The suggested naming for the documentation file is:
+
+    networkservicemesh/docs/README-cnfdataplane.md
+
+## DaemonSet Code
+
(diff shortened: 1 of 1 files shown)


docs/extending-nsm.md
```diff
@@ -0,0 +1,129 @@
...
+This document is initially aimed at developers looking to extend the NSM with new plugins to provide users with new functionality. It is based on the version of NSM available August 2018 in the [Network Service Mesh Github](https://github.com/ligato/networkservicemesh) Repository. It documents the steps that were used to create the simple-dataplane extension. 
+
+All steps assume the top level directory is networkservicemesh from github.
+
+
+# Development Steps
+
+The development steps are to add a new data plane called cnf-dataplane.
+
+
```
mestery: It's unclear to me if we need to also create the repository on Docker Hub or if the push will do that automatically. If it's the former, then we should note that one of the NSM maintainers should create the Docker Hub repository before the patch adding `docker push` merges.

docs/extending-nsm.md
```diff
@@ -0,0 +1,129 @@
...
+The top level Makefile [networkservicemesh/Makefile] to add to the following targets
+
+docker-build: add "docker-build-cnf-dataplane"
+
+docker-push: add "docker-push-cnf-dataplane"
+
+
+## Dockerfile
+
+A new dockerfile is added: networkservicemesh/build/Dockerfile-cnf.dataplane
```
mestery: It should be `Dockerfile.cnf-dataplane`

docs/extending-nsm.md
```diff
@@ -0,0 +1,129 @@
...
+
+
+## API Directory and Interface Definition
+
+Create a new directory for the cnf-dataplane daemonset CRD interface definition.
```
mestery: These are not CRDs, so: `s/CRD/API/`

docs/extending-nsm.md
```diff
@@ -0,0 +1,129 @@
...
+A new directory should be created: networkservicemesh/cmd/cnf-dataplane
+
+A new file is created in the directory: networkservicemesh/cmd/cnf-dataplane/cnd-dataplane.go
+
+## YAML Files
+
+### DaemonSet YAML File
+To activate the NSM plugin end users need to create a YAML file to deploy the DaemonSet. 
+
+This file should be created: networkservicemesh/conf/sample/vnf-dataplane.yaml
```
mestery: Typo here: `s/vnf-dataplane/cnf-dataplane/`

docs/extending-nsm.md
```diff
@@ -0,0 +1,129 @@
...
+
+To use the plugin the end user will need to add the following to their Pod Yaml file. Need to provide an example how to integrate into a simple application. The example should include a source and destination Pod that enables network traffic over NSM
+
+**TODO** What is the end user YAML file.
+
+# Proposed Directory Structure
+The current directory structure for adding new plugins is not optimal as files are scattered over multiple parts of the tree. It is suggested that a normalized directory structure is developed to isolate all new plugins in their own tree.
+
+
+A proposed sample tree is shown below:
```
mestery: This tree does not look correct to me. For example, `apis cmd conf` are all top-level and do not fall out from simple-dataplane. Can you correct this?

docs/extending-nsm.md
doonhammer: Yes me neither, this goes to how we want to separate out the "new" plugins developed on top of the base infrastructure. Open to suggestions.

docs/extending-nsm.md
doonhammer: Will fix

docs/extending-nsm.md
doonhammer: Will fix

docs/extending-nsm.md
doonhammer: Will fix

docs/extending-nsm.md
doonhammer: So it does not match the current tree. I am trying to think/propose a new tree that collects all the artifacts for a new plugin into a single branch of the main tree. Adding a new plugin today requires modifications to a lot of places in the main tree. I see this having a few issues:
1. Plugin developers need to edit a lot of different places and possibly make mistakes.
2. Core code could get broken by plugin developers.
3. If we have lots of plugins (hopefully) do I need to build everything and publish to the same container registry? If I develop a GKE specific plugin I just want to publish to GKE, same for AWS etc...

I know this would involve a lot of changes but just trying to think ahead.

docs/extending-nsm.md
mestery: To keep it simple for now, we can store plugins in the `networkservicemesh` namespace on Docker Hub. We can revisit later. /cc @fkautz @edwarnicke @sbezverk

docs/extending-nsm.md
mestery: I see where you're headed, but I'd like to discuss this on the call. I'd almost be leaning to developing the plugin out of tree at this point, because otherwise the directory structure and repository begins to become unwieldy.

doonhammer: Agreed on out of tree - not quite sure how to organize the build for that. If we could create an example and then proposed to the broader community would that be a way to go?

doonhammer:
I created a new repo [msn_template](https://github.com/doonhammer/nsm_template) that creates a new tree for sample code and fills in a lot of the boiler plate code. It reduces the development hopefully to editing the golang and protobuf files. If there was someway to have this (or something like it) in tree and out of tree would become  a developers choice.

docs/extending-nsm.md
```diff
@@ -0,0 +1,129 @@
+# Extending Network Service Mesh (NSM)
```
tfherbert: I wonder if this shouldn't be named Network Service Mesh Developer's Guide or Network Service Mesh Extensions Developer's Guide.

tfherbert: I think I should withdraw this comment but I am still wondering about the scope of this document the example described is a dataplane extension. Will this procedure also apply to any other plugins?

docs/extending-nsm.md
```diff
@@ -0,0 +1,129 @@
...
+
+    networkservicemesh/docs/README-cnfdataplane.md
+
+## DaemonSet Code
+
+This is the core file to implement the networking required for the plugin. It is deployed as a DaemonSet and the NSM infrastructure will call the entry points when a YAML file is deployed that activates the NSM infrastructure.
+
+A new directory should be created: networkservicemesh/cmd/cnf-dataplane
+
+A new file is created in the directory: networkservicemesh/cmd/cnf-dataplane/cnd-dataplane.go
```
tfherbert: Shouldn't this be:
networkservicemesh/cmd/cnf-dataplane/cnf-dataplane.go

PR #258, opened 2018-08-21
Update vendored dependencies

mestery:
According to this blog post [1], Dependabot does support golang with dep,
but it does not quite yet support vendored dependencies. This means it
will update Gopkg.[lock,toml] for us.

In practice, this means when we see a Dependabot PR, someone should pull
it down and manually run `dep ensure -update` to update our vendored
dependencies.

Note that Dependabot is fixing this behavior and hopefully soon we won't
have to do this anymore.

For now, this updates our dependencies after some recent Dependabot PRs.

[1] https://dependabot.com/blog/go-support


ondrej-fabry:
The command `dep ensure -update` actually tries to update all deps to latest possible version. It should be enough for running `dep ensure -v` to sync `Gopkg.lock` with `vendor` folder after dependabot updates it.

mestery:
Closing, will open a new one which only runs `dep ensure` per advice from @ondrej-fabry.

PR #259, opened 2018-08-21
Rename simple-dataplane to test-dataplane

mestery:
To better reflect the reality the in-tree dataplane was merely for testing
purposes in our CI system, rename it to `test-dataplane`. This will hopefully
avoid any confusion around whether or not you should run this anywhere near a
production system (you shouldn't), and what it is used for (it's used for
testing only).


mestery:
@fkautz Per our discussion in IRC yesterday, pushing this out to try and rename it before OSS next week.

PR #293, opened 2018-09-19
Adding NetworkServiceWiring doc

matrohon:


matrohon:
@edwarnicke @mestery @sbezverk this is my understanding of the NSW usage in NSM, please correct me if I'm wrong.

docs/network-service-wiring.md
```diff
@@ -0,0 +1,72 @@
+This document aims at identifying the workflow a network service provider (NSP) would use to manage network services (NS) with some network service wirings (NSW).
+
+The use case is a NSP who wants to provide a gold-network NS, compose by a vpngateway at the begining, and then by a firewall and a vpngateway.
```
edwarnicke: This is a good start, might be useful to have a picture of the intended state (people like pictures :) ).

docs/network-service-wiring.md
```diff
@@ -0,0 +1,72 @@
...
+
+To provide the NS, the NSP will first have to create the NS with the following k8s object creation:
```
edwarnicke: Some section breaks might be helpful to the reader.  For example, you could consider a structure like:

- The Problem the SP is trying to solve
- Step the SP takes to solve it in an NSM world
       - Deploy a Network Service
       - Deploy Network Service Wirings
       - Deploy NSEs
       - Deploy consuming Pods

(not necessarily exactly that, but that kind of thing :) ).

matrohon: ok, but I'd like to be sure about what exact steps the SP will follow, since it will probably have side-effect, as mentioned below, and impact on implementation details (datamodels and NSM components communications at least).

What I try to highlight in this doc is how the NSP can __dynamically__ manage the traffic flow inside the NS.

matrohon:
I've added an alternative section in the latest commit, but I'm not sure this is the good place to talk about alternatvies

edwarnicke:
I think its a great place to *start* the conversation :)

I started writing a bit of a response, and realized it would help if I drew a picture, so hope to get a more detailed response soon.

The net-net is: your solution *is* simpler for the use cases used as examples in this doc (and the slide decks) thus far.  There are some really fun and interesting use cases beyond those that I think you will like that it works less well for.  Thus the need to get those use cases into the light, so we can talk about the tradeoffs :)

PR #327, opened 2018-10-11
WIP Circle CI and Packet.net

platoff:
WIP Circle CI and Packet.net


scripts/travis-integration-tests.sh
```diff
@@ -51,6 +53,7 @@ kubectl() {
 # run_tests returns an error on failure
 run_tests
 exit_code=$?
+delete_k8s_cluster "${TEST_CONTEXT}"
```
mestery: Can you implement delete_k8s_cluster for the other types (dind and minikube)? That way this will continue to work for those as well.

scripts/cluster_common.sh
```diff
@@ -24,9 +24,16 @@ create_k8s_cluster() {
             echo "FATAL: bringing up k8s cluster '${ctx}' not supported"
             exit 255
         }
-        ${cluster_up}
+        . ${cluster_up}
```
mestery: Why is this needed?

scripts/cluster-up-packet.sh
```diff
@@ -0,0 +1,55 @@
...
+CLOUD_DEPLOYMENT_DATA=$(pwd)/.data
+
+docker run \
+  -v ${CLOUD_DEPLOYMENT_DATA}:/cncf/data \
+  --dns 147.75.69.23 --dns 8.8.8.8 \
+  -e NAME=${CLOUD_DEPLOYMENT_NAME} \
+  -e CLOUD=packet    \
+  -e COMMAND=deploy \
+  -e BACKEND=file  \
+  -e PACKET_AUTH_TOKEN=${PACKET_AUTH_TOKEN} \
```
mestery: Maybe add a note that PACKET_AUTH_TOKEN and PACKET_PROJECT_ID are defined in the circle-ci dashboard? May be worth a broader note that these would need to be defined before this could be run by hand.

scripts/cluster_common.sh
platoff: I do not like it... Thinking on this now, basically `cluster_up` may set some variables, which require later, at the moment it's KUBECONFIG, do you think this is a case?

scripts/cluster-up-packet.sh
platoff: Good point will add it

scripts/travis-integration-tests.sh
platoff: It should work for other types, delete_k8s_cluster just call "cluster-down-xxx.sh" script if it's present.

scripts/cluster_common.sh
platoff: For my particular case and to be consistent with everything else I can assume that KUBECONFIG is at default ~/.kube/config location and remove this change

mestery: That makes sense, thanks!

scripts/travis-integration-tests.sh
mestery: I missed that, so you are right, this will work ok. Thanks!

.circleci/config.yml
```diff
@@ -5,10 +5,15 @@ version: 2
...
     steps:
-      - checkout # check out the code in the project directory
-      - run: echo "hello world" # run the `echo` command
+      - checkout
+      - setup_remote_docker
+      - run: sudo apt-get install yamllint shellcheck python3-pkg-resources
+      - run: go install ./vendor/k8s.io/kube-openapi/cmd/openapi-gen
+      - run: make dep-check
+      - run: yamllint -c .yamllint.yml $(git ls-files '*.yaml' '*.yml' | grep -v 'vendor/')
+      - run: make all && ./scripts/travis-integration-tests.sh packet
```
mestery: By doing things in this order, you are building the images in the circleci Docker, and then creating the K8S cluster on packet.net. That will then pull the images from Docker Hub and test against those, not the built images from the PR it's supposed to be testing. We hit the same thing with travis-ci and just merged something to transfer the images between the two systems. But in this case, I wonder if it's better to just build the images in the Docker on packet.net? Either way, this needs to be addressed.

platoff: Currently, travis-ci push every successfully built image to registry... What about to push images always, and pull corresponding image from Docker Hub to use in integration tests?

.circleci/config.yml
```diff
@@ -1,14 +1,49 @@
...
-    docker: # use the docker executor type; machine and macos executors are also supported
-      - image: circleci/golang:1.9
+    <<: *defaults
+    steps:
+      - checkout
+      - setup_remote_docker
+      - run: echo "export COMMIT=${CIRCLE_SHA1::8}" >> $BASH_ENV
+      - run: echo "COMMIT=$COMMIT"
+      - run: go install ./vendor/k8s.io/kube-openapi/cmd/openapi-gen
+      - run: make verify docker-build docker-push
```
mestery: @platoff Lets not do the docker push from here just yet, first lets get the build working, and then lets decide when we want to shift docker push to here instead of the travis build. Otherwise we'll be pushing twice, but also lets figure out if we want to do docker push from this build or from travis-ci.

platoff: The last step I need to implement is to make freshly built images available on deployed k8s cluster. I want to make them available via docker hub, e.g. push freshly built images and let k8s pull them for integration tests. There should not be conflicts with Travis. Travis push images tagged as below:

```
if [ "${TRAVIS_PULL_REQUEST}" == "false" ]; then export TAG=`if [ "${TRAVIS_BRANCH}" == "master" ]; then echo "latest"; else echo ${TRAVIS_BRANCH}; fi` && make docker-push ; fi
```

I can provide CircleCI's generated tags which would not affect existing infrastructure and do not break anything. Thoughts?

platoff: @mestery Obviously, there are other ways to make freshly build images available on cluster deployed in the cloud, but I believe this way is most portable (consider deployment to GKE, etc where we can't potentially ssh into, or may have problems build docker images and/or make them available to a node). As for tagging, Circle CI will not use `latest` as well as we can use bytes 8-16 of commit sha not to interfere with travis, which use bytes 0-8

edwarnicke: My thoughts are: on the plus side, pushing the images to somewhere like docker hub with a CircleCI generated tag would allow us to pull them from not only Packet, but also from other cloud provides like AWS/GCE/Azure etc as we get them up and online.

@mestery What are your thoughts on @platoff 's idea?

mestery: Guys, my point is Travis-ci is already doing the push, and we need to coordinate this while we transition. The other thought is we have the packet.net stuff use the Travis artifacts for now. But clearly coordination and how we want to use both packet.net and Travis is needed before this is merged.

fkautz:
My preference would be for PRs to push to a private repo with a tag and merged patches in master push to a public repo once it passes CI.

fkautz:
I'll spend some time to help provide better guidance on this today (Saturday).

fkautz:
Ok instead of a private repo, let's have two public repos on docker. First is for CI, all CI images go there.

Second is for releases which is the official one. We can push latest master and stable there, keeps it unpolluted.

.circleci/config.yml
```diff
@@ -1,14 +1,76 @@
...
-    docker: # use the docker executor type; machine and macos executors are also supported
-      - image: circleci/golang:1.9
+  sanity-check:
+    working_directory: /go/src/github.com/ligato/networkservicemesh
+    docker:
+      - image: circleci/golang:1.10
+    steps:
+      - checkout
+      - run: sudo apt-get install yamllint shellcheck python3-pkg-resources
+      - run: yamllint -c .yamllint.yml $(git ls-files '*.yaml' '*.yml' | grep -v 'vendor/')
```
mestery: If these sanity-checks fail, does the build itself fail as well and stop here? I assume it does, just making sure, because this is the behavior we will want.

scripts/travis-integration-tests.sh
```diff
@@ -19,15 +19,35 @@ set -xe
...
+
+export K8S_DEPLOYMENT_NAME
+export K8S_DEPLOYMENT_DATA
+
 KUBECTL_VERSION=v1.11.3
 
 # Install kubectl
 curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/"${KUBECTL_VERSION}"/bin/linux/amd64/kubectl && \
-	chmod +x "kubectl" && sudo mv "kubectl" /usr/local/bin/
+ 	chmod +x "kubectl" && sudo mv "kubectl" /usr/local/bin/
```
mestery: Is this a spurious change?

scripts/travis-integration-tests.sh
```diff
@@ -19,15 +19,35 @@ set -xe
 
 # Default kubernetes context - if it's "dind" or "minikube" will
 # try to bring up a local (dockerized) cluster
+# Other kubernetes contexts like "packet", "aws", etc will bring
+# up a cluster in the cloud
+#
+# For cloud deployments, cloud-specific environment variables
+# must be set as below:
+#
+# Packet.net:
```
mestery: Maybe remove these here since this fail appears to be for travis-ci?

scripts/ci-cluster-provision.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
mestery: When does this file get called? Is this called from the cross-cloud-ci scripts itself somehow?

scripts/ci-cluster-provision.sh
```diff
@@ -0,0 +1,23 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This is a wrapper for cross-cloud's provision.sh
+# targeted to use in Circle CI environment
+
+export TF_VAR_packet_project_id="${PACKET_PROJECT_ID}"
+pushd /cncf
+./provision.sh "$1" nsm-ci-"${CIRCLE_WORKFLOW_ID:0:8}" file
```
mestery: Does this imply that we have a maximum of 8 concurrent clusters to test against?

mestery:
@platoff I've exported the following variables in the circle-ci config:

```
DOCKER_EMAIL
DOCKER_USERNAME
DOCKER_PASSWORD
```

scripts/ci-packet-deploy.sh
```diff
@@ -0,0 +1,54 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# The purpose of the script is to deploy k8s cluster in the packet.net
+# PACKET_AUTH_TOKEN and PACKET_PROJECT_ID MUST be provided externally
+
+K8S_DEPLOYMENT_NAME=nsm-ci-"$CIRCLE_BUILD_NUM"
+
+docker run \
```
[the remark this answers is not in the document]
sbezverk: Right, currently multinode functionality is not developed, well it is not even designed as we do not have yet multinode api. For now a single node cluster is the only working and tested scenario.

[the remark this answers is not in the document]
platoff: Currently, we're using the default cross-cloud ci configuration, which is 3 master, 1 worker node set up.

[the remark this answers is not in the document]
edwarnicke: @mestery The other option would be to simply pin the NSE and NSC to the same node for the moment.  Likely simpler than hacking at the Cross Cloud CI scripts.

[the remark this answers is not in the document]
edwarnicke: @mestery Also... pinning the NSE and NSC to the same node would make it simpler when we start testing for multi-node, because then the Packet.net CI stuff doesn't have to change.  Should make things simpler for the folks implementing and testing the multi-node functionality as they don't have to understand how to change the Packet.net CI to get more than one worker node.

Another possibility might be to simply taint 2 of the three workers... lots of ways to approach the issue :)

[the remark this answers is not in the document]
edwarnicke: @mestery And I just read @platoff 's comments and felt silly at commenting on a current non-issue ;)

mestery:
@platoff Notice that shellcheck has an error when running on circle-ci, see [here](https://circleci.com/gh/ligato/networkservicemesh/193?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) for an example of this failure.

This appears to be caused by a newer shellcheck and an older Linux kernel. See [this issue](https://github.com/koalaman/shellcheck/issues/785) for an example of this.

I opened #345 to track this issue.

.circleci/config.yml
```diff
@@ -1,14 +1,76 @@
...
     working_directory: /go/src/github.com/ligato/networkservicemesh
+    docker:
+      - image: circleci/golang:1.10
+    steps:
+      - checkout
+      - setup_remote_docker
+      - run: go install ./vendor/k8s.io/kube-openapi/cmd/openapi-gen
+      - run: make verify docker-build
 
+  packet-deploy:
```
mestery: I looked at the circle-ci dashboard, and it might be good to dump some information on the Kubernetes cluster after it's deployed. Since we do not have access to the logs after this run is done, the more information we can save here the easier it will be to debug these. For example, the items we collect after the integration tests [here](https://github.com/ligato/networkservicemesh/blob/master/scripts/integration-tests.sh#L138) would be useful. And it's worth collecting that information after deploy in case the deploy itself fails somehow.

I'm fine with an issue being opened and this being fixed after this merges however.

edwarnicke: Yeah, dumping K8s logs to CircleCI, particular on failure is going to be huge for debuggability :)

.travis-disable.yml
```diff
@@ -28,7 +28,7 @@ script:
 
 # We only want to push docker images on a push to master, not a pull request
 after_success:
-  - if [ "${TRAVIS_PULL_REQUEST}" == "false" ]; then export TAG=`if [ "${TRAVIS_BRANCH}" == "master" ]; then echo "latest"; else echo ${TRAVIS_BRANCH}; fi` && make docker-push ; fi
+  - if [ "${TRAVIS_PULL_REQUEST}" == "false" ]; then export TAG=`if [ "${TRAVIS_BRANCH}" == "master" ]; then echo "latest"; else echo ${TRAVIS_BRANCH}; fi` && export BUILD_TAG=travis-"${TRAVIS_BUILD_NUMBER}" && make docker-push ; fi
```
mestery: Love this! <3

.circleci/config.yml
platoff: Yes, if sanity-check fail the build itself will fail and stop here.

mestery:
@platoff This is looking pretty good! I suggest a rebase and we can pretty much merge this and then fix any issues.

I do have one question: Is there some sort of limit around how many concurrent deploys into packet.net we can have, and if so, is that set somewhere? I only wonder about a "denial-of-service" attack on packet.net via our CI infra here.

.circleci/config.yml
platoff: circle-ci run same integration-tests.sh script and will collect same logs after integration tests run. Going to add some log collection on deploy phase as well.

platoff:
@mestery There is no any kind of limit of concurrent builds in CI code, hope we have some limits with Packet.net account that would prevent denial-of-service, also is there any way to limit concurrent executions at CircleCI level?

edwarnicke:
@platoff CircleCI has a limit on the number of free executors that can run concurrently.  That should effectively avoid a serious DOS attack :)

PR #333, opened 2018-10-12
Skip the netmesh-test image and remove some debug

mestery:


mestery:
@sbezverk same netns failure so this can be merged as we debug that failure.

PR #337, opened 2018-10-14
Rename dataplaneinterface -> dataplane

edwarnicke:
This is build on PR #336.  Only the last commit in the chain
belongs to this PR.


.travis.yml
```diff
@@ -24,11 +24,14 @@ script:
   - go install ./vendor/k8s.io/kube-openapi/cmd/openapi-gen
   - make dep-check
   - yamllint -c .yamllint.yml $(git ls-files '*.yaml' '*.yml' | grep -v 'vendor/')
+  - cd dataplanes/vpp;make all; cd -
```
mestery: This may conflict since it merged as a part of #336, as well as some other changes below.

PR #352, opened 2018-10-17
Add dataplane build into CircleCI scripts

mestery:
Also cleanup the formatting of the dataplane build in .travis.yml.


mestery:
@edwarnicke This patch shows that the Dataplane was not being built in CircleCI. Still, it should have been built in travis-ci and failed there, so lets see if the travis-ci build fails or not.

edwarnicke:
@mestery #353 should fix the build breakage correctly

As to building dataplane in CircleCI, its a good thing, I suspect it hit a ships-in-the-night effect as it wasn't being built in TravisCI when CircleCI copied the build logic over, so good to fix that here.

PR #354, opened 2018-10-17
Make packet-destroy more stable

edwarnicke:
packet-destroy as part of the packet-integraion-tests job
was slow and unstable, due to issue with pulling
the provision docker image from the remote_docker
being very very slow (sometimes taking 30 minutes, or
timing out).

This same docker image, when used as the basis for a standalone
job (as is done for packet-deploy) is zippy and reliable.

This patch breaks packet-destroy out into a separate job.
In the event that integration-tests fails, it will use
the CircleCI API to launch the packet-destroy job.

This is done by storing the /cncf/data needed in a cache
with key cncf-data-{{.Environment.CIRCLE_WORKFLOW_ID}}
and having the API set CIRCLE_WORKFLOW_ID in the packet-destroy
job to match.

This has been tested and works.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -47,10 +47,11 @@ jobs:
           command: kubectl create -f /cncf/rbac/
           environment:
             KUBECONFIG: /cncf/data/kubeconfig
-      - persist_to_workspace:
-          root: /cncf/data
+      - run: find /cncf/data
+      - save_cache:
+          key: cncf-data-{{.Environment.CIRCLE_WORKFLOW_ID}}
           paths:
-            - ./*
+            - data
 
   packet-integration-tests:
     working_directory: /go/src/github.com/ligato/networkservicemesh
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -62,15 +63,49 @@ jobs:
...
+          key: cncf-data-{{.Environment.CIRCLE_WORKFLOW_ID}}
+      - run: find /cncf/data
+      - checkout:
+          path:
+            nsm
+      - run: nsm/.circleci/set-nameserver.sh
+      - run: nsm/.circleci/provision.sh packet-destroy
+      - run:
+          when: on_fail
+          command: nsm/.circleci/provision.sh packet-destroy
```
platoff: @edwarnicke should we remove on_fail -> packet_destroy here (in packet-destroy job)?

edwarnicke: @platoff I'm honestly OK with either keeping it or leaving it.  My thinking in leaving it had been that in the event of some kind of problem destroying the cluster, it would 'retry' destroy.  That said, running from the provision container for destroy seems super reliable, so it may not be necessary.

Do you have any strong opinions here?

platoff: @edwarnicke +1 for the purpose of 'retry' destroy.

.circleci/config.yml
```diff
@@ -47,10 +47,11 @@ jobs:
           command: kubectl create -f /cncf/rbac/
           environment:
             KUBECONFIG: /cncf/data/kubeconfig
-      - persist_to_workspace:
-          root: /cncf/data
+      - run: find /cncf/data
```
mestery: Are these `find /cncf/data` simply for debug purposes?

.circleci/config.yml
```diff
@@ -47,10 +47,11 @@ jobs:
...
+      - save_cache:
+          key: cncf-data-{{.Environment.CIRCLE_WORKFLOW_ID}}
           paths:
-            - ./*
+            - data
```
mestery: For my own understanding, is this saving `cncf-data-{{.Environment.CIRCLE_WORKFLOW_ID}}` in the directory `data`? If so, should we put an absolute path here rather than what looks to be a relative path?

.circleci/config.yml
```diff
@@ -63,15 +64,49 @@ jobs:
...
             sudo apt-get install gettext-base
-      - attach_workspace:
-          at: ./data
+      - run:
+          command: |
+            sudo mkdir -p /cncf/data
+            sudo chown -R circleci:circleci /cncf/
+          name: Cache Prep
+      - restore_cache:
+          key: cncf-data-{{.Environment.CIRCLE_WORKFLOW_ID}}
```
mestery: This is restored from a CircleCI internal cache? Is it restoring a directory?

.circleci/config.yml
```diff
@@ -63,15 +64,49 @@ jobs:
...
           environment:
             KUBECONFIG: /go/src/github.com/ligato/networkservicemesh/data/kubeconfig
       - run:
-          when: always
-          command: ./.circleci/destroy-cluster.sh
+          when: on_fail
+          name: Trigger packet-destroy
+          command: |
+            curl --user ${CIRCLE_API_PROJECT_TOKEN}: \
+                --data build_parameters[CIRCLE_JOB]=packet-destroy \
```
mestery: This is fun :)

.circleci/config.yml
edwarnicke: You are correct, let me remove them.

.circleci/config.yml
edwarnicke: Not quite ;)  For reference, the [CircleCI Caching docs](https://circleci.com/docs/2.0/caching/) are pretty good, but here's the tl;dr version :)
```cncf-data-{{.Environment.CIRCLE_WORKFLOW_ID}}``` is the cache key.  You can write to a cache key precisely *once*.  Its write once, and then read only.

The ```{{.Environment.CIRCLE_WORKFLOW_ID}}``` expands to the environment variable CIRCLE_WORKFLOW_ID, which has the id for this workflow run.  paths tells us which directories to put into the 'cache' for that cache key.  So it goes and grabs everything in directory (relative to the working dir, which is /cncf/ and puts it in the cache under the cache key cncf-data-{{.Environment.CIRCLE_WORKFLOW_ID}}

Then when we use the CircleCI API via curl later we set the CIRCLE_WORKFLOW_ID, and it will look up the correct cache we put aside here.  The /cncf/data dir has all the info needed to clean up in packet-destroy.

Make sense?

.circleci/config.yml
edwarnicke: It is restoring from the cache.  ```cncf-data-{{.Environment.CIRCLE_WORKFLOW_ID}}``` is the cache key to be restored.  It restores the directory originally cached, which was /cncf/data

.circleci/config.yml
edwarnicke: Super fun!

PR #356, opened 2018-10-18
WIP: Transition to new Dataplane API

edwarnicke:
This PR is to Transition NSM to using the new Dataplane API

Current plan is:

1. Store Connections in ObjectStore
2. Rework nsmserver_client and finalizer to use the new Dataplane API
3.  Finish implementing the new API in testdataplane (its there, but just stubbed)
4.  Make sure that all passes CI
5.  Start doing additional CI on Packet for the new Dataplane


edwarnicke:
Based on conversation with @sbezverk we should only be putting Connection in the Objectstore if its to be saved between plugins, which I suspect is not the case, so closing this PR.

PR #361, opened 2018-10-20
Experiment: Mock k8s API and decouple NSM implementation

platoff:
This is an experiment and probably not good idea, however, I want to try to mock k8s API which let to run NSM and components in an environment which simulate k8s cluster (at the level required by NSM components to function properly)...

## Why?
Potential benefits are (if the experiment succeed) :
 - increased development time: many parts can be executed and debugged on the local machine without docker and k8s 
 - simplified testing (both unit and integration): a huge amount of tests may be written without cluster set up; since this experiment is an attempt to decouple code in some form, decoupling leads to easier testing (both unit and integration)
 - because of decoupling, many functional components that require real environment (e.g. docker container) should be fine-grained, which leads to the more structured codebase and can be tested separately

## Risks
 - abstracting stuff may result in more work to maintain codebase and void benefits above
 - it would not be easy to reach experiment goal
 - I do not see a whole picture and this is a really bad idea :)

Anyway, I'd like to create this PR to use CI environment this weekend and see how this may progress

### DIFF
--- a/plugins/crd/crd_utils.go
+++ b/plugins/crd/crd_utils.go
@@ -73,7 +73,7 @@ func newCustomResourceDefinition(plugin *Plugin, FullName, Group, Version, Plura
 	return nil
 }
 
-func createCRDObject(newCRD *apiextv1beta1.CustomResourceDefinition, crdClient *apiextcs.Clientset) error {
+func createCRDObject(newCRD *apiextv1beta1.CustomResourceDefinition, crdClient apiextcs.Interface) error {
 	// First check if the CRD already exists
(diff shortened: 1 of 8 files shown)


edwarnicke:
@platoff I like the idea in general, as it would let us write unit tests for things like the NSM that talk to K8s.

@fkautz @sbezverk @mestery Thoughts?

mestery:
This is the right move because it abstracts away our dependency on Kubernetes and is moving us in the right direction of being able to integrate with other orchestrators such as Nomad. I'm 👍 for this direction and work.

PR #362, opened 2018-10-20
Add a cleanup step to build CircleCI Job

edwarnicke:
This is far from perfect, as it basically waits
10 minutes for packet-deploy to finish before triggering
packet-destroy.  I hate sleeps in tests... but its the best
we can do.

This has been tested:

1.  If the build is broken, cleanup happens correctly.
2.  If the build is broken *and* packet-deploy is broken,
    the packet-destroy triggered by the broken build
    does nothing, and reports success unless packet-deploy
    has completed.

So the only available mode in which this doesn't behave well is:

packet-deploy succeeds more than 10 minutes after the build breaks.
In that case we can leak servers.

The additional drawback is that build fails will take a bit longer to report.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,6 +29,24 @@ jobs:
             go install ./vendor/k8s.io/kube-openapi/cmd/openapi-gen
             cd dataplanes/vpp ; make all ; cd -
             make verify docker-build docker-push
+      - run:
+          when: on_fail
+          name: Trigger packet-destroy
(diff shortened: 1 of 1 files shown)


platoff:
There is 25-minute deploy timeout in cross-cloud CI scripts, so we may expect packet-deploy to be succeeded in 30 minutes max. Maybe increasing wait time from 10 minutes to 15 would make things even more stable.

PR #366, opened 2018-10-22
NSM client and endpoint connecting over memif interface

platoff:
This is a boilerplate to implement nsm client connecting icmp-responder endpoint over memif interface


conf/sample/nsc-memif.yaml
```diff
@@ -0,0 +1,104 @@
...
+            limits:
+              nsm.ligato.io/socket: 1
+          env:
+            - name: INIT_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+      volumes:
+        - hostPath:
+            path: /var/lib/networkservicemesh
```
edwarnicke: We will need to get per-Pod directories injected into NSCs and NSEs rather than mounting the whole NSM directory to ensure proper isolation and security.

conf/sample/nse-memif.yaml
```diff
@@ -0,0 +1,71 @@
...
+      containers:
+        - name: nse-1
+          image: networkservicemesh/nse-memif
+          imagePullPolicy: IfNotPresent
+          env:
+            - name: INIT_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
+          volumeMounts:
```
edwarnicke: We will need to get per-Pod directories injected into NSCs and NSEs rather than mounting the whole NSM directory to ensure proper isolation and security.

PR #364, opened 2018-10-21
Reworked Makefiles to use pattern rules

edwarnicke:

PR #367, opened 2018-10-22
Remove requirement for hugepages for vpp container

edwarnicke:
VPP only *requires* hugepages in for HW (via DPDK for example)

    If a vpp dataplane is not using DPDK, it doesn't need hugepages.

    Hugepages does require additional work (and additional possibility for
    error).  So temporarily removing it by:

    1)  Disabling the DPDK plugin by copying in a new startup.conf
    2)  Removin the hugepages resource requirement from vpp.yaml

This is built on PR #364 and only the last patch is part of this PR


edwarnicke:
Please hold off on merging... still testing :)

edwarnicke:
OK, tested and ready to go :) (see PR #368 for how I tested :) )

dataplanes/vpp/conf/startup.conf
```diff
@@ -0,0 +1,156 @@
+
```
mestery: Pedantic but important nit: In the interest of making this easier to understand for new contributors, we may want to put this in the directory `dataplanes/vpp/conf/vpp/startup.conf` so it's obvious to new contributors this is a VPP startup configuration file.

PR #369, opened 2018-10-23
Add build of dataplane/vpp to CircleCI

edwarnicke:


.circleci/config.yml
```diff
@@ -29,6 +29,7 @@ jobs:
             go install ./vendor/k8s.io/kube-openapi/cmd/openapi-gen
             cd dataplanes/vpp ; make all ; cd -
             make verify docker-build docker-push
+            cd dataplanes/vpp;make verify docker-build docker-push;cd -
```
fkautz: We should switch this to pushd and popd for robustness.

PR #374, opened 2018-10-23
Make integration tests working with vpp-dataplane

platoff:
This change makes integration tests work both with test-dataplane and vpp-dataplane via new dataplane API. 

Just replace `kubectl create -f conf/sample/test-dataplane.yaml` with `kubectl create -f dataplanes/vpp/vpp-daemonset.yaml`

Note: NSM use hardcoded addresses (as supposed by current integration tests), @edwarnicke please update me with the proper way to make current sample working without hardcoded addresses: how this should be configured and negotiated among client, nse, and nsm in the current sample?


scripts/integration-tests.sh
```diff
@@ -66,7 +68,7 @@ function run_tests() {
     #
     # Now let's wait for nsm-cient pod to get into running state
     #
-    wait_for_pods default
+    wait_for_pods default 360
```
edwarnicke: The 360 in this line seems to be breaking CI.

conf/sample/networkservice-daemonset.yaml
```diff
@@ -28,6 +29,10 @@ spec:
               mountPath: /var/lib/kubelet/device-plugins
             - name: nsm-socket
               mountPath: /var/lib/networkservicemesh
+            - name: run
+              mountPath: /var/run
+            - name: sys
+              mountPath: /sys
```
edwarnicke: Why are we adding these very broad mounts to the NSM container?

conf/sample/networkservice-daemonset.tmpl.yaml
```diff
@@ -28,6 +29,10 @@ spec:
               mountPath: /var/lib/kubelet/device-plugins
             - name: nsm-socket
               mountPath: /var/lib/networkservicemesh
+            - name: run
+              mountPath: /var/run
+            - name: sys
+              mountPath: /sys
```
edwarnicke: Why are we adding these very broad mounts?

conf/sample/networkservice-daemonset.yaml
platoff: My understanding that dataplanes are orchestrator-agnostic, and does not know anything about k8s (and probably) nothing about docker. If my understanding is correct, it's NSM responsibility to provide pids/linux namespaces of pods to be interconnected down to dataplane. In order to get conainer pids we have to have access /var/run/docker.pid and /sys, I'll check if /sys access can be narrower, but anyway either dataplane of NSM should have access to these paths. Please suggest

edwarnicke: OK... I'm not sure whey we need to be able to get the docker PIDs... what are we doing with them?

edwarnicke: OK... I think I get the pid thing, but we are getting the netns from the Pid via the netns lib... which did not appear to require mounting /sys before... so I'm wondering we we need it now?

PR #377, opened 2018-10-24
Extending Reply with Connection property

lobkovilya:


edwarnicke:
Could you rerun:
scripts/update-codegen.sh

and commit?  That should update things to get you past the sanity check.

edwarnicke:
Looks like you need a rebase because #357 was merged

edwarnicke:
Could you please do this as a rebase rather than merging master in?  Merging master in results in a bunch of extra files in the change and a more confusing history.  It should be the case that a rebase here would just result in one patch, that is doing the one thing of Extending Reply with Connection property... this patch has a lot of other stuff going on.

lobkovilya:
Ok, I'll fix, but I think now is easier to create new PR and remove this one :)

edwarnicke:
@lobkovilya That's cool.  If creating a new PR is easier for you, please do close this one and create a new one :)

PR #379, opened 2018-10-25
Extend Reply with Connection property

lobkovilya:


edwarnicke:
As soon as CI passes, I'm good with this :)

PR #381, opened 2018-10-25
Traverse /proc/*/ns/net for specified inode

platoff:
Code to traverse /proc/*/ns/net for specified inode


edwarnicke:
I'm good with this once CI passes.

PR #383, opened 2018-10-25
Fix vpp version to 1807 and collect vpp logs properly

platoff:
This PR fixes vpp version to 1807 to match current bindings and introduce change to collect vpp logs properly

### DIFF
--- a/dataplanes/vpp/build/Dockerfile.vpp
+++ b/dataplanes/vpp/build/Dockerfile.vpp
@@ -1,6 +1,6 @@
 FROM ubuntu:bionic
 ARG DEBIAN_FRONTEND=noninteractive
-ARG REPO=release
+ARG REPO=1807
 RUN apt-get update
 RUN apt-get -y install curl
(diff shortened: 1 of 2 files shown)


dataplanes/vpp/build/Dockerfile.vpp
```diff
@@ -1,6 +1,6 @@
 FROM ubuntu:bionic
 ARG DEBIAN_FRONTEND=noninteractive
-ARG REPO=release
+ARG REPO=1807
```
edwarnicke: Don't fix the repo.  You still want the release repo.  Instead use"

```bash
apt-get install <package name>=<version>
```
Where version is 18.07.1

platoff: Oh, ok, thank you

edwarnicke: The fdio/1807 repo tracks the packages from the stable/1807 branch... which has moved on since the last 18.07.1 release ( which is in fdio/release ) :)

PR #387, opened 2018-10-26
Implement memif to memif local connection, add unit tests

lobkovilya:


edwarnicke:
@lobkovilya You appear to have a conflict, you may want to rebase.

dataplanes/vpp/pkg/nsmutils/nsm-utils.go
```diff
@@ -91,3 +99,13 @@ func Ipv4prefixlength(value string) error {
 	}
 	return nil
 }
+
+func Empty(value string) error {
```
platoff: I think there is `Any` validator function already, which allow any value to be specified as a parameter

PR #390, opened 2018-10-28
Make connection working with inode namespace parameters

platoff:
Current vpp-daemon implementation accepts namespaces in form of pid:xxxx only. This change makes it working with inode values provided for namespaces.

### DIFF
--- a/dataplanes/vpp/pkg/nsmutils/nsm-utils.go
+++ b/dataplanes/vpp/pkg/nsmutils/nsm-utils.go
@@ -67,11 +67,7 @@ func ValidateParameters(parameters map[string]string, keyList Keys) error {
 }
 
 // Keys validator functions, for each new Keys there should be a validator function.
-func Namespace(value string) error {
-	// This is not true anymore
-	// if !strings.HasPrefix(value, "pid:") {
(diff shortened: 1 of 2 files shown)


dataplanes/vpp/pkg/nsmutils/nsm-utils.go
```diff
@@ -69,9 +68,10 @@ func ValidateParameters(parameters map[string]string, keyList Keys) error {
 
 // Keys validator functions, for each new Keys there should be a validator function.
 func Namespace(value string) error {
-	if !strings.HasPrefix(value, "pid:") {
-		return fmt.Errorf("malformed namespace %s, must start with \"pid:\" following by the process id of a container", value)
-	}
+	// This is not true anymore
+	// if !strings.HasPrefix(value, "pid:") {
+	// 	return fmt.Errorf("malformed namespace %s, must start with \"pid:\" following by the process id of a container", value)
+	// }
```
edwarnicke: We should probably either eliminate this function entirely or expand the validator to the new circumstances.

platoff: @edwarnicke Done... current validation mechanism seems to be limited in terms of expressiveness, and kernel-kernel, for example, performs additional validation steps, so I'd better remove this validation in the future in favour of hard-coded verification of parameters

PR #395, opened 2018-10-29
Any-any (extensible) cross-connects in vpp dataplane

platoff:
tested with nsmd, ready for @lobkovilya to add memif stuff... There is more work required here, especially recovery from errors in the middle of cross-connect set up, like if we have one interface up, and second failed.

### DIFF
--- a/dataplanes/vpp/pkg/nsmvpp/nsm_vpp.go
+++ b/dataplanes/vpp/pkg/nsmvpp/nsm_vpp.go
@@ -24,6 +24,7 @@ import (
 	govpp "git.fd.io/govpp.git/core"
 	"github.com/ligato/networkservicemesh/dataplanes/vpp/bin_api/tapv2"
 	"github.com/ligato/networkservicemesh/dataplanes/vpp/bin_api/vpe"
+	"github.com/ligato/networkservicemesh/pkg/nsm/apis/common"
 	"github.com/sirupsen/logrus"
 )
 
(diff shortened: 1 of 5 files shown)


dataplanes/vpp/pkg/nsmvpp/nsm_vpp.go
```diff
@@ -24,17 +24,27 @@ import (
...
-	Validate(parameters map[string]string) error
+type mechanism interface {
+	createInterface(apiCh govppapi.Channel, parameters map[string]string) (uint32, error)
+	deleteInterface(apiCh govppapi.Channel, intf uint32) error
+	validate(parameters map[string]string) error
+}
+
+var mechanisms = []mechanism{
+	kernelMechanism{}, // default
+	kernelMechanism{},
```
edwarnicke: Why do we have kernelMechanism twice/

dataplanes/vpp/pkg/nsmvpp/nsm_vpp_noimpl.go
```diff
@@ -7,19 +7,19 @@ import (
...
 
-type UnimplementedMechanism struct {
+type unimplementedMechanism struct {
 	Type common.LocalMechanismType
 }
 
 // CreateLocalConnect return error for unimplemented mechanism
-func (m UnimplementedMechanism) CreateLocalConnect(apiCh govppapi.Channel, srcParameters, dstParameters map[string]string) (string, error) {
-	return "", fmt.Errorf("%s mechanism not implemented", common.LocalMechanismType_name[int32(m.Type)])
+func (m unimplementedMechanism) createInterface(apiCh govppapi.Channel, parameters map[string]string) (uint32, error) {
```
edwarnicke: Would it make sense to do this with the LocalMechanism or RemoteMechanism object rather than just the parameters map?

platoff: You are right, semantically more correct to go with Local/Remote Mechanisms. Technically at this point, Local/Remote mechanisms are just Type+Parameters, passing Type would be redundant, so until Local/RemoteMechanisms would introduce more information it's simpler to stay with parameters (as it's uniform type)

dataplanes/vpp/pkg/nsmvpp/nsm_vpp.go
edwarnicke: Excellent :)  Thank you for the explanation!

PR #402, opened 2018-10-29
fix Vagrant README nits

matrohon:
Check that /vagrant/images exists to avoid failure of the vagrant
provisioner when 'vagrant up' is run before nsm images are built


edwarnicke:
@matrohon Many thanks 👍

PR #406, opened 2018-10-29
Initial nsmdp implementation

platoff:
nsmdp seems to be working. All components deployed into k8s does not -- it looks like we have problems in vpp-daemon (vpp-daemon communicate with vpp and nsm in some messy way and very likely that code was not tested well).


controlplane/pkg/nsmdapi/nsmd.proto
```diff
@@ -13,12 +13,12 @@ message ClientConnectionRequest {
 message ClientConnectionReply {
     bool accepted = 1;
     string admission_error = 2;
-    string socket_location = 3;
+    string workspace = 3;
```
edwarnicke: I like the word choice of “workspace” a lot

PR #417, opened 2018-10-31
Network Service CRD

fkautz:


k8s/pkg/apis/networkservice/v1/types.go
```diff
@@ -0,0 +1,51 @@
...
+	meta_v1.ObjectMeta `json:"metadata,omitempty"`
+
+	Items []NetworkService `json:"items"`
+}
+
+type NetworkServiceEndpoint struct {
+	meta_v1.TypeMeta   `json:",inline"`
+	meta_v1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   NetworkServiceEndpointSpec `json:"spec"`
```
edwarnicke: Do we want URI here or URL?

k8s/pkg/apis/networkservice/v1/types.go
```diff
@@ -0,0 +1,51 @@
...
+	Status NetworkServiceEndpointStatus `json:"status"`
+}
+
+type NetworkServiceEndpointSpec struct {
+	NetworkServiceName string `json:"networkservicename"`
+	NsmURL string `json:"networkserviceurl"`
```
edwarnicke: Shouldn't the json tag here be nsmurl ?

PR #420, opened 2018-11-01
Add Network Service Registry support

fkautz:


k8s/pkg/apis/networkservice/v1/types.go
```diff
@@ -42,16 +54,20 @@ type NetworkServiceEndpointSpec struct {
 }
 
 type NetworkServiceEndpointStatus struct {
-	State string `json:"state"`
+	LastSeen meta_v1.Time `json:"lastseen"`
```
edwarnicke: Do we want LastSeen for NSEs?  It kind of commits us to updating them regularly, which may scale poorly

fkautz: Let's remove it, we'll need to think about the lifecycle of a NSE over time.

PR #426, opened 2018-11-06
Adding simple icmp-responder example

fkautz:


iawells:
Your title is quite straightforward, but what is missing anywhere in the commit or comment is 'why would I need a simple ICMP responder'?

fkautz:
I've moved the nse to an examples directory to help make it obvious that this isn't a core component. Thanks for the feedback!

PR #428, opened 2018-11-06
WIP: vppagent dataplane

edwarnicke:


dataplane/vppagent/pkg/vppagent/converter.go
```diff
@@ -0,0 +1,80 @@
...
+		logrus.Infof("Failed with Connection Converter: %s: %s", converter.Name(), err)
+	}
+	return nil
+}
+
+func NewMechanismConverter(c *dataplane.Connection, s SrcDst) Converter {
+	for _, converterFactory := range mechanismConverterFactories {
+		converter := converterFactory(c, s)
+		logrus.Infof("Attempting Mechanism Converter: %s", converter.Name())
+		err := converter.Validate()
```
lobkovilya: I think converter.Validate() will always primarily check the type of local mechanism. Maybe it's better to store mechanism factories in a map (type -> mechanism factory)?

edwarnicke: Good point.  Will look at that on follow up :)

dataplane/vppagent/pkg/vppagent/converter.go
```diff
@@ -0,0 +1,80 @@
...
+	Validate() error
+	FullySpecify() error
+	ToDataRequest(*rpc.DataRequest) (*rpc.DataRequest, error)
+}
+
+type ConnectionConverterFactory func(*dataplane.Connection) Converter
+type MechanismConverterFactory func(*dataplane.Connection, SrcDst) Converter
+
+var connectionConverterFactories = []ConnectionConverterFactory{
+	// TODO: MemifDirectConnectionConverter,
```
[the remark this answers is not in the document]
edwarnicke: You are converting the *Connection*... but you are right, we need a way that doesn’t involve DataRequests for that one

PR #431, opened 2018-11-06
Debug in kubernetes

haiodo:
Adds support for building debug images with dlv debug enabled.


edwarnicke:
This is very good, could we get it working for the containers in controlplane/ dataplane/ ?

haiodo:
Yes, will add tomorrow.

PR #434, opened 2018-11-07
Implement RegisterNSE for nsmd instead of AdvertiseEndpoint

lobkovilya:


controlplane/pkg/nsmd/endpoints.go
```diff
@@ -46,41 +48,27 @@ type nsmEndpointServer struct {
...
 	// success will be returned to NSE, since it is a case of NSE pod coming back up.
-	ep := e.model.GetEndpoint(endpointName)
+	ep := es.model.GetEndpoint(request.EndpointName)
 	if ep != nil {
-		logrus.Warnf("Network Service Endpoint object %s already exists", endpointName)
-		return &nseconnect.EndpointAdvertiseReply{
-			RequestId: ar.RequestId,
-			Accepted:  true,
-		}, nil
+		return nil, fmt.Errorf("Network Service Endpoint object %s already exists", request.EndpointName)
```
edwarnicke: This doesn't seem to match your comments as it returns an error if we have already seen the NSE registration.

fkautz:
Not fully comfortable with the socket being included in the reply, but also recognize that it is the current implementation. Will work to remove it and set a well known socket rather than creating a new one every time in near future work.

PR #436, opened 2018-11-08
Adding upstream register and remove NSE

fkautz:


controlplane/cmd/nsmd/nsmd.go
```diff
@@ -14,11 +17,23 @@ import (
 func main() {
 	model := model.NewModel()
 
+	registryAddress := os.Getenv("K8S_REGISTRY_ADDRESS")
```
fkautz: On second look, I don't like this name. It is K8s specific when it should be NSM specific.

controlplane/cmd/nsmd/nsmd.go
```diff
@@ -14,11 +17,23 @@ import (
...
+	registryAddress = strings.TrimSpace(registryAddress)
+	if registryAddress == "" {
+		registryAddress = "localhost:5000"
+	}
+
+	registryConn, err := grpc.Dial("localhost:5000")
```
edwarnicke: Shouldn’t this be registryAddress rather than a literal?

fkautz: Going to add a defer close here

fkautz: good catch

PR #438, opened 2018-11-08
Vagrant: Install docker-ce 18.06

matrohon:
Ubuntu is now providing docker 18.09 by default.
Kubeadm reports an incompatibility error which prevent k8s
to be installed in the Vagrant VM.


matrohon:
An alternative would be to write a docker installation script, to use with a shell provisioner

PR #441, opened 2018-11-08
Example NSC now uses new grpc api

fkautz:


examples/cmd/nsc/nsc.go
```diff
@@ -46,42 +47,46 @@ func main() {
...
+			ConnectionId:   netns + "-" + strconv.FormatInt(time.Now().Unix(), 36),
+			NetworkService: "icmp-responder",
+			ConnectionContext: &networkservice.ConnectionContext{
+				ConnectionContext: map[string]string{
+					"requires": "src_ip,dst_ip",
+				},
+			},
+			Labels: make(map[string]string),
+		},
+		LocalMechanismPreference: []*common.LocalMechanism{
```
lobkovilya: Does NSM always choose the first LocalMechanism from this array?

edwarnicke: NSM treats this as a preference ordered list.  So it picks the first one it can *do*.  Given where we are at the moment, just pick the first one.

lobkovilya: So NSC gives to NSM a preference ordered list and NSE gets mechanisms from NSM and chooses itself? Is it correct understanding?
I thought you discussed in IRC some kind of symmetry, so I'm little confused :)

edwarnicke: Yes :)

The symmetry here is that the NetworkService API that the NSC uses to address the NSM is the same NetworkService API the NSM uses to address the NSE.

PR #443, opened 2018-11-09
Allow delete dataplane operation

haiodo:
+ Few dataplane model tests

### DIFF
--- a/controlplane/pkg/model/model.go
+++ b/controlplane/pkg/model/model.go
@@ -66,7 +66,7 @@ func (i *impl) AddEndpoint(endpoint *registry.NetworkServiceEndpoint) {
 	serviceName := endpoint.NetworkServiceName
 	services := i.networkServices[serviceName]
 	if services == nil {
-		services = []*registry.NetworkServiceEndpoint{ endpoint }
+		services = []*registry.NetworkServiceEndpoint{endpoint}
 	} else {
 		services = append(services, endpoint)
 	}
@@ -97,7 +97,6 @@ func (i *impl) DeleteEndpoint(name string) error {
 
 		delete(i.endpoints, name)
 	}
-
 
 	return fmt.Errorf("no endpoint with name: %s", name)
 }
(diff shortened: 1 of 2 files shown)


controlplane/pkg/model/model.go
```diff
@@ -103,7 +103,17 @@ func (i *impl) AddDataplane(dataplane *Dataplane) {
 	logrus.Infof("Dataplane added: %v", dataplane)
 }
 
-func (i *impl) DeleteDataplane(name string) {}
+func (i *impl) DeleteDataplane(name string) {
+	i.Lock()
+	defer i.Unlock()
+
+	for idx, dp := range i.dataplanes {
```
edwarnicke: Why not store dataplanes in a map rather than a list so that deleting them is easier?

controlplane/pkg/model/model.go
```diff
@@ -103,7 +103,17 @@ func (i *impl) AddDataplane(dataplane *Dataplane) {
 	logrus.Infof("Dataplane added: %v", dataplane)
 }
 
-func (i *impl) DeleteDataplane(name string) {}
+func (i *impl) DeleteDataplane(name string) {
```
edwarnicke: What triggers the call to DeleteDataplane?

controlplane/pkg/model/model.go
haiodo: Changed to work as maps, for both dataplanes and endpoints.

controlplane/pkg/model/model.go
haiodo: Triggered from un register of data plane and if monitor find connection is down.

controlplane/pkg/model/model.go
edwarnicke: @haiodo Woot!

controlplane/pkg/model/model.go
edwarnicke: Cool

PR #455, opened 2018-11-12
Remove build stage from circleci

fkautz:


.circleci/config.yml
```diff
@@ -255,15 +258,19 @@ workflows:
...
-      - build:
-          requires:
-            - sanity-check
       - packet-deploy:
           requires:
             - sanity-check
       - packet-integration-tests:
           requires:
-            - build
+            - build-nsmd
```
edwarnicke: Do we really require all of these?   Looking below it looks like:

build-nsmd
build-nsmdp
build-nsmd-k8s
build-nsc
build-icmp-responder-nse

PR #466, opened 2018-11-12
Clean up deprecated grpc cancel methods and variable shadowing scope

fkautz:


controllers/sriov-controller/dpapi-controller.go
```diff
@@ -180,8 +180,10 @@ func (s *serviceInstanceController) startServer() error {
 
 	go s.server.Serve(sock)
 
-	// Wait for server to start by launching a blocking connexion
-	conn, err := dial(s.socket, 5*time.Second)
+	// Wait for server to start by launching a blocking connection
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	defer cancel()
```
edwarnicke: So how do we handle cancels?

Ed

fkautz: This should be good enough for now, if something fails underneath, defer cancel() causes a cleanup to occur in grpc.

PR #475, opened 2018-11-15
Wait for port to become available before connecting to grpc

fkautz:


pkg/tools/tools.go
```diff
@@ -85,3 +87,19 @@ func dial(ctx context.Context, unixSocketPath string) (*grpc.ClientConn, error)
 
 	return c, err
 }
+func WaitForPortAvailable(ctx context.Context, protoType string, registryAddress string) error {
```
edwarnicke: faster pattern:
```go
for ;true;<-time.After(1 * time.Second) {
      select {
           case <-ctx.Done():
               ...
           default:
               ...
      }
} 
```

that way you try before you wait :)

PR #476, opened 2018-11-15
Move constants/validators into package with proto files

edwarnicke:
And refactored (and vastly simplified) dataplane/nsmd as a result.


controlplane/pkg/apis/local/connection/helpers.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/ligato/networkservicemesh/utils/fs"
+)
+
+// IsValid - returns nil if Connection is minimally valid.
+func (c *Connection) IsValid() error {
+	if c == nil {
+		return fmt.Errorf("Connection cannot be nil")
+	}
+	if c.GetNetworkService() == "" {
+		return fmt.Errorf("Connection.NetowrkService cannot be empty: %v", c)
```
fkautz: NetowrkService -> NetworkService

PR #485, opened 2018-11-16
Implement memif connection using vppagent

lobkovilya:


edwarnicke:
@lobkovilya I think if you rebase this on master it may work now.  #486 calls resync to reset the vppagent when the vppagent-dataplane starts.  This causes the vppagent to learn about memif id 0.
This is effectively a workaround for the bug you hit in vppagent.

edwarnicke:
@lobkovilya I look forward to the direct memif memif :)

edwarnicke:
Ah... I see the direct memif memif now :)

PR #500, opened 2018-11-19
Migrate CI to kubeadm

fkautz:


.packet.mk
```diff
@@ -0,0 +1,59 @@
...
+
+.ONESHELL:
+packet-init:
+	@pushd scripts/terraform
+	@terraform init
+	@popd
+
+.ONESHELL:
+.PHONY: packet-start
+packet-start:
```
edwarnicke: Does this check to see if you have a packet cluster running already before starting a new on?

Note: lots and lots of places in the Makefiles just call k8s-start (which would call this) in passing... so it needs to not be creating more and more clusters :)

PR #501, opened 2018-11-19
NSM cleanup

nickolaev:
Introduce resource cleanup to NSE and NSM-K8s.

Not tested in HA/multi-service environment.

### DIFF
--- a/.vagrant.mk
+++ b/.vagrant.mk
@@ -16,9 +16,13 @@
 vagrant-start:
 	@cd scripts/vagrant; vagrant up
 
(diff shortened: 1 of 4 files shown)


.vagrant.mk
```diff
@@ -16,9 +16,13 @@
 vagrant-start:
 	@cd scripts/vagrant; vagrant up
 
+.PHONY: vagrant-destroy
+vagrant-destroy:
```
edwarnicke: Thanks for this... I kept noticing its lack and getting distracted :)

Issue #490, opened 2018-11-18
Expose MonitorCrossConnect service northbound from nsmd


edwarnicke:
Currently, the Dataplane exposes a MonitorCrossConnect service to nsmd to report on CrossConnects.  By exposing MonitorCrossConnect northbound from nsmd, it should be possible for monitoring services like skydive.network to build complete topologies.

This should be a pretty simple task:

1.  In [nsmd](
https://github.com/ligato/networkservicemesh/tree/master/controlplane/pkg/nsmd) instantiate a [MonitorCrossConnectServer](
https://github.com/ligato/networkservicemesh/tree/master/controlplane/pkg/monitor_crossconnect_server) and have it listen to 0.0.0.0:${pick a tcp port number}
2. In nsmd write a simple MonitorCrossConnectClient that talks to the Dataplane’s MonitorCrossConnectServer.
3. Have the MonitorCrossConnectClient Send any CrossConnectEvents its receives to the nsmd’s MonitorCrossConnectServer.

haiodo:
PR: https://github.com/ligato/networkservicemesh/pull/505

edwarnicke:
And merged!

PR #505, opened 2018-11-19
#490

haiodo:
Local tests works, I think about need to add one more test to verify Update/Delete of cross connects and disappear of Dataplane.

### DIFF
--- a/controlplane/pkg/monitor_crossconnect_server/monitor_crossconnect_client.go
+++ b/controlplane/pkg/monitor_crossconnect_server/monitor_crossconnect_client.go
@@ -1,0 +1,133 @@
+// Copyright (c) 2018 Cisco and/or its affiliates.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package monitor_crossconnect_server
+
+import (
+	"context"
+	"github.com/golang/protobuf/ptypes/empty"
+	"github.com/ligato/networkservicemesh/controlplane/pkg/apis/crossconnect"
+	"github.com/ligato/networkservicemesh/controlplane/pkg/apis/registry"
+	"github.com/ligato/networkservicemesh/controlplane/pkg/model"
+	"github.com/sirupsen/logrus"
+	"google.golang.org/grpc"
+	"net"
+	"time"
+)
+
+type nsmMonitorCrossConnectClient struct {
+	monitor    MonitorCrossConnectServer // All connections is here
+	model      model.Model
+	dataplanes map[string]*dataplaneCrossConnectInfo
+}
+
+type dataplaneCrossConnectInfo struct {
+	crossConnects map[string]*crossconnect.CrossConnect
+}
+
+func (client *nsmMonitorCrossConnectClient) EndpointAdded(endpoint *registry.NetworkServiceEndpoint) {
+}
+
+func (client *nsmMonitorCrossConnectClient) EndpointDeleted(endpoint *registry.NetworkServiceEndpoint) {
+}
+
+func (client *nsmMonitorCrossConnectClient) DataplaneAdded(dataplane *model.Dataplane) {
+	info := &dataplaneCrossConnectInfo{
+		crossConnects: make(map[string]*crossconnect.CrossConnect),
+	}
+	client.dataplanes[dataplane.RegisteredName] = info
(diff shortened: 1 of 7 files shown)


controlplane/pkg/monitor_crossconnect_server/monitor_crossconnect_server.go
```diff
@@ -122,3 +130,27 @@ func (m *monitorCrossConnectServer) GetCrossConnect(crossconnectId string) (*cro
 func (m *monitorCrossConnectServer) SendCrossConnectEvent(event *crossconnect.CrossConnectEvent) {
 	m.crossConnectEventCh <- event
 }
+
+func StartNSMCrossConnectServer(model model.Model, address string) (error, *grpc.Server, MonitorCrossConnectServer) {
```
edwarnicke: Let's not promulgate this pattern.  

We are going to need to have multiple APIs listening on whatever the URL we have for the NSM.  This "Start${ServiceName}Server" pattern explicitly gives ownership of a Listener to a *single* service.

controlplane/pkg/monitor_crossconnect_server/monitor_crossconnect_server.go
```diff
@@ -50,13 +53,15 @@ func NewMonitorCrossConnectServer() MonitorCrossConnectServer {
 
 func (m *monitorCrossConnectServer) MonitorCrossConnects(_ *empty.Empty, recipient crossconnect.MonitorCrossConnect_MonitorCrossConnectsServer) error {
 	m.newMonitorRecipientCh <- recipient
-	go func() {
+
+	// We need to wait until it will be done and do not exit
+	for {
```
edwarnicke: Good catch

controlplane/pkg/monitor_crossconnect_server/monitor_crossconnect_client.go
```diff
@@ -0,0 +1,133 @@
+// Copyright (c) 2018 Cisco and/or its affiliates.
```
edwarnicke: Could we put this client code in the nsmd, as it looks like its quite specific to the nsmd.  The MonitorCrossConnectServer in this folder is quite intentionally *generic*, so it can be used by basically anyone who needs to use it :)

controlplane/pkg/monitor_crossconnect_server/monitor_crossconnect_server.go
edwarnicke: Also... this MonitorCrossConnectServer was written to be generic, not nsmd specific, so we wouldn't be passing in nsmd.model.

controlplane/pkg/monitor_crossconnect_server/monitor_crossconnect_client.go
```diff
@@ -0,0 +1,133 @@
...
+		crossConnects: make(map[string]*crossconnect.CrossConnect),
+	}
+	client.dataplanes[dataplane.RegisteredName] = info
+	go client.dataplaneCrossConnectMonitor(dataplane, info)
+}
+
+func (client *nsmMonitorCrossConnectClient) DataplaneDeleted(dataplane *model.Dataplane) {
+}
+
+func StartNSMMonitorCrossConnectClient(model model.Model, monitor MonitorCrossConnectServer) {
```
edwarnicke: Could we simply make this NewMonitorCrossConnectClient(...) rather than Start...

controlplane/pkg/monitor_crossconnect_server/monitor_crossconnect_client.go
```diff
@@ -0,0 +1,133 @@
...
+
+func (client *nsmMonitorCrossConnectClient) EndpointAdded(endpoint *registry.NetworkServiceEndpoint) {
+}
+
+func (client *nsmMonitorCrossConnectClient) EndpointDeleted(endpoint *registry.NetworkServiceEndpoint) {
+}
+
+func (client *nsmMonitorCrossConnectClient) DataplaneAdded(dataplane *model.Dataplane) {
+	info := &dataplaneCrossConnectInfo{
+		crossConnects: make(map[string]*crossconnect.CrossConnect),
```
edwarnicke: The MonitorCrossConnectServer already encapsulates within itself a store for CrossConnects.  I'm curious why we are storing them a second time here :)

controlplane/pkg/monitor_crossconnect_server/monitor_crossconnect_server.go
edwarnicke: I've opened #506 to track this, but am going to go ahead and merge this patch.

controlplane/pkg/monitor_crossconnect_server/monitor_crossconnect_client.go
haiodo: Sure, will do as part of #506

controlplane/pkg/monitor_crossconnect_server/monitor_crossconnect_client.go
haiodo: We need this to be able to remove in case of dataplane will be removed, since in connections itself there is no information about how was added/updated it. Also it is a map of references to same cross connection objects, so should not eat so many memory.

Issue #491, opened 2018-11-19
Write vppagent based icmp-responder-nse


edwarnicke:
We have a [bare icmp-responder](
https://github.com/ligato/networkservicemesh/tree/master/examples/cmd/icmp-responder-nse) written to test kernel interface Mechanisms.

We need something analogous for memif Mechanisms.

This is probably best achieved by following the pattern we utilized for the vppagent-dataplane:
1.  Write vppagent-icmp-responder-nse that exposes the NetworkService API
2.  Have the vppagent-I’m icmp-responder-nse advertise it provides the icmp-responder network service via [RegisterNSE](
https://github.com/ligato/networkservicemesh/blob/master/controlplane/pkg/apis/registry/registry.proto#L28)
3.  Have the vppagent-icmp-responder-nse [complete](
https://github.com/ligato/networkservicemesh/blob/master/examples/cmd/icmp-responder-nse/complete.go#L29) the incoming Connections and ‘convert’ them to DataRequests.
4.  Send the DataRequests to vppagent (presumed to be run in a separate container in the Pod, as is done with the Dataplane).
5.  Change [nsmd](
https://github.com/ligato/networkservicemesh/blob/master/controlplane/pkg/nsmd/network_service_server.go#L106) to send...

NOTE: This isssue is not yet complete and ready to be worked on.

PR #513, opened 2018-11-20
WIP: Issue #491: Create vppagent based icmp-responder-nse 

lobkovilya:


k8s/conf/vppagent-icmp-responder-nse.yaml
```diff
@@ -0,0 +1,39 @@
...
+          securityContext:
+            privileged: true
+          image: networkservicemesh/vppagent-nse:latest
+          imagePullPolicy: IfNotPresent
+          env:
+            - name: INITIAL_LOGLVL
+              value: debug
+          volumeMounts:
+            - name: workspace
+              mountPath: /var/lib/networkservicemesh
```
edwarnicke: NSE's and NSC's can't mount /var/lib/networkservicemesh in their Pod specs.  Their workspaces get mounted there by the DevicePlugin API mechanism.  See https://github.com/ligato/networkservicemesh/blob/master/k8s/conf/icmp-responder-nse.yaml#L17

controlplane/pkg/nsmd/constants.go
```diff
@@ -3,4 +3,5 @@ package nsmd
 const (
 	NsmServerSocketEnv = "NSM_SERVER_SOCKET"
 	NsmClientSocketEnv = "NSM_CLIENT_SOCKET"
+	WorkspaceEnv       = "WORKSPACE"
```
edwarnicke: Have a look at #515.  This doesn't have to be done in this patch, but keep it in mind.

examples/cmd/vppagent-icmp-responder-nse/nse.go
```diff
@@ -0,0 +1,149 @@
...
+	logrus.Infof("nse: listening socket %s", nsmClientSocket)
+	connectionServer, err := net.Listen("unix", nsmClientSocket)
+	if err != nil {
+		logrus.Fatalf("nse: failure to listen on a socket %s with error: %v", nsmClientSocket, err)
+	}
+	grpcServer := grpc.NewServer()
+
+	// Registering NSE API, it will listen for Connection requests from NSM and return information
+	// needed for NSE's dataplane programming.
+	nseConn := New(DefaultVPPAgentEndpoint, path.Join("/var/lib/networkservicemesh", workspace))
```
edwarnicke: I'd suggest here that rather than a join, you simply have the nsmdp pass through a Workspace env variable (which will usually simply point to "/var/lib/networkservicemesh" and just use that here :)

dataplane/vppagent/pkg/converter/memif_interface_converter.go
```diff
@@ -36,23 +37,39 @@ func NewMemifInterfaceConverter(c *connection.Connection, name string) Converter
...
 	}
-	socketPath := path.Join(c.Mechanism.Parameters[connection.Workspace], socketFilename)
+	socketPath := path.Join(memif.BuildMemifDirectory(c.Mechanism), socketFilename)
 	isMaster, err := strconv.ParseBool(c.Mechanism.Parameters[connection.Master])
 	if err != nil {
 		isMaster = false
 	}
+
+	var ipAddresses []string
+	if c.ipAddress != "" {
```
edwarnicke: Just to confirm my understanding, in the vppagent-dataplane case, you simply don't pass in an ipAddress, and thus don't assign one to the vpp side (correct behavior) and in the nse case, you do pass in an ipAddress and thus *do* assign it on the vpp side (also correct behavior)? :)

edwarnicke:
Overall this looks like super good work, we just need to get the confusion around workspaces resolved so it can be merged.

Specifically: NSE's don't mount /var/lib/networkservicemesh in their Pod spec.

Instead, the nsmdp (via the DevicePlugin) causes the workspace specific to this Pod to be mounted into the Pod (usually, though not necessarily at /var/lib/networkservicemesh in the container, though /var/lib/networkservicemesh in the container is typically a mount of /var/lib/networkservicemesh/${workspaceDir} from the host perspective).

Your use of an Env variable for WORKSPACE is good, but that env variable should be passed through nsmdp (thus guaranting it agrees with nsmd's opinion on the matter, since nsmd decides what directory in the container it gets mounted to).  

Once those things are fixed I think we are good to go :)

k8s/conf/vppagent-icmp-responder-nse.yaml
lobkovilya: This mount is created only for vppagent container. Is it ok?

examples/cmd/vppagent-icmp-responder-nse/nse.go
```diff
@@ -0,0 +1,149 @@
...
+
+	nsmServerSocket, _ := os.LookupEnv(nsmd.NsmServerSocketEnv)
+	logrus.Infof("nsmServerSocket: %s", nsmServerSocket)
+	// TODO handle missing env
+
+	nsmClientSocket, _ := os.LookupEnv(nsmd.NsmClientSocketEnv)
+	logrus.Infof("nsmClientSocket: %s", nsmClientSocket)
+	// TODO handle missing env
+
+	workspace, _ := os.LookupEnv(nsmd.WorkspaceEnv)
```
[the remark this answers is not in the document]
lobkovilya: Now this env is set by DevicePluginAPI https://github.com/ligato/networkservicemesh/pull/513/files#diff-63d6957eaddf3c7e04797de53e05457cR66. Or you mean that DevicePluginAPI should return it as a reply?

examples/cmd/vppagent-icmp-responder-nse/nse.go
lobkovilya: Yes, it makes sense :)

examples/cmd/vppagent-icmp-responder-nse/nse.go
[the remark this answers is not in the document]
edwarnicke: I mean it should be returned as an Env variable [in the nsmdp](https://github.com/ligato/networkservicemesh/blob/master/k8s/cmd/nsmdp/nsmdp.go#L63) which will cause it to be present in the Pod.

As to where the value comes from in nsmdp, it should be added to [ClientConnectionReply](https://github.com/ligato/networkservicemesh/blob/master/controlplane/pkg/apis/nsmdapi/nsmd.proto#L13) which is [returned by the nsmd to the nsmdp](https://github.com/ligato/networkservicemesh/blob/master/controlplane/pkg/nsmd/nsmd.go#L75)

This way, rather than cross linking a bunch of constants, that env variable starts in nsmd, gets returned in ClientConnectionReply to the nsmdp, the nsmdp puts it in the AllocateResponse it returns the kubelet, and kubelet will inject the Env variable into the Pod :)

edwarnicke:
CircleCI is having some issues with DockerHub, which is the source of the current build failures.

.circleci/config.yml
```diff
@@ -257,6 +273,9 @@ workflows:
       - build-icmp-responder-nse:
           requires:
             - sanity-check
+      - build-vppagent-icmp-responder-nse:
```
fkautz: We should make the integration tests depend on this task.

examples/cmd/vppagent-icmp-responder-nse/nse.go
```diff
@@ -0,0 +1,148 @@
...
+	logrus.Infof("workspace: %s", workspace)
+	// TODO handle missing env
+
+	// For NSE to program container's dataplane, container's linux namespace must be sent to NSM
+	linuxNS, err := tools.GetCurrentNS()
+	if err != nil {
+		logrus.Fatalf("nse: failed to get a linux namespace with error: %v, exiting...", err)
+		os.Exit(1)
```
fkautz: These exits are unnecessary, fatal kills the app.

examples/cmd/vppagent-icmp-responder-nse/nse.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	grpcServer := grpc.NewServer()
+
+	// Registering NSE API, it will listen for Connection requests from NSM and return information
+	// needed for NSE's dataplane programming.
+	nseConn := New(DefaultVPPAgentEndpoint, workspace)
+
+	networkservice.RegisterNetworkServiceServer(grpcServer, nseConn)
+
+	wg.Add(1)
```
fkautz: This looks like it doesn't have a matching wg.Done

.circleci/config.yml
edwarnicke: Fixed

examples/cmd/vppagent-icmp-responder-nse/nse.go
edwarnicke: Fiex

examples/cmd/vppagent-icmp-responder-nse/nse.go
edwarnicke: Fixed

PR #516, opened 2018-11-20
Enable the VXLAN Mechanism in the vpp-dataplane (#342)

nickolaev:
The remote VXLAN enablement was modelled based on the Kernel interface implementation form the local mechanism.
Tests are still TBD.

### DIFF
--- a/controlplane/pkg/apis/remote/connection/constants.go
+++ b/controlplane/pkg/apis/remote/connection/constants.go
@@ -2,7 +2,7 @@ package connection
 
 // Remote connection constants
 const (
-	VXLANSrcIP = "srcIp"
-	VXLANDstIP = "dstIp"
+	VXLANSrcIP = "src_ip"
+	VXLANDstIP = "dst_ip"
 	VXLANVNI   = "vni"
 )
--- a/controlplane/pkg/apis/remote/connection/constants.go
+++ b/controlplane/pkg/apis/remote/connection/constants.go
@@ -1,0 +1,8 @@
+package connection
+
+// Remote connection constants
+const (
+	VXLANSrcIP = "srcIp"
+	VXLANDstIP = "dstIp"
+	VXLANVNI   = "vni"
+)
--- a/controlplane/pkg/apis/remote/connection/helpers.go
+++ b/controlplane/pkg/apis/remote/connection/helpers.go
@@ -1,11 +1,134 @@
+// Copyright 2018 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+//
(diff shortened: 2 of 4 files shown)


controlplane/pkg/apis/remote/connection/constants.go
```diff
@@ -0,0 +1,8 @@
+package connection
+
+// Remote connection constants
+const (
+	VXLANSrcIP = "srcIp"
```
edwarnicke: At some point we should discuss whether to use camel case or _ for parameters names for Mechanisms.  This is fine for now, but before the arteries harden we should get to consistency :)

controlplane/pkg/apis/remote/connection/helpers.go
```diff
@@ -1,11 +1,134 @@
...
+	if m.GetType() == MechanismType_VXLAN {
+		if _, err := m.SrcIP(); err != nil {
+			return fmt.Errorf("Mechanism.Type %s requires Mechanism.Parameters[%s] for network namespace", m.GetType(), VXLANSrcIP)
+		}
+
+		if _, err := m.DstIP(); err != nil {
+			return fmt.Errorf("Mechanism.Type %s requires Mechanism.Parameters[%s] for network namespace", m.GetType(), VXLANDstIP)
+		}
+
+		if _, err := m.VNI(); err != nil {
```
edwarnicke: VNI's are 24 bit, we should range check the uint32 here to make sure we are actually in the valid 24 bit range :)

controlplane/pkg/apis/remote/connection/constants.go
edwarnicke: Example: https://github.com/ligato/networkservicemesh/blob/master/controlplane/pkg/apis/connectioncontext/constants.go#L4

nickolaev: `src_ip` works for me, thanks for pointing this.

controlplane/pkg/apis/remote/connection/helpers.go
```diff
@@ -1,11 +1,135 @@
...
+	if m.GetParameters() == nil {
+		return 0, fmt.Errorf("Mechanism.Parameters cannot be nil: %v", m)
+	}
+
+	vxlanvni, ok := m.Parameters[VXLANVNI]
+	if !ok {
+		return 0, fmt.Errorf("Mechanism.Type %s requires Mechanism.Parameters[%s]", m.GetType(), VXLANVNI)
+	}
+
+	vni, err := strconv.ParseUint(vxlanvni, 10, 24)
```
nickolaev: VNI 24-bit is handeld here. If the number can not fit, it will generate an error.

controlplane/pkg/apis/remote/connection/helpers.go
nickolaev: m.VNI() will always return a 24-bit number or an error if it does not fit

PR #519, opened 2018-11-20
fake monitor client and server for Skydive development

yf-yang:

### DIFF
--- a/.monitor.mk
+++ b/.monitor.mk
@@ -1,0 +1,10 @@
+.PHONY: mon
+mon: mon-server mon-client
+
+.PHONY: mon-server
+mon-server:
+	CGO_ENABLED=0 GOOS=linux go build -ldflags '-extldflags "-static"' -o $(GOPATH)/bin/mon-server ./controlplane/cmd/monitor-server.go
+
+.PHONY: mon-client
+mon-client:
+	CGO_ENABLED=0 GOOS=linux go build -ldflags '-extldflags "-static"' -o $(GOPATH)/bin/mon-client ./controlplane/cmd/monitor-client.go
--- a/controlplane/cmd/monitor-client.go
+++ b/controlplane/cmd/monitor-client.go
@@ -1,0 +1,58 @@
+package main
+
+import (
+	"context"
+	"net"
+	"os"
+	"os/signal"
+	"syscall"
+	"time"
+
+	"github.com/golang/protobuf/ptypes/empty"
+	"github.com/ligato/networkservicemesh/controlplane/pkg/apis/crossconnect"
+	"github.com/sirupsen/logrus"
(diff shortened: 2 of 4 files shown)


controlplane/cmd/monitor-client.go
```diff
@@ -0,0 +1,58 @@
+package main
```
edwarnicke: Could we put these in examples/skydive/ rather than in controlplane?

yf-yang: Yep, no problem

yf-yang: @matrohon Just represented you to reply to Ed (I think you won't mind that haha)

controlplane/cmd/monitor-client.go
```diff
@@ -0,0 +1,58 @@
...
+func main() {
+	// Capture signals to cleanup before exiting
+	c := make(chan os.Signal, 1)
+	signal.Notify(c,
+		syscall.SIGHUP,
+		syscall.SIGINT,
+		syscall.SIGTERM,
+		syscall.SIGQUIT)
+	var err error
+	conn, err := dial(context.Background(), "tcp", "127.0.0.1:5007")
```
fkautz: Where do you intend to run this? Running this anywhere but within the pod exposing 5007 will not work.

controlplane/cmd/monitor-server.go
```diff
@@ -0,0 +1,36 @@
...
+	// Capture signals to cleanup before exiting
+	c := make(chan os.Signal, 1)
+	signal.Notify(c,
+		syscall.SIGHUP,
+		syscall.SIGINT,
+		syscall.SIGTERM,
+		syscall.SIGQUIT)
+
+	myModel := model.NewModel("127.0.0.1:5000")
+	crossConnectAddress := "127.0.0.1:5007"
```
fkautz: Same question as above, where is this going to run?

.monitor.mk
```diff
@@ -0,0 +1,10 @@
+.PHONY: mon
+mon: mon-server mon-client
+
+.PHONY: mon-server
+mon-server:
```
fkautz: Adding it here does not automatically add it to circleci. Please add these binaries to the circleci config.

controlplane/cmd/monitor-client.go
yf-yang: For the time being we are not going to run it inside the k8s cluster. Since we just take over these code, we intend to run it on the host machine. The goal is simple: a simple model to simulate events sending by MonitorServer, and extract information from the client.  
Maybe we should develop this functionality in another repository?  What's your advice?  
[#520](https://github.com/ligato/networkservicemesh/pull/520) is the same PR, I just move those file to another folder.

PR #522, opened 2018-11-21
NetworkServiceServer/NetworkService client.

haiodo:
1. usage of NetworkServiceDiscovery
2. services registry to do connections to internal services
3. services apiRegistry for starting up network apis.
4. partial test of NSC -> NSMD/Registry

This is work in progress, no merge are required yet.

@edwarnicke Please take a look on serviceregistry.go, it is initial step to have one place for all constants and connectivity implementations.


edwarnicke:
Direction looks good so far 👍

edwarnicke:
Also... don’t forget to rebase on master, found a bug in getting the NetworkServiceManager from the nsmd-k8s that I fixed while you were sleeping :)

haiodo:
Yes, will rebase, Thanks.

edwarnicke:
It looks like something is breaking the nsmd returning a workspace to nsmdp:
[Take a look at the Dump nsmd logs](https://circleci.com/gh/ligato/networkservicemesh/6043)

edwarnicke:
I rebased this onto master to work with the code that makes memif work.

I also added vagrant machinery for a two node vagrant cluster.

Additionally, we are now deploying four NSCs (all on the same node).  They attempt to connect to the icmp-responder Network Service.  We have two each of icmp-responder-nse (which uses kernel interfaces) and vppagent-icmp-responder-nse (which uses memif interfaces).  They are configured such that one each of them runs on each node.

nsmd currently is round robining among endpoints for the same service, so each nsc attempts to connect to one of the icmp-responders.  Because all nscs are on one node, and each node has one each of icmp-responder-nse and icmp-responder-nse, two of the nsc's will get connected locally (same) node, and two remotely (the other node).

If you run:

make k8s-check

it will attempt a ping check across all four nscs.  Right now two succeed (because local connections are working) and two fail (because remote connections are not).

Also, commands like:
make k8s-nsc-logs
make k8s-vppagent-icmp-responder-nse-logs

etc will conveniently dump logs in a single command.

edwarnicke:
Some hopefully helpful comments:

Have a look at #516 for a deep dive on VXLAN.  VXLAN requires [three parameters to be complete](https://github.com/ligato/networkservicemesh/blob/master/controlplane/pkg/apis/remote/connection/constants.go): src_ip, dst_ip, and vni.

The originator of the remote.Connection using Mechanism VXLAN has to provide src_ip... the destination for remote.NetworkService.Request(remote.Connection) for Connection with Mechanism VXLAN has to fill in dst_p and vni.

This should leave you with the question:

How do I get these parameters.

From the Dataplane.MonitorMechanisms() call.  vppagent-dataplane needs to start advertising a Mechanism for VXLAN, and in that advertisement it needs to have src_ip.

I've just pushed my commit here to cause vppagent-dataplane to correctly advertise a usable srcIP in the response to MonitorMechanisms.  Hopefully that helps.

edwarnicke:
Quick question... am I correct that you have not yet implemented remote.NetworkServiceServer ?

edwarnicke:
One other note: the integration tests are failing currently for this PR because we need a second node for packet (@fkautz is working on it).  If that passes, currently they will fail because we don't actually succeed in the remote pinging.

edwarnicke:
A few more comments on design.

Effectively, an nsmd has a bunch of Peers.

Think of a Peer as an entity against which you can call remote or local NetworkService and MonitorConnections.

Workspace is an example of such a Peer, a LocalPeer, LocalPeers speak local.NetworkService and local.MonitorConnection

nsmd also has RemotePeers , which are other nsmds. and which speak remote.NetworkService and remote.MonitorConnection

If we define a common interface:
```go
type interface Peer {
    CompleteCrossConnect(CrossConnect) error
}
```
Then we can have:

```model.LookupPeerByEndpoint(NetworkServiceEndPoint) Peer```

and have it return either a LocalPeer, or a RemotePeer, but both providing the Peer Interface.

We would have one LocalPeer for each Workspace (you might even simply give Workspace the Peer API), and one RemotePeer per nsmd with which we have connections.

if model.LookupPeerByEndpoint(NetworkServiceEndpoint) is called for a NetworkServiceEndpoint which is not served by a LocalPeer (aka Workspace) for which we don't have a RemotePeer already, it can create it, note it (so we only have one per Remote nsmd) and return it.

Each Peer should keep its own MonitorConnectionsServer (with its own data about what Connections it has with its peer).

Because each LocalPeer has a specific unix file socket pair in its Workspace, this is simple for LocalPeers.

For RemotePeers, we have a single [remote.MonitorConnections(MonitorScopeSelector)](https://github.com/ligato/networkservicemesh/blob/master/controlplane/pkg/apis/remote/connection/connection.proto#L65) (where MonitorScopeSelector is basically an nsm name), we would want a Delegate for it listening for incoming tcp connections.  Basically, a MonitorMechanismDelegateServer that simply calls:

model.LookupPeerByNsmName(nsmName string) Peer
and passes the MonitorMechanism call off to that RemotePeer

This way the global MonitorMechanismDelegateServer is stateless, essentially only calling the servers within each Peer.  The Peer continues to manage its own Connection state.

This will become important in the future, as each Peer will be able to track state differences between what it knows about Connections and what it can learn from its peer... but that is for the future.

Each RemotePeer can implement (or have a member that implements) remote.NetworkService, (as is true for LocalPeers with Workspaces.  We can have a:

RemoteNetworkServiceDelegateServer that implements the remote.NetworkService that listens for incoming TCP connections, looks in them for their [source_network_service_name ](https://github.com/ligato/networkservicemesh/blob/master/controlplane/pkg/apis/remote/connection/connection.proto#L46) and uses that to lookup the proper peer to delegate to.

**Comment by edwarnicke** [comment]
In this way, a Peer acts as both a NetworkServiceServer and MonitorConnectionServer (but only for its peer) and also as a NetworkServiceClient (but only to its peer).

For LocalPeers, Workspaces and their unix sockets handle the delegation to the correct Peer for incoming calls.

For RemotePeers, we have the NetworkServiceDelegate and MonitorConnectionsDelegate listening for tcp connections, and delegating to the RemotePeer.

When a RemotePeer receives a MonitorConnections call, it responds from its own MonitorConnections information.

When a RemotePeer receives a NetworkService.Request or NetworkService.Close() from a remote nsmd, it uses model.LookupPeerByEndpoint() and calls CompleteCrossConnect.

**Comment by edwarnicke** [comment]
Oh... and please note, when I say model above, if serviceregistry is a better fit, feel free to use something like that.

This is meant to be brainstorming, not necessarily point by point dictation :)

**Comment by haiodo** [comment]
> Quick question... am I correct that you have not yet implemented remote.NetworkServiceServer ?
It is implemented.

It is here: /controlplane/pkg/remote/network_service_server/remote_network_service_server.go

**Comment by edwarnicke** [comment]
> > Quick question... am I correct that you have not yet implemented remote.NetworkServiceServer ?
> > It is implemented.
> 
> It is here: /controlplane/pkg/remote/network_service_server/remote_network_service_server.go

Ah... I see it now... that's a good place for it, but not where I had been looking :)

As to the design comments... for the moment, lets take the fastest path to getting this all working.  But as we go forward we will need to deal with the restart issue, and that may be a good time to rethink things.

Which is to say: if you are close to having things working, don't refactor the world just now to follow the design suggestions.  I leave it to your judgement how much to adopt at this point.

File: .vagrant.mk [hunk 1]
```diff
@@ -34,7 +34,11 @@ vagrant-resume:
 
 .PHONY: vagrant-ssh
 vagrant-ssh:
-	@cd scripts/vagrant; vagrant ssh
+	@cd scripts/vagrant; vagrant ssh master
```
edwarnicke: Good catch :)

controlplane/pkg/nsmd/network_service_server.go
```diff
@@ -102,7 +102,7 @@ func (srv *networkServiceServer) Request(ctx context.Context, request *networkse
 		defer dataplaneConn.Close()
 	}
 
-	endpoint, err := srv.model.SelectEndpoint(request.GetConnection().GetNetworkService())
+	endpoint, err := srv.getEndpoint(request)
```
edwarnicke: Could we preserve the round robin behavior of the SelectEndpoint?  Its important for the integration testing.

edwarnicke: Ah, I see, you do unless there's an error... got it :)

controlplane/pkg/nsmd/network_service_server.go
```diff
@@ -158,19 +158,20 @@ func (srv *networkServiceServer) createLocalNSERequest(endpoint *registry.NSEReg
 	return message
 }
 
-func getEndpoint(srv *networkServiceServer, request *networkservice.NetworkServiceRequest) (*registry.NSERegistration, error) {
-	endpoints := srv.model.GetNetworkServiceEndpoints(request.GetConnection().GetNetworkService())
+func (srv *networkServiceServer) getEndpoint(request *networkservice.NetworkServiceRequest) (*registry.NSERegistration, error) {
+	endpoint, err := srv.model.SelectEndpoint(request.GetConnection().GetNetworkService())
```
edwarnicke: Could we incorporate remote endpoints into SelectEndpoints... we really do want to do *one* selection, whether remote or local.  For right now, that's going to be round robin, but @nickolaev is working on more sophisticated selection.  So in that case, SelectEndpoint would need to make the call the the registry to get its list of candidates, and round robin among them.

controlplane/pkg/nsmd/network_service_server.go
haiodo: Yes, it ask for local ones and if there is none go for nsmd-k8s registry.

controlplane/pkg/nsmd/network_service_server.go
haiodo: Sure, this make seanse

haiodo: On the other side, Model should not do any remote calls.

haiodo:
@edwarnicke NSC -> NSMD -> NSMD -> NSE flow is functional, right now I've have issues with passing CrossConnection to remote Dataplan. Also I've choose "VNI" as 1 for now, not sure how to generate them uniq way.

Error I've receiving right now:

`Error Converting CrossConnect id:"1" payload:"IP" remote_source:<id:"1" network_service:"icmp-responder" mechanism:<type:VXLAN parameters:<key:"dst_ip" value:"10.32.0.4" > parameters:<key:"src_ip" value:"10.44.0.1" > parameters:<key:"vni" value:"1" > > context:<key:"dst_ip" value:"10.20.1.2/30" > context:<key:"src_ip" value:"10.20.1.1/30" > > local_destination:<id:"2" network_service:"icmp-responder" mechanism:<type:KERNEL_INTERFACE parameters:<key:"name" value:"nsm2" > parameters:<key:"netnsInode" value:"4026532618" > > context:<key:"dst_ip" value:"10.20.1.2/30" > context:<key:"src_ip" value:"10.20.1.1/30" > > : No file found in /proc/*/ns/net with inode 4026532618: error stat file: stat /proc/22502/ns/net: no such file or directory `

controlplane/pkg/nsmd/network_service_server.go
edwarnicke: I agree with you, but it might be even better to say that model doesn't make any calls that leave the process (whether to another nsmd or to an NSE).

edwarnicke:
Is this the error you are getting running in K8s?

edwarnicke:
OK... so this now *works*.

make k8s-save k8s-deploy
# You may want to pause a beat before the make k8s-check
make k8s-check

All the pings go through for the testing (at least locally).  CI is failing for some reason that will need to be tracked down, but log inspection looks fine.

I would recommend:
1.  Confirm this works for you.
2. Have a look at what I did.
3. Some of the things I did were ugly, and we should clean them up.
4. Dig into why the CI tests may be failing.

haiodo:
1. Works locally, with one change, I've forced all nsc to be on master node. On my case they all was on worker node.
2. Looks good, except we loose functionality of just local NSM without nsmd-k8 and because of this local model endpoints are not really used. I suppose we need to choose from local endpoints and if there is none choose from remote ones. So it would be not a fair round-robin.
3. Does "nvi" should be same for local/remote dataplan configuration? right now they could be different ones?
4. Looking to CI, not sure for reason yet.
5. @edwarnicke could you please do not "force" push changes into merge requests, this is not required until "signing is missed", it is not very user friendly to update from force push :-) Since github create a separate branch for pull requests we could push here without issues.

edwarnicke:
> * Works locally, with one change, I've forced all nsc to be on master node. On my case they all was on worker node.

Interesting.  What problems were you seeing?  The nscs should all land on the same node (master or worker), and I’m curious why landing on a particular node was making a difference.

> * Looks good, except we loose functionality of just local NSM without nsmd-k8 and because of this local model endpoints are not really used. I suppose we need to choose from local endpoints and if there is none choose from remote ones.

Yeah, round robin ( and eventually #531 ) are the correct selection process.

> * Does "nvi" should be same for local/remote dataplan configuration? right now they could be different ones?

“nvi”?

> * Looking to CI, not sure for reason yet.
> * @edwarnicke could you please do not "force" push changes into merge requests, this is not required until "signing is missed", it is not very user friendly to update from force push :-) Since github create a separate branch for pull requests we could push here without issues.

I understand.  I only force pushed after rebasing on master.  Master had moved a fair bit, and there were conflicts that had to be resolved.  I was hopeful CI would work.  Do you have a suggestion for a better way to handle rebasing on master?

haiodo:
Vagrant file also statically configure this ones, problem is with my env, eth1 could switch with eth2 sometimes, not sure why :-(

edwarnicke:
@haiodo Apology for the force push, but your merge push lacked a signoff, and so DCO didn't pass.

Issue #493, opened 2018-11-19
Eliminate race condition between nsmdp, nsmd, and Dataplane


edwarnicke:
Currently, there is a race condition between nsmdp, nsmd, and Dataplane.

nsmd will accept [NSMD.RequestClientConnection(..)](https://github.com/ligato/networkservicemesh/blob/master/controlplane/pkg/apis/nsmdapi/nsmd.proto#L31) calls before it has a Dataplane with which to service them.  This can lead to a race condition where an nsc can come up and seek to request a NetworkService before the nsmd has a Dataplane to service them, and thus fail.

As with all race conditions, this is a failure to enforce ordering of events.   The proper order of events is:
1.  Dataplane registers with nsmd
2. nsmd begins listening for incoming [NSMD.RequestClientConnection(..)](https://github.com/ligato/networkservicemesh/blob/master/controlplane/pkg/apis/nsmdapi/nsmd.proto#L31) calls

This can be easily rectified in [StartNSMServer](https://github.com/ligato/networkservicemesh/blob/master/controlplane/pkg/nsmd/nsmd.go#L98) by looping over a call to model.SelectDataplane() until it returns a non-nil value.

This solves the race condition between nsmd and the Dataplane.

But it reveals a new race condition between the nsmdp and the nsmd.

If nsmdp receives an [Allocate call](https://github.com/ligato/networkservicemesh/blob/master/k8s/cmd/nsmdp/nsmdp.go#L46) before the nsmd is listening for [NSMD.RequestClientConnection(..)](https://github.com/ligato/networkservicemesh/blob/master/controlplane/pkg/apis/nsmdapi/nsmd.proto#L31) calls, it will reply with an error to kubelet.

Again, this is resolved by enforcing proper ordering:
1.  [NSMD.RequestClientConnection(..)](https://github.com/ligato/networkservicemesh/blob/master/controlplane/pkg/apis/nsmdapi/nsmd.proto#L31) calls become serviceable.
2. nsmdp [Registers](https://github.com/ligato/networkservicemesh/blob/master/k8s/cmd/nsmdp/nsmdp.go#L75) with kubelet.

[...]

PR #528, opened 2018-11-22
nsmd begins listening for incoming calls after data plane registers

platoff:
This change enforce a proper order of events as described in #493

The proper order of events is:
1. Dataplane registers with nsmd
1. nsmd begins listening for incoming NSMD.RequestClientConnection(..) calls

This can be easily rectified in StartNSMServer by looping over a call to model.SelectDataplane() until it returns a non-nil value.

### DIFF
--- a/controlplane/pkg/nsmd/nsmd.go
+++ b/controlplane/pkg/nsmd/nsmd.go
@@ -103,18 +103,11 @@ func (nsm *nsmServer) DeleteClientConnection(context context.Context, request *n
 
 func waitForDataplaneAvailable(model model.Model) {
 	logrus.Info("Waiting for dataplane available...")
-	ticker := time.NewTicker(100 * time.Millisecond)
-	available := make(chan bool)
-	go func() {
-		for _ = range ticker.C {
-			dp, _ := model.SelectDataplane()
-			if dp != nil {
-				available <- true
-				ticker.Stop()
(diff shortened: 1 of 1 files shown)


controlplane/pkg/nsmd/nsmd.go
```diff
@@ -100,10 +101,27 @@ func (nsm *nsmServer) DeleteClientConnection(context context.Context, request *n
 	return &nsmdapi.DeleteConnectionReply{}, nil
 }
 
+func waitForDataplaneAvailable(model model.Model) {
+	logrus.Info("Waiting for dataplane available...")
+	ticker := time.NewTicker(100 * time.Millisecond)
+	available := make(chan bool)
+	go func() {
+		for _ = range ticker.C {
```
edwarnicke: Use construct:
```go
for ;true; <-time.After(time.Duration) {
       ...
       if condition {
           break
       }
}
```
ticker.C will wait for a full tick before trying to do anything (not a huge issue with 100*.time.Millisecond, but if things get retuned to a less aggressive interval, it can become important.

Spawning a go routing and waiting on a channel from that go routine is more complex than the simple for loop above with a break :)

platoff: Thank you for the suggestion!

PR #529, opened 2018-11-22
DO NOT MERGE: VXLAN Dataplane API testing

nickolaev:
A simple tool to check the operability of the VXLAN enablement code.

In a vagrant deployment:
```
    make k8s-save-deploy

    kubectl exec nsm-vppagent-dataplane-<ID> vppctl show vxlan tunnel
```

### DIFF
--- a/.k8s.mk
+++ b/.k8s.mk
@@ -15,7 +15,7 @@
 K8S_CONF_DIR = k8s/conf/
 
 # Need nsmdp and icmp-responder-nse here as well, but missing yaml files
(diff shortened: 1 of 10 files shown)


nickolaev:
@edwarnicke I was thinking that maybe this binary can be made part of NSMD container and can be invoked through `kubectl exec`. Of course it will need some refactoring. But it can become a generic datplane query tool. Does it make sense?

edwarnicke:
We have integration testing now :)

Issue #496, opened 2018-11-19
Dynamically expand nsmdp pool of DeviceIds


edwarnicke:
nsmdp currently has a fixed n (currently n=10) DeviceIds it advertises via ListAndWatch.  nsmdp needs to be augmented so that it always has n (currently n=10) free DeviceIds, so that as new requests come in from kubelet, we always have a pool of at least n available.

It is likely better to think of this in concert with #495

Issue #495, opened 2018-11-19
Refactor nsmdp to have DeviceId match nsmd's id for ClientConnections


edwarnicke:
Currently, nsmdp dynamically calls RequestClientConnection in response to an Allocate via the DevicePlugin API.  This late binding means we have two ids in play:

1.  DeviceId that nsmdp has advertised in ListAndWatch
2. nsmd's internal id for ClientConnection

nsmdp needs to be refactored to acquire n (currently n=10) ClientConnections at startup, and use the id for the ClientConnection as the DeviceId when responding to ListAndWatch.  This then allows nsmdp in the Allocate call to make a call (see #494 ) to Reset that ClientConnection, rather than ask it to be Allocated.

Issue #494, opened 2018-11-19
Refactor nsmdapi to better match device plugin api


edwarnicke:
Currently [nsmdapi](https://github.com/ligato/networkservicemesh/tree/master/controlplane/pkg/apis/nsmdapi) allows the nsmdp to request clientconnections from nsmd.

The semantics of DevicePlugin API involves streaming (via ListAndWatch) a list of DeviceIds (and health status) to kubelet.   When kubelet has a Pod that requires one or more instances of a resource, it asks via Allocate (by name) for the DeviceIds it has selected.

It should be noted that Kubelet *recycles*, meaning that while it won't tell nsmdp that its finished with a DeviceId, it will ask it eventually to reallocate it to another purpose.

Currently [nsmdapi](https://github.com/ligato/networkservicemesh/tree/master/controlplane/pkg/apis/nsmdapi) allows the nsmdp to request clientconnections from nsmd allows:

```proto
service NSMD {
    rpc RequestClientConnection (ClientConnectionRequest) returns (ClientConnectionReply);
    rpc DeleteClientConnection (DeleteConnectionRequest) returns (DeleteConnectionReply);
}
```

It needs to be refactored in a manner that:

Allows the 'recycling' of a Client Connection, letting nsmd know to reset whatever internal state it has for that Client Connection so it can be used by others

PR #536, opened 2018-11-25
nsdmp refactorings

platoff:
Dynamically expand nsmdp pool of DeviceIds #496, Refactor nsmdp to have DeviceId match nsmd's id for ClientConnections #495, Refactor nsmdapi to better match device plugin api #494

### DIFF
--- a/k8s/cmd/nsmdp/nsmdp.go
+++ b/k8s/cmd/nsmdp/nsmdp.go
@@ -41,12 +41,15 @@ const (
 )
 
 type nsmClientEndpoints struct {
+	allocatedDevices map[string]bool
 }
 
 func (n *nsmClientEndpoints) Allocate(ctx context.Context, reqs *pluginapi.AllocateRequest) (*pluginapi.AllocateResponse, error) {
 	logrus.Info("Client request for nsmdp resource...")
 	responses := &pluginapi.AllocateResponse{}
-	for range reqs.ContainerRequests {
+	for _, req := range reqs.ContainerRequests {
+		id := req.DevicesIDs[0]
+		n.allocatedDevices[id] = true
 		workspace, err := nsmd.RequestWorkspace()
 		logrus.Infof("Received Workspace %v", workspace)
 		if err != nil {
@@ -109,7 +112,7 @@ func (n *nsmClientEndpoints) ListAndWatch(e *pluginapi.Empty, s pluginapi.Device
 	logrus.Infof("ListAndWatch was called with s: %+v", s)
 	for {
 		resp := new(pluginapi.ListAndWatchResponse)
-		for dev := 1; dev < 10; dev++ {
+		for dev := 1; dev < 10+len(n.allocatedDevices); dev++ {
 			resp.Devices = append(resp.Devices, &pluginapi.Device{
 				ID:     fmt.Sprintf("%d", dev),
 				Health: pluginapi.Healthy,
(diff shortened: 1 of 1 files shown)


edwarnicke:
This is good as far as it goes.

We should think about what happens on restart of the nsmd and nsmdp.

On restart the nsmd can *sort* of figure out what workspaces it has by looking at the workspace directories it created in the past.

The nsmdp does not have that luxury (at least without teaching it about those nsmdp directories which  overcomplicates it).

For the moment, if nsmdp could "MonitorClients" to figure out what Clients the nsmd has, it could then

1.  Request new clients if the pool is to small
2. Advertise all know clients in ListAndWatch

Lets talk through an example:

1. nsmd and nsmdp run. 
2. nsmdp calls MonitorClient on nsmd, and discovers nsmd has no clients.
3. nsmdp requests up to the pool size (10 currently) of clients from nsmd
4. nsmdp Registers with kubelet, and responds to ListAndWatch with those 10 clients.
5. Time passes, things happen, and the total number of clients grows to 15.
6. nsmd and nsmdp restart
7.  nsmd sees that it has 15 clients
8. nsmdp calls MonitorClients against nsmd, and gets told about the 15 clients
9. nsmdp evaluates whether it needs to get more clients to meet the pool size, and does so
10. nsmdp registers with kubelet and responds to ListAndWatch with all the clients

platoff:
@edwarnicke Let's look at this from another angle. We're using device plugin API to "provision" workspace. The reason we're using device plugin API not because we provide devices, but because this is the only API available to fulfil our provisioning needs. 

So we need to think in "devices" when using device plugin API, and I propose to look at our "device" as a "factory", or workspace provisioner. Looking from this angle means: a device is stateless, so any device can be used anytime, the only thing we have to care is to report enough devices to the system, so a provisioner can be allocated anytime. 

That's how things work currently, did I miss anything?

edwarnicke:
The question becomes: how do we determine ‘enough’ between nsmd and nsmdp restarts?

edwarnicke:
When nsmdp restarts, we have no way of knowing how many devices kubelet thinks are already allocated.

platoff:
Do you have an idea how this works with real devices? If we have a GPU device plugin and device plugin restarts?

edwarnicke:
Yes.  

When a device plugin is registered, it indicates what resource type it is providing.  Then kubelet calls ListenAndWatch against it to get the list of DeviceIds.  Since for a physical device, the number of things you have is fixed, it simply sends all of them each time.

Kubelet tracks whcih DeviceIds it has allocated.  The only indication the DeviceId gets that a DeviceId is no longer in use is that Kubelet calls the allocate call on the DevicePlugin with that DeviceId again.

Example:
I have 4 foo.gpu resources: Foo-1, Foo-2, Foo-3, Foo-4.

The Device Plugin registers for resource type foo.gpu.  When kubelet calls ListenAndWatch on the DevicePlugin it returns the full list {Foo-1,Foo-2, Foo-3, Foo-4}.  Kubelet then adds the count of Devices of type foo.gpu to the Node Resource with the API server, so that the scheduler knows how many available foo.gpu there are on the Node.

Say Kubelet Allocates Foo-3.  Kublet reduces by 1 the number of foo.gpu it has in the Node Resource.

Now say the DevicePlugin restarts.

The DevicePlugin simply readvertises Foo-[1-4] in ListAndWatch.  Kubelet is happy, because it knows the pool of available DeviceIds is still there.  The DevicePlugin is happy, because it doesn't do anything to reset or initialize those Devices until it gets an Allocate call.  Its still got four devices in the pool, three unallocated.

Issue #531, opened 2018-11-22
Extending NetworkService API to allow for Chaining


edwarnicke:
This takes inspiration for Istio VirtualService.

Currently, NetworkService CRs look roughly like:

```yaml
kind: NetworkService
metadata:
  name: secure-intranet-connectivity
spec:
  payload: IP
```

Basically, each NetworkService has a name and a payload type.

If more than one NetworkServiceEndpoint provides a NetworkService, in response to a request for a Connection to a Network Service from a NetworkServiceClient, the nsmd round robins among them.

There are lots of cases however where it is helpful to have more control, to be able to Chain together NetworkServiceEndpoints (NSEs) to provide a single Network Service:

![network service extensions for chaining](https://user-images.githubusercontent.com/1455208/48921483-b7597180-ee65-11e8-978c-f54bc01dac69.png)

This issue is about putting together a basic extension to NetworkService to allow us to build such chains.

Each Network Service Endpoint when it [advertise itself](https://github.com/ligato/networkservicemesh/blob/master/controlplane/pkg/apis/registry/registry.proto#L50) includes a [set of labels](https://github.com/ligato/networkservicemesh/blob/master/controlplane/pkg/apis/registry/registry.proto#L13).  For the purpose of this discussion, lets call those ```destinationLabels```.

Each Network Service Client with it [requests a Connection to a NetworkService](https://github.com/ligato/networkservicemesh/blob/master/controlplane/pkg/apis/local/networkservice/networkservice.proto#L34) includes a [set of labels](https://github.com/ligato/networkservicemesh/blob/master/controlplane/pkg/apis/local/connection/connection.proto#L44).  For the purposes of this discussion, lets call those ```sourceLabels```.

We can extend the Network Service to be more like:

```yaml
kind: NetworkService
metadata:
  name: secure-intranet-connectivity
spec:
  payload: IP
  matches:
  - match:
    sourceSelector:
        app: firewall
    route:
      - destination:
          destinationSelector:
            app: vpn-gateway
  - match:
    route:
      - destination:
          destinationSelector:
            app:firewall
```

Now when selecting the Network Service Endpoint to connect an NSC to, the nsmd goes through the 'matches' until one matches.

So when a Client Pod, which has no sourceLabel "app=firewall" requests a connection, it will fail the first match.  The second match, which has no sourceSelector will match *any* connection, so it matches, and the Client Pod is connected to a Network Service Endpoint that has a destinationLabel "app=firewall".

[...]

PR #537, opened 2018-11-26
Extending NetworkService API #531

nickolaev:


controlplane/pkg/apis/registry/gen.go
```diff
@@ -1,3 +1,3 @@
 package registry
 
-//go:generate protoc -I . registry.proto --go_out=plugins=grpc:. --proto_path=$GOPATH/src
+//go:generate protoc -I ../../../../vendor -I . registry.proto --go_out=plugins=grpc:. --proto_path=$GOPATH/src
```
nickolaev: I don't like this approach but that's something working. Could there be a better way to point protocol to the `vendor` folder?

edwarnicke: So interestingly, I don't have to do this when I use vendored things:
https://github.com/ligato/networkservicemesh/blob/master/controlplane/pkg/apis/local/connection/connection.proto#L23

https://github.com/ligato/networkservicemesh/blob/master/controlplane/pkg/apis/local/connection/gen.go#L3

What problem did you hit that led to this?

controlplane/pkg/apis/registry/registry.proto
```diff
@@ -17,6 +18,15 @@ message NetworkServiceEndpoint {
 message NetworkService {
     string name = 1;
     string payload = 2;
+    message Match {
+        k8s.io.apimachinery.pkg.apis.meta.v1.LabelSelector source_selector = 1;
+        message Destination{
+            k8s.io.apimachinery.pkg.apis.meta.v1.LabelSelector destination_selector = 1;
+            uint32 weight = 2;
```
edwarnicke: What did you have in mind for weight?

controlplane/pkg/apis/registry/registry.proto
```diff
@@ -17,6 +18,15 @@ message NetworkServiceEndpoint {
 message NetworkService {
     string name = 1;
     string payload = 2;
+    message Match {
+        k8s.io.apimachinery.pkg.apis.meta.v1.LabelSelector source_selector = 1;
+        message Destination{
```
edwarnicke: Could we define this not-inline in another message?  It turns out to make the generated Go super messy.

controlplane/pkg/apis/registry/registry.proto
edwarnicke: Oh wait... of course, 'weight'...

controlplane/pkg/apis/registry/registry.proto
```diff
@@ -17,6 +18,15 @@ message NetworkServiceEndpoint {
...
+            k8s.io.apimachinery.pkg.apis.meta.v1.LabelSelector destination_selector = 1;
+            uint32 weight = 2;
+        }
+        Destination route = 2;
```
edwarnicke: Should this be

repeated Destination

?

k8s/pkg/apis/networkservice/v1/types.go
```diff
@@ -25,7 +25,18 @@ type NetworkService struct {
...
 
 type NetworkServiceSpec struct {
-	Payload string `json:"payload"`
+	Payload string  `json:"payload"`
+	Matches []Match `json:"matches"`
+}
+
+type Match struct {
+	SourceSelector *metaV1.LabelSelector `json:"sourceSelector,omitempty"`
+	Route          Destination           `json:"route"`
```
edwarnicke: Should this be a []*Destination ?

controlplane/pkg/apis/registry/registry.proto
nickolaev: I was not sure about this either.

PR #539, opened 2018-11-26
controlplane: implement a simple Ednpoint selection. Implements 534

nickolaev:
This implements an exact match when requesting a NSE selection.

Based on PR https://github.com/ligato/networkservicemesh/pull/537

### DIFF
--- a/controlplane/pkg/model/model.go
+++ b/controlplane/pkg/model/model.go
@@ -203,7 +203,7 @@ func NewModel() Model {
 		networkServices: make(map[string][]*registry.NSERegistration),
 		endpoints:       make(map[string]*registry.NSERegistration),
 		listeners:       []ModelListener{},
-		selector:        selector.NewRoundRobinSelector(),
+		selector:        selector.NewMatchSelector(),
 		lastVNI:         1,
 	}
 }
--- a/controlplane/pkg/nsmd/network_service_server.go
+++ b/controlplane/pkg/nsmd/network_service_server.go
@@ -41,7 +41,7 @@ func NewNetworkServiceServer(model model.Model, workspace *Workspace, serviceReg
 	}
 }
 
-func (srv *networkServiceServer) getEndpointFromRegistry(ctx context.Context, networkServiceName string, networkServiceLabels map[string]string) *registry.NSERegistration {
+func (srv *networkServiceServer) getEndpointFromRegistry(ctx context.Context, requestConnection *connection.Connection) *registry.NSERegistration {
 	// Get endpoints
 	discoveryClient, err := srv.serviceRegistry.NetworkServiceDiscovery()
 	if err != nil {
@@ -49,14 +49,14 @@ func (srv *networkServiceServer) getEndpointFromRegistry(ctx context.Context, ne
 		return nil
 	}
 	nseRequest := &registry.FindNetworkServiceRequest{
-		NetworkServiceName: networkServiceName,
+		NetworkServiceName: requestConnection.GetNetworkService(),
 	}
 	endpointResponse, err := discoveryClient.FindNetworkService(ctx, nseRequest)
 	if err != nil {
 		logrus.Error(err)
 		return nil
 	}
-	endpoint := srv.model.GetSelector().SelectEndpoint(endpointResponse.GetNetworkService(), networkServiceLabels, endpointResponse.GetNetworkServiceEndpoints())
+	endpoint := srv.model.GetSelector().SelectEndpoint(requestConnection, endpointResponse.GetNetworkService(), endpointResponse.GetNetworkServiceEndpoints())
 	if endpoint == nil {
 		return nil
 	}
@@ -97,7 +97,7 @@ func (srv *networkServiceServer) Request(ctx context.Context, request *networkse
 		defer dataplaneConn.Close()
 	}
 
-	endpoint := srv.getEndpointFromRegistry(ctx, request.GetConnection().GetNetworkService(), request.GetConnection().GetLabels())
+	endpoint := srv.getEndpointFromRegistry(ctx, request.GetConnection())
 	if err != nil {
 		return nil, err
 	}
--- a/controlplane/pkg/nsmd/network_service_server.go
+++ b/controlplane/pkg/nsmd/network_service_server.go
@@ -41,7 +41,7 @@ func NewNetworkServiceServer(model model.Model, workspace *Workspace, serviceReg
 	}
 }
 
-func (srv *networkServiceServer) getEndpointFromRegistry(ctx context.Context, networkServiceName string) *registry.NSERegistration {
+func (srv *networkServiceServer) getEndpointFromRegistry(ctx context.Context, networkServiceName string, networkServiceLabels map[string]string) *registry.NSERegistration {
 	// Get endpoints
 	discoveryClient, err := srv.serviceRegistry.NetworkServiceDiscovery()
 	if err != nil {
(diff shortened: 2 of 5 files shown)


edwarnicke:
I rebased on #537 to make this easier to review :)

controlplane/pkg/model/model.go
```diff
@@ -35,7 +36,7 @@ type Model interface {
 	GetEndpoint(name string) *registry.NSERegistration
 	AddEndpoint(endpoint *registry.NSERegistration)
 	DeleteEndpoint(name string) error
-	SelectEndpoint(nsName string) (*registry.NSERegistration, error)
+	SelectEndpoint(nsName string, nsLabels *map[string]string) (*registry.NSERegistration, error)
```
edwarnicke: So I'd actually suggest

SelectEndpoing(*registry.NetworkService)

that way you get the name, the matches, the destinations etc :)

controlplane/pkg/model/model.go
```diff
@@ -153,13 +179,13 @@ func (i *impl) SelectEndpoint(nsName string) (*registry.NSERegistration, error)
 	}
 	i.Lock()
 	defer i.Unlock()
-	idx := i.roundRobin[nsName] % len(nseRegistrations)
-	endpoint := nseRegistrations[idx]
-	if endpoint == nil {
-		return nil, errors.New("should not see this error, scaffolding called")
+
+	if len(*nsLabels) > 0 {
```
edwarnicke: If you pass in ns *registry.NetworkService ... then you can do this same len(ns.GetMatches()) test here.

controlplane/pkg/model/model.go
```diff
@@ -143,7 +144,32 @@ func (i *impl) DeleteEndpoint(name string) error {
 	return fmt.Errorf("no endpoint with name: %s", name)
 }
 
-func (i *impl) SelectEndpoint(nsName string) (*registry.NSERegistration, error) {
+func (i *impl) selectEndpointRR(nsName string, registrations *[]*registry.NSERegistration) (*registry.NSERegistration, error) {
```
edwarnicke: if this becomes:

func (i *impl) selectEndpo

controlplane/pkg/model/model.go
```diff
@@ -143,7 +144,32 @@ func (i *impl) DeleteEndpoint(name string) error {
...
+	idx := i.roundRobin[nsName] % len(nseRegistrations)
+	endpoint := nseRegistrations[idx]
+	if endpoint == nil {
+		return nil, errors.New("should not see this error, scaffolding called")
+	}
+	i.roundRobin[nsName] = (idx + 1) % len(nseRegistrations)
+	return endpoint, nil
+}
+
+func (i *impl) selectEndpointMatch(nsName string, nsLabels *map[string]string, registrations *[]*registry.NSERegistration) (*registry.NSERegistration, error) {
```
edwarnicke: If this becomes:

func (i *impl) selectEndpointMatch(ns *registry.NetworkService, sourceLabels map[string]string,registrations *[]*registry.NSERegistration)

Then you can loop over it something like:

```go
var registration *registry.NSERegistration
var match *registry.Match
// Loop over the matches
for _,match = range ns.GetMatches() {
  // Note: pseudocode, so add in the appropriate safety checks :)
  // Loop over the SourceSelector labels in the match
  for key,value := range match.GetSourceSelector() {
      // If the sourceLabels fail to completely match the sourceSelector, we don't match, and we break from ths
      // sourceSelector loop
      if sourceLabel[key] != value {
        match = nil
        break
      }
  }
  var candidateRegistrations []*registry.Destinations
  // If we still have a match that isn't nil, match.GetRoutes() will give us the list of destinations
  // if match == nil, then this loop is a noop.
  for _,destination := range match.GetRoutes() {
       // Loop through the registrations looking for ones that have matching DestinationLabels
        for _,registration := range registrations {
             isCandidate = true
             for key,value := registration.GetLabels() {
                   // If the registration Labels don't match the DestinationSelector, break and declare it so
                   if destination.GetDestinationSelector()[key] != value {
                       isCandidate = false
                       break
                   }
             }
             if isCandidate {
                candidateRegistrations := append(candidateRegistrations,registration)
             }
        }
    }
    // until we get to LBing by weight, just return the first one:
    return candidateRegistrations[0]
}
```

edwarnicke: Note: this shows the full logic (I think) don't necessarily do this all in one long nested loop ;)

controlplane/pkg/model/model.go
nickolaev: I'm not sure how this can be invoked in this case. Please check my comment in `network_service_server.go` below.

controlplane/pkg/nsmd/network_service_server.go
```diff
@@ -67,7 +67,8 @@ func (srv *networkServiceServer) Request(ctx context.Context, request *networkse
 	dpCtx, dpCancel := context.WithTimeout(context.Background(), nseConnectionTimeout)
 	defer dpCancel()
 
-	endpoint, err := srv.model.SelectEndpoint(request.GetConnection().GetNetworkService())
+	labels := request.GetConnection().GetLabels()
+	endpoint, err := srv.model.SelectEndpoint(request.GetConnection().GetNetworkService(), &labels)
```
nickolaev: @edwarnicke `request.GetConnection().GetNetworkService()` returns the `string` name of the requested service.
In order to get `registry.NetworkService` as you suggest, I need to execute some lookup. I guess this could be the model's `GetNetworkServiceEndpoints`, but that returns a slice.
Or am I getting it completely wrong?

edwarnicke:
Re selectors: see:
https://github.com/ligato/networkservicemesh/blob/1cc12b71b7ca2a2c330c66ead88f20f6213fcd67/controlplane/pkg/selector/selector.go

https://github.com/ligato/networkservicemesh/blob/1cc12b71b7ca2a2c330c66ead88f20f6213fcd67/controlplane/pkg/selector/round_robin_selector.go

https://github.com/ligato/networkservicemesh/blob/1cc12b71b7ca2a2c330c66ead88f20f6213fcd67/controlplane/pkg/model/model.go#L225

and

https://github.com/ligato/networkservicemesh/blob/1cc12b71b7ca2a2c330c66ead88f20f6213fcd67/controlplane/pkg/nsmd/network_service_server.go#L59

controlplane/pkg/selector/selector.go
```diff
@@ -3,5 +3,5 @@ package selector
 import "github.com/ligato/networkservicemesh/controlplane/pkg/apis/registry"
 
 type Selector interface {
-	SelectEndpoint(ns *registry.NetworkService, networkServiceEndpoints []*registry.NetworkServiceEndpoint) *registry.NetworkServiceEndpoint
+	SelectEndpoint(ns *registry.NetworkService, nsLabels map[string]string, networkServiceEndpoints []*registry.NetworkServiceEndpoint) *registry.NetworkServiceEndpoint
```
edwarnicke: Perhaps we could just pass in the Connection and list of endpoints.  This gives us maximum flexibility.  We could for example choose to allow a Connection to specify an NSE by name.  Thoughts?

edwarnicke:
Curious, how have you tested this?

controlplane/pkg/selector/selector.go
nickolaev: I thought of that too. I'll play with the idea and see how it works.

nickolaev: @edwarnicke I've implemented it and patch pushed here.

PR #551, opened 2018-11-28
nsmdp refactorings no.2

platoff:
Dynamically expand nsmdp pool of DeviceIds #496, Refactor nsmdp to have DeviceId match nsmd's id for ClientConnections #495, Refactor nsmdapi to better match device plugin api #494

### DIFF
--- a/controlplane/pkg/apis/nsmdapi/nsmd.proto
+++ b/controlplane/pkg/apis/nsmdapi/nsmd.proto
@@ -4,6 +4,7 @@ package nsmdapi;
 
 // ConnectionRequest is sent by a NSM client to build a connection with NSM.
 message ClientConnectionRequest {
+    string workspace = 1;
 }
 
 // ClientConnectionReply is sent back by NSM as a reply to ClientConnectionRequest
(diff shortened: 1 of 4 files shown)


platoff:
This PR obsoletes #536

controlplane/pkg/apis/nsmdapi/nsmd.proto
```diff
@@ -27,7 +28,15 @@ message DeleteConnectionReply {
 
 }
 
+message EnumConnectionRequest {
```
edwarnicke: Consider using [empty](https://github.com/ligato/networkservicemesh/blob/master/controlplane/pkg/apis/local/connection/connection.proto#L59) here :)

controlplane/pkg/apis/nsmdapi/nsmd.proto
```diff
@@ -27,7 +28,15 @@ message DeleteConnectionReply {
...
+}
+
+message EnumConnectionReply {
+    repeated string workspace = 1;
+}
+
 service NSMD {
     rpc RequestClientConnection (ClientConnectionRequest) returns (ClientConnectionReply);
+    rpc EnumConnection (EnumConnectionRequest) returns (EnumConnectionReply);
```
edwarnicke: We might want to use a prefix like "List" instead of "Enum" (might get confused with actual enums :) ).

PR #559, opened 2018-11-29
Support of routes in ConnectionContext

haiodo:
PR include following changes:
* Make Connection Context a structured/typed one for better error handling and clean picture.
* Adds support for passing Routes for Kernel/Memif interfaces from NSE to NSC
* Adds support for passing Arp Entries for Kernel interface from NSE to NSC

Also have working tests with local NSC/NSM/NSE and test to check full workflow for Remote NSC -> NSE with Routes/Arps.


controlplane/pkg/apis/connectioncontext/connectioncontext.proto
```diff
@@ -0,0 +1,39 @@
...
+    string dst_ip_addr = 1;             /* destination ip address + prefix in format <address>/<prefix> */
+    string gw_addr     = 2;             /* gateway address */
+    uint32 metric      = 3;             /* routing metric (weight) */
+}
+
+message ConnectionContext {
+    string src_ip_addr = 1;             /* source ip address + prefix in format <address>/<prefix> */
+    string dst_ip_addr = 2;             /* destination ip address + prefix in format <address>/<prefix> */
+
+    bool src_ip_reqiured = 3;           /* if specified src_ip is required from NSE. */
```
edwarnicke: type: src_ip_reqiured should be src_ip_required

controlplane/pkg/apis/connectioncontext/connectioncontext.proto
```diff
@@ -0,0 +1,39 @@
...
+// This contains the core Network Service Mesh API definitions for external
+// consumption via gRPC protobufs.
+
+syntax = "proto3";
+
+package connectioncontext;
+option go_package = "connectioncontext";
+
+message Route {
+    string dst_ip_addr = 1;             /* destination ip address + prefix in format <address>/<prefix> */
```
edwarnicke: Lets call this "prefix" not "dst_ip_addr".

For the moment lets start simply with the single field "prefix".  As a general rule, all routes from an NSE  should point to the connection to the NSE.

dataplane/vppagent/pkg/converter/memif_interface_converter.go
```diff
@@ -72,5 +72,15 @@ func (c *MemifInterfaceConverter) ToDataRequest(rv *rpc.DataRequest) (*rpc.DataR
 			SocketFilename: path.Join(fullyQualifiedSocketFilename),
 		},
 	})
+
+	// Process static routes
+	for _, route := range c.Connection.GetContext().GetRoutes() {
+		rv.LinuxRoutes = append(rv.LinuxRoutes, &l3.LinuxStaticRoutes_Route{
+			DstIpAddr:	route.DstIpAddr,
+			GwAddr: route.GwAddr,
```
edwarnicke: For the moment lets just make the route point to c.Connection.GetContext().DstIpAddr and leave off the metric

controlplane/pkg/apis/connectioncontext/connectioncontext.proto
haiodo: I supposed dst_ip_addr is a target so "prefix" is confusing for me. Maybe better it to have named "target_address"?

controlplane/pkg/apis/connectioncontext/connectioncontext.proto
haiodo: Thanks, Fixed.

dataplane/vppagent/pkg/converter/memif_interface_converter.go
haiodo: Done.

edwarnicke:
@haiodo This look good, but needs a rebase

controlplane/pkg/apis/connectioncontext/helpers.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+	for _, route := range c.GetRoutes() {
+		if route.GetPrefix() == "" {
+			return fmt.Errorf("ConnectionContext.Route.Prefix is required and cannot be empty/nil: %v", c)
+		}
+	}
+	return nil
+}
+
+func (c *ConnectionContext) IsCompleteAgainsOriginal(original *ConnectionContext) error {
```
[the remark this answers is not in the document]
haiodo: Yes, MeetsRequirements looks better. Thanks.

controlplane/pkg/apis/remote/connection/helpers.go
```diff
@@ -38,6 +39,16 @@ func (c *Connection) IsValid() error {
 	return nil
 }
 
+func (c *Connection) UpdateContext(newContext *connectioncontext.ConnectionContext) error {
+	oldCtx := c.Context
+	c.Context = newContext
```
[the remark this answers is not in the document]
haiodo: Hmps, good catch, I will revert changes in case of any errors found.

edwarnicke:
ip route

Which outputs routes in the nsc pods 

Can be used to see if routes got programmed correctly in the nsc pods

haiodo:
Fixed review changes, also merged #571 into this PR, since very close. Still trying to make routes working.

haiodo:
@edwarnicke I think I also need to add validation for prefixes on NSE, to make IP/mask STRICT requirement. Do you have any code to validate it property already somewhere?

edwarnicke:
@haiodo Perhaps [net.ParseCIDR](https://golang.org/pkg/net/#ParseCIDR)

edwarnicke:
@haiodo How are things going?  I had read your last comment as "more to come" on this PR.

haiodo:
Finally it seems working for Routes, but not for Ip_Neightbour. 

@edwarnicke I think better would be to complete with this PR and do fixes to Ip_Neightbour into a separate one. I've trying to debug this inside vpp and could not found a support for them at all.

Probable I've misses something, but in codebase there is not obvious place where LinuxArpEntries are used

haiodo:
Fixed local/remote tests. Was required to pass a temporary root for storing socket files in case of tests.

dataplane/vppagent/pkg/converter/remote_connection_converter.go
```diff
@@ -82,5 +82,18 @@ func (c *RemoteConnectionConverter) ToDataRequest(rv *rpc.DataRequest) (*rpc.Dat
 		},
 	})
 
+	// Process routes
+	if c.side == SOURCE {
```
edwarnicke: I'm a little confused here... if the c.side == SOURCE in the remoteconnection converter... then the source is remote, and we don't have the namespace here to configure the routes... what am I missing here?

dataplane/vppagent/pkg/converter/utils.go
```diff
@@ -26,3 +27,21 @@ func TempIfName() string {
 	logrus.Infof("Generated unique TempIfName: %s len(TempIfName) %d", rv, len(rv))
 	return rv
 }
+
+func TempRouteName() string {
```
edwarnicke: Question... how do we recover these names to delete the route/arp info on Close() ?

edwarnicke:
This looks good except for my one remaining question about programming routes for remoteConnectionConverter on c.side == SOURCE

dataplane/vppagent/pkg/converter/remote_connection_converter.go
haiodo: Hmps, Let me check. Based on https://github.com/ligato/networkservicemesh/blob/dc17571debd044f378408f0b83413cc973b81ac5/dataplane/vppagent/pkg/converter/cross_connect_converter.go#L40

It is really Source in case of LocalSource. And for our remote case it is so.

dataplane/vppagent/pkg/converter/utils.go
haiodo: On close, interface will be deleted, so I hope it will happen to routes, arp entries, but it will be good idea to have a integration test for this.

dataplane/vppagent/pkg/converter/remote_connection_converter.go
edwarnicke: Right, but its called in:

https://github.com/ligato/networkservicemesh/blob/ebdbac97d13a3ab7add93e24fc4a7accd38d5efc/dataplane/vppagent/pkg/converter/remote_connection_converter.go#L44

A NewRemoteConnectionConverter with side==SOURCE  is only called if its a RemoteSource (not a LocalSource):

https://github.com/ligato/networkservicemesh/blob/ebdbac97d13a3ab7add93e24fc4a7accd38d5efc/dataplane/vppagent/pkg/converter/cross_connect_converter.go#L50

So I'm still confused...

haiodo: Yeh, You are right, I'm also confused about this SOURCE/DEST stuff.

https://github.com/ligato/networkservicemesh/blob/ebdbac97d13a3ab7add93e24fc4a7accd38d5efc/controlplane/pkg/nsmd/network_service_server.go#L284

Here is how I've created it for Remote case, and I've put it as LocalSource, so checked this variant.

I suppose we need update Dataplane to be more clear.

dataplane/vppagent/pkg/converter/utils.go
edwarnicke: @haiodo A NSC may have multiple connections to multiple NSEs... so ConnectionContext related to one may need to be deleted, but not all.  Certainly everything goes away when the netns does... but ConnectionContext stuff provisioned needs to be deleted when the Connection is torn down.

dataplane/vppagent/pkg/converter/remote_connection_converter.go
edwarnicke: @haiodo Yeah... clarity will be good.

The basic idea is this:

SOURCE == "Who requested this connection" ie, an NSC
DESTINATION == "Who was connected to in response to the request" ie, an NSE.

ConnectionContext gets set in the SOURCE, but only if the SOURCE is Local (if its remote, we can't meaningfully set it, its just a tunnel) and *only* if the Mechanism type of the LocalSource is KernelInterface (if its Memif, its not our responsibility).

dataplane/vppagent/pkg/converter/utils.go
haiodo: Have a interface name prefixed routes/arps.

dataplane/vppagent/pkg/converter/remote_connection_converter.go
haiodo: Checked, in remote case indeed this is not required. Thanks.

haiodo:
@edwarnicke  Any changes for this will be merged soon? I've fixed comments from you. Thanks.

controlplane/pkg/tests/connection_context_test.go
```diff
@@ -0,0 +1,77 @@
+package tests
```
nickolaev: Tests! Great!

dataplane/vppagent/pkg/converter/memif_interface_converter.go
```diff
@@ -15,11 +15,11 @@
 package converter
 
 import (
-	fmt "fmt"
+	"fmt"
+	l32 "github.com/ligato/vpp-agent/plugins/vpp/model/l3"
```
nickolaev: Why call this `l32`?

examples/cmd/vppagent-nsc/nsc.go
```diff
@@ -75,8 +76,9 @@ func main() {
 
 		Connection: &connection.Connection{
 			NetworkService: "icmp-responder",
-			Context: map[string]string{
-				"requires": "src_ip,dst_ip",
+			Context: &connectioncontext.ConnectionContext{
+				SrcIpRequired: true,
+				DstIpRequired: true,
```
nickolaev: This way is so much better.

.vagrant.mk
```diff
@@ -46,13 +46,5 @@ vagrant-restart-kublet:
...
-		echo "Loading image $*.tar to master"; \
-		vagrant ssh master -c "sudo docker load -i /vagrant/images/$*.tar" > /dev/null 2>&1; \
-		echo "Loading image $*.tar to worker"; \
-		vagrant ssh worker -c "sudo docker load -i /vagrant/images/$*.tar" > /dev/null 2>&1; \
-	else \
-		echo "Cannot load $*.tar: scripts/vagrant/images/$*.tar does not exist.  Try running 'make k8s-$*-save'"; \
-		exit 1; \
-	fi
\ No newline at end of file
+	@./scripts/load-image.sh $*
```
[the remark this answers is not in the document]
haiodo: Will move to separate PR

Issue #571, opened 2018-12-02
Support IP Neighbor entries in ConnectionContext


edwarnicke:
This is an extension of #555 

Add to ConnectionContext an ```ConnectionContext.IpNeighbors []*IpNeighbor```
Where

```
message IpNeighbor {
   string ip = 1;
   string hardware_address = 2;
}
```

For the case of NSC's with MechanismType == KernelInterface, use [LinuxStaticArpEntries](https://github.com/ligato/vpp-agent/blob/v1.7/plugins/linux/model/l3/l3.proto#L50) to enter those IpNeighbor entries into the namespace of the NSC.

haiodo:
Working solution in https://github.com/ligato/networkservicemesh/pull/559

Issue #555, opened 2018-11-29
Support of routes in ConnectionContext


edwarnicke:
ConnectionContext is a map[string]string, it is currently used for communicating src_ip and dst_ip.

It can also be used to communicate 'routes'

In [connectioncontext](https://github.com/ligato/networkservicemesh/blob/master/controlplane/pkg/apis/connectioncontext/constants.go) build out constants and helpers to validate for routes of the form:

```routes="10.10.5.5/24":"10.1.1.1","5.5.5.5/12":"1.1.1.1"```

basically, quoted json style map.

Also expand to have the ability to contain:

```excluded_prefixes="2.2.2.2/30","10.10.10.1/16"```
Think a quited json style list.  

Things we will need to figure out how to do:

1.  Have the nsmd add the PodCDIR and ServiceCDIR to the "excluded_prefixes" list
2. Have a simple helper function an NSE can use to assess whether an ip or route it wishes to send in the connection context collides with the excluded_prefixes.  This may take the form of an isValid() method that checks if the src_ip, dst_ip, or routes conflict with an excluded_prefix.
3. Have the dataplane set the routes in the requesting NSC's linux namespace.  This will most likely involve the [vppagent Linux L3 API](https://github.com/ligato/vpp-agent/blob/v1.7/plugins/linux/model/l3/l3.proto)

edwarnicke:
What is the obstacle with defining it inside a map?

haiodo:
Put initial thoughts here: https://github.com/ligato/networkservicemesh/pull/559

1. @edwarnicke  I've could not found anything about PodCDIR/ServiceCDIR, advice will be helpful.
2. I've have no idea yet how to handle excluded_prefixes, advice will be helpful.
Found a way using https://github.com/ligato/vpp-agent/blob/d2ae51e30bb34e5dec059de71c2d0f42ecb10a09/plugins/vpp/model/l3/l3.proto#L16 not sure if this is right way.
3. add src/dst ip is required fields, as was in nsc examples. I assume to add validation if NSC is asking for this stuff and NSE is not provided. Thoughts?

Code is not yet tested, still need to add validations, will continue tomorrow.

edwarnicke:
Ah... I was insufficiently clear about excluded_prefixes.

When a Request(Connection) comes into an NSE with excluded_prefixes, the prefixes returned in routes and the src/dst ip in the ConnectionContext cannot be in those prefixes.

So if
[...]

PR #575, opened 2018-12-03
#571

haiodo:
PR is based on #555 so PR for it https://github.com/ligato/networkservicemesh/pull/559 should be merged before this one.


haiodo:
Will merge changes into #559, since some code review modifications are required.

PR #577, opened 2018-12-03
Allow using libvirt with vagrant

matrohon:
Use sshfs for mounting /vagrant directory.
Use generic/debian9 box that is compatible with libvirt
and generic/ubuntu1804 has resolver issues.

### DIFF
--- a/.k8s.mk
+++ b/.k8s.mk
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-K8S_CONF_DIR = k8s/conf/
+K8S_CONF_DIR = k8s/conf
 
 # Need nsmdp and icmp-responder-nse here as well, but missing yaml files
 DEPLOY_NSM = nsmd vppagent-dataplane
--- a/.vagrant.mk
+++ b/.vagrant.mk
@@ -14,7 +14,7 @@
 
 .PHONY: vagrant-start
 vagrant-start:
-	@cd scripts/vagrant; vagrant up
+	@cd scripts/vagrant; vagrant up --no-parallel; source env.sh
 
 .PHONY: vagrant-destroy
 vagrant-destroy:
@@ -22,7 +22,7 @@ vagrant-destroy:
 
 .PHONY: vagrant-restart
 vagrant-restart: vagrant-destroy
-	@cd scripts/vagrant; sleep 2;vagrant up
+	@cd scripts/vagrant; sleep 2;vagrant up --no-parallel
 
 .PHONY: vagrant-suspend
 vagrant-suspend:
--- a/.vagrant.mk
+++ b/.vagrant.mk
@@ -14,7 +14,7 @@
 
 .PHONY: vagrant-start
 vagrant-start:
-	@cd scripts/vagrant; vagrant up --no-parallel; source env.sh
+	@cd scripts/vagrant; vagrant up --no-parallel;
 
 .PHONY: vagrant-destroy
 vagrant-destroy:
--- a/scripts/vagrant/Vagrantfile
+++ b/scripts/vagrant/Vagrantfile
@@ -3,8 +3,13 @@
 
 Vagrant.configure("2") do |config|
 
-  config.vm.synced_folder "../..", "/go/src/github.com/ligato/networkservicemesh"
+  config.vm.synced_folder "../..", "/go/src/github.com/ligato/networkservicemesh", type: "sshfs"
+  config.vm.synced_folder ".", "/vagrant", type: "sshfs"
 
+  config.vm.provider "libvirt" do |v|
+        v.memory = 4096
+        v.cpus = 2
+  end
   config.vm.provider "virtualbox" do |v|
         v.memory = 4096
         v.cpus = 2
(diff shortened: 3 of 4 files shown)


.vagrant.mk
```diff
@@ -14,15 +14,15 @@
 
 .PHONY: vagrant-start
 vagrant-start:
-	@cd scripts/vagrant; vagrant up
+	@cd scripts/vagrant; vagrant up --no-parallel; source env.sh
```
edwarnicke: the source env.sh here will have no effect, as it dies with the shell spawned by make.

matrohon: You're right!
The idea was to export the KUBECONFIG env so that parent shell knows about it. Without that, subsequent kubectl calls will fail. It prevent a fresh install to work out of the box with 'make k8s-deploy' after a 'git clone' of the project.

scripts/vagrant/Vagrantfile
```diff
@@ -3,8 +3,13 @@
 
 Vagrant.configure("2") do |config|
 
-  config.vm.synced_folder "../..", "/go/src/github.com/ligato/networkservicemesh"
+  config.vm.synced_folder "../..", "/go/src/github.com/ligato/networkservicemesh", type: "sshfs"
+  config.vm.synced_folder ".", "/vagrant", type: "sshfs"
```
nickolaev: `config.vm.allowed_synced_folder_types = [:sshfs]`
or you can do it in the `libvirt` section. See below.

scripts/vagrant/Vagrantfile
```diff
@@ -3,8 +3,13 @@
...
 
+  config.vm.provider "libvirt" do |v|
+        v.memory = 4096
+        v.cpus = 2
```
nickolaev: You can do `v.synced_folder_type = 'sshfs'`
This way you don't bind VBox users to `sshfs`.

scripts/vagrant/Vagrantfile
nickolaev: Oh even better I guess you should be able to do something like:
`config.vm.allowed_synced_folder_types = [:virtualbox, :sshfs]`

So when you use libvirt it will fail on the fu=irst available sync type and will try the sshfs one.

matrohon: Thanks, for your review @nickolaev, I'll test your proposals and update my PR.

edwarnicke:
@matrohon  If you rebase on master, the !-z to -n should now be fixed in master.

scripts/vagrant/Vagrantfile
edwarnicke: Second.  I'm all for what it takes to make libvirt work well for you, but sshfs would also not be good for VMWare Fusion users.

scripts/vagrant/Vagrantfile
```diff
@@ -4,7 +4,13 @@
 Vagrant.configure("2") do |config|
 
   config.vm.synced_folder "../..", "/go/src/github.com/ligato/networkservicemesh"
+  config.vm.synced_folder ".", "/vagrant"
+  config.vm.allowed_synced_folder_types = [:virtualbox, :sshfs]
```
edwarnicke: Will this also work for VMWare?

matrohon:
I don't know, maybe @nickolaev can answer

scripts/vagrant/Vagrantfile
matrohon: I don't know, maybe @nickolaev can answer

nickolaev: Well `sshfs` should work on "any"® provider.

edwarnicke: @nickolaev Yeah... but poorly I would expect...

scripts/vagrant/Vagrantfile
```diff
@@ -4,7 +4,13 @@
 Vagrant.configure("2") do |config|
 
   config.vm.synced_folder "../..", "/go/src/github.com/ligato/networkservicemesh"
+  config.vm.synced_folder ".", "/vagrant"
+  config.vm.allowed_synced_folder_types = [:virtualbox, :sshfs, :vmware]
```
nickolaev: Please prioritize the higher performance options:
[:virtualbox, :vmware, :sshfs]

scripts/vagrant/Vagrantfile
```diff
@@ -17,7 +23,7 @@ Vagrant.configure("2") do |config|
     v.vmx["memsize"] = "4096"
     v.vmx["numvcpus"] = "2"
   end
-  config.vm.box = "bento/ubuntu-18.04"
+  config.vm.box = "generic/ubuntu1604"
```
nickolaev: @matrohon we discussed on the call that it might be a good idea if we make this conditional. I.e. when the provider is libvirt, use `generic/ubuntu1604` otherwise stick with bento 1804. What do you think?

.k8s.mk
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-K8S_CONF_DIR = k8s/conf/
+K8S_CONF_DIR = k8s/conf
```
nickolaev: Is this needed.

scripts/vagrant/Vagrantfile
matrohon: done

.k8s.mk
matrohon: this is just a cleanup

PR #584, opened 2018-12-05
Fix error in switch to not building skydive container

edwarnicke:


matrohon:
+1 we missed it  for the skydive agent containers on #580

PR #586, opened 2018-12-05
Add skydive port-forward makefile rule

matrohon:


edwarnicke:
@matrohon Could you say a few words about how to use this?  This looks like it might be cooler than what I'm doing with make skydive-url... but I can't tell :)

matrohon:
once you run "make skydive-port-forward" you'll be able to access skydive with https://127.0.0.1:8082

It gives you a stable url to bookmark

edwarnicke:
Woot!

PR #588, opened 2018-12-05
WIP: Implement 553 fixes

nickolaev:


edwarnicke:
@nickolaev I believe I've fixed this.

It came down to two things:  #589 (which is why I rebased your patch on master)

And an issue with how fully qualified socket directories are computed by CrossConnectConverter (which is fixed and explained in the latest commit here).

PR #590, opened 2018-12-06
Implement memif proxy

lobkovilya:


edwarnicke:
In the course of trying out this PR I saw some other strange behavior that *probably* isn't related to this PR, and *probably* is some existing bug... but would like to chase it down before merging.

The direct memif memif *does* work for me though :)

edwarnicke:
Confirmed the strange things I was seeing were not from this patch.

One (not showing up in skydive) is fixed by #599

The other (vxlan issues) by #589.

Neither are actually issues with this PR.  I just found them while testing this PR

PR #597, opened 2018-12-07
Makefile: split deployment to infra, icmp example and vpn example

nickolaev:


nickolaev:
#596 changes docs. If this patch gets merged, we'll need to uptake the docs with the new targets.

.k8s.mk
```diff
@@ -41,6 +48,15 @@ include .docker.mk
 .PHONY: k8s-deploy
 k8s-deploy: k8s-delete $(addsuffix -deploy,$(addprefix k8s-,$(DEPLOYS)))
 
+.PHONY: k8s-deploy-infra
```
edwarnicke: The pattern for everything else is k8s-${something}-deploy ... is there a reason to reverse that here?

nickolaev: Oh I see. Well I was thinking that `-deploy` is the master deployer then you have "child" deployments with descend. Maybe it gets too complicated like that. Will change it. Thanks.

edwarnicke: Honestly, the reason I did:

k8s-${something}-deploy was because it made it easier to change the verb (say k8s-${something}-save).  It can easily be argued either way.  My primary concern is consistency :)

edwarnicke:
@nickolaev Also, #596 is merged, so you can change the docs in this patch as well if you like :)

PR #603, opened 2018-12-09
WIP: bridge domain

fkautz:


examples/cmd/vppagent-bridge-domain-nse/service.go
```diff
@@ -80,18 +80,30 @@ func New(vppAgentEndpoint, baseDir, ip, bridgeDomainName string) networkservice.
 		shg = rand.Int()
 	}
 
+	bridgeDomain := &l2.BridgeDomains_BridgeDomain{
+		Name:                 bridgeDomainName,
+		Flood:                true,
+		UnknownUnicastFlood:  true,
+		Forward:              true,
+		Learn:                true,
```
tfherbert: For a simple bridge example, I generally turn off learning especially since flooding is set. It probably won't due any harm though.

PR #604, opened 2018-12-10
Introduce the NSM SDK

nickolaev:
WIP. For review.

This package splits the common functionality needed to build a NSE, NSC or a NSE/NSC (i.e. such that can be used in a service composition), into a separate folder. It allows for simplification of the user code, focusing on the business logic and not establishing the NSE/NSC specifics.

As a result, the NSM native applications, such as `vppagent-firewall-nse` are faster to code and easier to explain.


nickolaev:
This PR's ambition is to get #611 implemented.

k8s/conf/icmp-responder-nse.yaml
```diff
@@ -28,7 +28,9 @@ spec:
           image: networkservicemesh/icmp-responder-nse:latest
           imagePullPolicy: IfNotPresent
           env:
-            - name: NSE_LABELS
+            - name: ADVERTISE_NSE_NAME
```
edwarnicke: I like ADVERTISE_NSE_NAME and ADVERTISE_NSE_LABELS

k8s/conf/nsc.yaml
```diff
@@ -19,10 +19,10 @@ spec:
           image: networkservicemesh/nsc:latest
           imagePullPolicy: IfNotPresent
           env:
-            - name: NSC_LABELS
+            - name: OUTGOING_NSC_LABELS
```
edwarnicke: I like OUTGOING_NSC_NAME and OUTGOING_NSC_LABELS

edwarnicke:
Looking through this I see a few parts:

1)  Configuration: It extracts info from env variables to fill in config if not explicitly provided
2)  NSC (aka client): It looks like it will create Connections to a particular Network Service
3)  NSE (aka endpoint): It looks like it will advertise and accept connections for a particular Network Service.
4)  Some things around IPAM

And this seems to be focused heavily on the case where we want to do something transparent in an NSE via L2 Cross Connect.

Would it make sense to structure it a bit differently:

First directory wise have 
1. sdk/nsc/cf - cf stands for ConnectionFactory.  So this is a ConnectionFactory for Network Services.
2. sdk/nse - NSE related helpers
3. sdk/ipam - IPAM related helpers

In sdk/nsc/cf have a NSConnectionFactory something like:

```go
type NSConnectionFactory interface {
      Request() Connection
      Close(Connection)
}
```
effectively it becomes a wrapper for some configuration (and possibly logic) around manufacturing NetworkServices.  You could then do some particular cases like:

```go
type EnvNSCConnectionFactory struct { ... }
func NewEnvNSCConnectionFactory() *EnvNSCConnectionFactory { ... }
```
That would read Env Variables etc. 

On the NSE side you could do something like:

```go
type CompositeNSE interface {
        local.NetworkService
        AddHandler(*local.NetworkService)
        RemoveHandler(*local.NetworkService)
}

type compositeNSE struct {
      handlers []*local.NetworkService
}

func NewCompositeNSE() *CompositeNSE {
...
}
```

Where on CompositeNSE.Request(NetworkServiceRequest) it does:

```go
for handler := range c.handlers {
      nsrequest.connection := handler.Request(ctx,nsrequest)
}
```

and on CompositeNSE.Close(ctx,Connection) it does:
```go
for handler := range c.handlers {
     connection := handler.Close(ctx,connection)
}
```

Then we could write things like:

```go
type VPPInterfaceNSE struct {...}
```

that just plugs the request into the VPP Dataplane

```go
type NSCConnectionNSE struct {...}
```
That just creates a NSCConnection

```go
type CrossConnectNSE struct {...}
```

That CrossConnects NSCConnection, etc.

And use the Composite NSE to chain them together:

cnse := NewCompositeNSE()
cnse.AddHandler(NewVPPInterfaceNSE())
cnse.AddHandler(NewNSCConnectionNSE()) 
cnse.AddHandler(NewCrossConnectNSE())

edwarnicke:
Apologies, didn’t mean to close, just comment :)

edwarnicke:
Refinement of the idea of CompositeNSE ...

```go
type CompositeNSE interface {
     local.NetworkService
     SetNext(local.NetworkService)
}
```

Rather than having a 'list' of Handlers... simply set the 'Next' handler.

We could provide a 'base' CompsiteNSE:

```go
type BaseCompositeNSE struct {
     next local.NetworkService
}

func (b *BaseCompositeNSE) SetNext(next local.NetworkService) { b.next = next }

func (b *BaseCompositeNSE) Next() local.NetworkService { return b.next }
```

Then you can write an implementation of a CompositeNSE as:

```go
type MyCompositeNSE struct {
    BaseCompositeNSE
    ... // Whatever else you need
}

func (c *MyCompositeNSE) Request(ctx context.Context, request local.NetworkServiceRequest) {
     // Whatever we do on Request, including mungin request
     // Add any necessary context for subsequent NSEs in the Composite
     nextCtx := context.WithValue(ctx,key,value)
     connection:=  c.Next().Request(nextCtx, request)
    // Whatever we need to do after any subsequent NSEs
    return connection
}

// Same with 'Close'
```

Then we could do something like FirewallNSE as:

```go
nse := NewConnectToVPP().Next(
           NewApplyACL. Next(
                 NewCreateConnection(config).Next(
                         NewCrossConnect()
                 )
            )
     )              
)
```

Essentially... we can 'compose' primitive pieces of NSE implementation *inside* an executable NSE, and write a small library of those primitive pieces :)

Thoughts?

nickolaev:
@edwarnicke in go I can't do multi-folder packages. If I want to split, the content I'd have to actually have different packages. That should not be a problem of course.

nickolaev:
I do agree that the SDK was designed to wrap the existing examples commomn code, which influenced its design. However, I have my doubts (and probably misunderstandings) about your proposals.

1. I deliberately chose to not use NSC and NSE and instead use client and endpoint. I think we already have too many acronyms starting with NS and it only gets confusing if we expose them to the SDK. I am ready to discuss which naming is better.

2. Putting the ConnectionFactory under NSC(client) is not an obvious choice. If we want a connection factory, let's keep it in `sdk/common/cf` package.

I'll refactor it a bit in the direction you're suggestion while keeing clinet and endpoint for now. Let's see how it ends up be.

nickolaev:
@edwarnicke was looking at your proposal to change the client to use a connection factory.
The thing is that I want to have let clients know when the connection is created or closed. That is why I have the concept of a `backend` which essentially is an interface that the Client can supply at the creation time. Then the related New, Connect and Close functions are called on the respected events.
I think that the best illustration is the VPP NSC example here:

https://github.com/nickolaev/networkservicemesh/blob/nscomposer/examples/cmd/vppagent-nsc/nsc.go

Again, I understand that what I propose is reflecting the current examples. Still this particular one fits very well and am not sure how this can be migrated to the propose ConnectionFactory approach.

nickolaev:
On the other hand, the NSE composition that you suggest is so appealing:

```Go
nse := NewConnectToVPP().Next(
           NewApplyACL(aclConfig). Next(
                 NewCreateConnection(nscConfig).Next(
                         NewCrossConnect()
                 )
            )
     )              
)
```

But I am a bit sceptical about letting the implementators call the next registered guy:

```Go
connection:=  c.Next().Request(nextCtx, request)
```

Maybe we can split the Request to `top` and `bottom` half function?

edwarnicke:
> Maybe we can split the Request to `top` and `bottom` half function? And call the Next() in between.

I don't quite understand the splitting it into a 'top' and 'bottom' half function?  Could you say more?

edwarnicke:
> @edwarnicke was looking at your proposal to change the client to use a connection factory.

My gut is telling me that this gets to be an easier discussion after we think through the CompositeEndpoint part of the discussion :)  So don't take what I said there to seriously just yet :)

examples/cmd/icmp-responder-nse/nse.go
```diff
@@ -15,150 +16,43 @@
...
-	registration := &registry.NSERegistration{
-		NetworkService: &registry.NetworkService{
-			Name:    networkServiceName,
-			Payload: "IP",
-		},
-		NetworkserviceEndpoint: nse,
-	}
+	composite := composite.NewMonitorCompositeEndpoint(nil).SetNext(
+		composite.NewIpamCompositeEndpoint(nil).SetNext(
+			composite.NewConnectionCompositeEndpoint(nil)))
```
edwarnicke: This makes me very very happy!

Issue #622, opened 2018-12-21
Properly handle NSC/NSE dies


lobkovilya:
Handle NSC/NSE dies according to https://github.com/ligato/networkservicemesh/issues/618#issuecomment-448801302 and https://github.com/ligato/networkservicemesh/issues/618#issuecomment-448825428

PR #631, opened 2018-12-25
Issue #622: Properly handle Close() on NSMD

lobkovilya:
Implement Close() for local and remote cases. Also for deleting resource without error in vppagent we have to create separate DataRequests for LinuxInterfaces, Interfaces and Xcons in particular order. So I have changed converters to return list of DataRequests


controlplane/pkg/nsmd/nsmd_crossconnect_client.go
```diff
@@ -38,6 +38,12 @@ type nsmMonitorCrossConnectClient struct {
 	dataplanes map[string]*dataplaneCrossConnectInfo
 }
 
+func (client *nsmMonitorCrossConnectClient) ClientConnectionAdded(clientConnection *model.ClientConnection) {
```
nickolaev: Please add `error` return type here. Since the function is empty, just return `nil`.
Handel the error accordingly when you call the function.

controlplane/pkg/model/model.go
```diff
@@ -20,13 +21,24 @@ type Dataplane struct {
 	RemoteMechanisms []*remote.Mechanism
 }
 
+type ClientConnection struct {
```
edwarnicke: It might be a good idea to store which 'side' of the Xcon the connection is on, which would allow retrieval of the complete Connection object.

Issue #635, opened 2019-01-02
Add OpenTracing support


platoff:
Add [OpenTracing](https://opentracing.io) support to get more insight into what's happening across components and connections.

PR #636, opened 2019-01-02
OpenTracing support #635

platoff:
This is initial contribution exposing Jaeger service in k8s cluster and nsmd logging client requests without details.


edwarnicke:
You may want to look at [this](https://github.com/grpc-ecosystem/grpc-opentracing/tree/master/go/otgrpc) - it looks simpler and cleaner in terms of the internal code changes.

platoff:
Thank you, hopefully, it would be helpful to propagate span context across all components.

edwarnicke:
@platoff Exactly... I share your concerns about doing big expensive (and thus intrinsically fragile) changes for OpenTracing support... but this *looks* like it will be fairly light.

That said... just debugged an issue on IRC this afternoon with someone for which OpenTracing would have been amazing... :)

platoff:
Definitely, OpenTracing seems to be amazing to debug some issues :)

edwarnicke:
@platoff What I realized today, is I am almost always doing 'tracing' via logs.  Which is a really hard way to do it.

platoff:
@edwarnicke yeah, I believe it's already hard to impossible without too much time invested, OpenTracing will improve this, we just need to pass more relevant information there.

edwarnicke:
@platoff Yep... good news is much of what we do is already grpc, so Interceptors capturing goes in and goes out will get a *lot* of what we want.  Also, looks like much of it can be sent with grpc context... so little to no change in existing grpc implementations.

platoff:
@edwarnicke let me please dig more into it, I'm guessing we will need to pass context from server's request handler to client calls by hands anyway, I will report more on this.

platoff:
Here is I already see continuous client requests coming in and finishing in 5 secs (timeout?) do not know yet if it's because vpn-gateway-nsc crashes on my machine...
screen shot 2019-01-02 at 9 32 50 pm

edwarnicke:
@platoff #631 broke master, #637 reverts it :)

But net-net ... yes... good example of seeing things easily :)

edwarnicke:
#637 has been merged, master should be happy again.

platoff:
Here is what I have so far :)
screen shot 2019-01-08 at 2 15 16 pm

.k8s.mk
```diff
@@ -15,7 +15,7 @@
 K8S_CONF_DIR = k8s/conf/
 
 # Need nsmdp and icmp-responder-nse here as well, but missing yaml files
-DEPLOY_NSM = nsmd vppagent-dataplane
+DEPLOY_NSM = jaeger nsmd vppagent-dataplane
```
edwarnicke: Could we move jaeger to DEPLOY_INFRA ? similar to what is done with skydive via DEPLOY_MONITOR... perhaps a DEPLOY_TRACING?

controlplane/cmd/nsmd/nsmd.go
```diff
@@ -44,4 +48,6 @@ func main() {
 		wg.Done()
 	}()
 	wg.Wait()
+
+	closer.Close()
```
edwarnicke: Perhaps do this with a 'defer' right after ```tracer, closer := tools.InitJaeger("nsmd")```

edwarnicke:
@platoff I played with this a bit and it's *super* cool.

It does look like we should probably still instrument:
* vppagent-nsc
* vppagent-icmp-responder
* vppagent-firewall-nse

One other question... is there any way to capture the arguments and returns as part of the trace?

edwarnicke:
@platoff Also... it looks like the call to the vppagent-dataplane from the nsmd in response to a remote.NetworkService.Request is not getting instrumented.

platoff:
@matrohon Is absence of save and load-images targets break any of machinery? OpenTracing is a library we compile against so there are no agent images, just a server which is supplied as pod.

matrohon:
@platoff yes, make k8s-load-images is failing without this target. But then I tested on master, and this machinery is failing too, so maybe we need to cleanup the machinery.

platoff:
@matrohon Thank you! Added k8s-jaeger-load-images NOOP to conform make machinery

PR #637, opened 2019-01-02
Revert "Issue #622: Properly handle Close() on NSMD (#631)"

edwarnicke:
This reverts commit c4d7adc61900f504b2c8dd84cd6878b31378b965.


edwarnicke:
#631 changed the basic API for Converters to return a list of DataRequests rather than a single one.  I have no idea if this is a good or bad idea, but I do know it breaks downstream consumers and results in failures.  So for the moment reverting #631 while we sort out what we want the proper behavior to be and making sure whatever we do is handled properly in downstream consumers.

PR #640, opened 2019-01-07
Issue #622: Properly handle NSC/NSE dies

lobkovilya:
Created class `ClientConnectionManager`, that represent kind of "service layer" for our application. It provides some high-level logic over the model and implements methods:
```
UpdateClientConnectionState(clientConnection *model.ClientConnection, state crossconnect.CrossConnectState) {}
DeleteClientConnection(clientConnection *model.ClientConnection, closeDataplane bool, closeEndpoint bool) {}
GetClientConnectionByXcon(xconId string) *model.ClientConnection {}
GetClientConnectionByDst(dstId string) *model.ClientConnection {}
GetClientConnectionsByDataplane(name string) []*model.ClientConnection {}
```
And also implements some auxiliary methods:
```
CloseRemotes(clientConnection *model.ClientConnection, closeDataplane bool, closeEndpoint bool) {}
CloseXconOnDataplane(clientConnection *model.ClientConnection) error {}
CloseXconOnEndpoint(clientConnection *model.ClientConnection) error {}
```

Modified class `NsmMonitorCrossConnectClient`. It listens to `Model` and when new Dataplane or new NSM are added, `NsmMonitorCrossConnectClient` starts go routine with cross-connect monitor client.


edwarnicke:
@lobkovilya Looks like a rebase is needed.

lobkovilya:
Have some ideas for refactoring in this PR, will rebase a bit later

controlplane/pkg/apis/crossconnect/crossconnect.proto
```diff
@@ -14,22 +14,31 @@ enum CrossConnectEventType {
 
 message CrossConnectEvent {
     CrossConnectEventType type = 1;
-    map<string,CrossConnect> cross_connects = 2;
+    map<string, CrossConnect> cross_connects = 2;
+}
+
+enum CrossConnectState {
```
edwarnicke: I haven't fully wrapped my head around your PR yet, so you may have already done something exceptionally clever here... but if we put the state on the CrossConnect rather than the connections, how does the source nsmd find out that the dst nsmd has marked the remote connection as DOWN?

lobkovilya: Actually nothing clever I haven't done. I just subscribe src-nsmd to cross-connect monitor of dst-nsmd. I thought that we can leave only cross-connect monitor and eliminate connection monitor for simplicity reason. But redoing doesn't take long

PR #646, opened 2019-01-10
Dependabot/dep/k8s.io/kubernetes 1.13.1

haiodo:


haiodo:
Not required, since already merged.

PR #650, opened 2019-01-14
Remove privilege from vppagent based NSEs

edwarnicke:


nickolaev:
OK, so in the end we just need to disable the linux plugin. That sounds pretty reasonable. Merging.

PR #651, opened 2019-01-14
Icmp responder ns eprefix pool

haiodo:
Property switch to use Prefix Pool for ICM Responder NSE demo project.

+ Fixes some issues and add more tests.


test/kube_testing/pods/nsc.go
```diff
@@ -47,6 +47,8 @@ func NSCPod(name string, node *v1.Node, env map[string]string) *v1.Pod {
 						"tail", "-f", "/dev/null",
 					},
 				},
+			},
+			InitContainers: []v1.Container {
```
haiodo: This caused few issues during testing.

haiodo:
@edwarnicke Please take a look if you will have a moment. Thanks.

scripts/vagrant/Vagrantfile
```diff
@@ -3,8 +3,10 @@
 
 Vagrant.configure("2") do |config|
 
-  config.vm.synced_folder "../..", "/go/src/github.com/ligato/networkservicemesh"
-  config.vm.synced_folder ".", "/vagrant"
+  config.vm.synced_folder "../..", "/go/src/github.com/ligato/networkservicemesh",
```
nickolaev: Why we need this change here?
Can you put it in a separate PR and explain.

nickolaev:
Can you please split everything under `examples` in a separate PR.
We can do 3 PRs:
 - `controlplane/pkg/prefix_pool/prefixpool.go` + `tests/`
 - `examples/`
 - Vagrant changes if we need them

scripts/vagrant/Vagrantfile
haiodo: Without it, one of variants is ignored, in my case it is /vagrant and images are not published, happens sometimes.

Issue #669, opened 2019-01-18
#640 introduces monitoring problems


nickolaev:
After merging #640 "Issue #622: Properly handle NSC/NSE dies (#640)" the monitoring for skydive is not behaving as expected. Reverting 605d99b8e2f1afc4f538a08d38250db5d0fb86b9 solves the problems and retruns back skydive to its normal operation.

nickolaev:
@lobkovilya, please check. @matrohon can provide more info if needed.

matrohon:
here is the way to reproduce this bug:

```
make vagrant-start
make k8s-save
make k8s-infra-deploy
make k8s-vpn-deploy
```

edwarnicke:
@matrohon Was this closed by #673 ?

matrohon:
There is still an issue with skydive that doesn't display crossconnects properly after skydive is restarting. Let me test further to see if it comes from the nsmd monitoring stack or from the skydive probe.

matrohon:
I've fixed the skydive bug. Still, something is weird with the crossconnect-monitor container: when it starts and no crossconnect exists, it doesn't logs empty "INITIAL_STATE_TRANSFER" messages. @lobkovilya Is it expected?

matrohon:
#691 aims at fixing this "issue"

edwarnicke:
Merged #691, closing.

PR #673, opened 2019-01-18
Issue #669: Fix bug with cross-connect monitor

lobkovilya:
Bug was related to synchronization in nsmd. Sometimes it received CrossConnectEvent from dataplane before CrossConnect was saved to Model and ignored it.

Fix: save ClientConnection to Model before Request to dataplane. Cross-connect for that ClientConnection will be set later, when appropriate event from dataplane will be received.


matrohon:
I just tested this patch. It is working fine, thanks @lobkovilya

PR #675, opened 2019-01-21
Migrate to go modules

fkautz:


fkautz:
@nickolaev I've set it up so that mod.go gets pushed by itself, followed by `go mod download` which prevents the need to redownload all deps each time. Only when deps have changed does it trigger a redownload.

For team or build environments, this feature will also help solve the problem:

https://tip.golang.org/cmd/go/#hdr-Module_proxy_protocol

PR #677, opened 2019-01-21
Stability fixes

nickolaev:
A couple of stability fixes found during more extensive testing.


controlplane/pkg/model/model.go
```diff
@@ -283,7 +284,10 @@ func (i *impl) AddDataplane(dataplane *Dataplane) {
 
 func (i *impl) DeleteDataplane(name string) {
 	i.Lock()
-	dataplane := i.dataplanes[name]
+	dataplane, ok := i.dataplanes[name]
+	if !ok {
+		return
+	}
 	delete(i.dataplanes, name)
 	i.Unlock()
```
edwarnicke: You need to convert this to a defer i.Unlock() further up, otherwise if the key isn't found, you will return before releasing the lock.

sdk/endpoint/composite/connection.go
```diff
@@ -87,6 +88,14 @@ func (cce *ConnectionCompositeEndpoint) Close(ctx context.Context, connection *c
 	return &empty.Empty{}, nil
 }
 
+func (cce *ConnectionCompositeEndpoint) generateIfName() string {
+	ifName := "nsm" + cce.id.MustGenerate()
```
edwarnicke: FYI... ifNames are limited to 15 chars...

nickolaev: That's fine. The UUID generated should be 10 chars.

edwarnicke: I figured :)  I just know it drove me nuts when I hit it (which is why we now check in the dataplane and issue a super super clear error if you try to go to big :) ).

sdk/endpoint/composite/connection.go
```diff
@@ -87,6 +88,14 @@ func (cce *ConnectionCompositeEndpoint) Close(ctx context.Context, connection *c
...
+	ifName = strings.Replace(ifName, "-", "", -1)
```
edwarnicke: Why are we doing this twice?

nickolaev: Right, I am not even sure we need to replace these characters. Was assuming it may cause issues with vpp-agent and/or vpp itself if the name contains `-` or `_`. Felt safer this way.

sdk/endpoint/composite/connection.go
edwarnicke: Ah... I misread... got it :)

PR #678, opened 2019-01-21
examples: add proxy-nsc

nickolaev:
The proxy-nsc is a sample HTTP proxy that creates a new
NS connection for each request it receives.
It can also turn HTTP headers to connection request labels.
Example: "NSM-App: Firewall" turns into "app=firewall"

The new examples is embedded in the vpn example deployment.


nickolaev:
This will go in the new "networkservicemesh/examples" repo.
Closing.

PR #686, opened 2019-01-22
Fix nsmd containers handlings of signals so they can terminate fast.

edwarnicke:


edwarnicke:
Closed in favor of #756

Issue #664, opened 2019-01-17
Integration testing for MonitorCrossConnect from nsmd


edwarnicke:
We have wonderful go based integration testing in test/integration

Would be good to extend it to check to make sure MonitorCrossConnect from nsmd is working as expected as well :)

nickolaev:
#689 merged

nickolaev:
@lobkovilya the tests faile randomly with: https://circleci.com/gh/networkservicemesh/networkservicemesh/11122

If the CI is rerun, the tests do pass.

Reopening.

matrohon:
This race may be due to the portforward error that happens before the seg fault, don't you think?

matrohon:
the same error is happenig here : 

https://circleci.com/gh/networkservicemesh/networkservicemesh/11122

matrohon:
Do you have any other occurrence of this failure?, I can't find some in CircleCI.

nickolaev:
Ups, it did it again: https://circleci.com/gh/networkservicemesh/networkservicemesh/11314

I don't think it is normal to see:
```shell
panic: runtime error: invalid memory address or nil pointer dereference
```

Whatever happens, this nil pointer should be handled more gracefully at least.
@lobkovilya what do you think?

edwarnicke:
Done.

PR #689, opened 2019-01-22
Issue #664: Integration tests for CrossConnectMonitor

lobkovilya:


nickolaev:
@lobkovilya Aparently the tests fail randomly:
https://circleci.com/gh/networkservicemesh/networkservicemesh/11122

This actually passes if the test is re-run. Still we have to fix it.

PR #696, opened 2019-01-23
WIP: Google cloud builder make machinery

edwarnicke:


.k8s.mk
```diff
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+BUILD_CONTAINERS=nsc nsmd nsmdp nsmd-k8s icmp-responder-nse vppagent-dataplane devenv crossconnect-monitor
```
nickolaev: Are we sure these are all we build? We lack at least `vppagent-icmp-responder-nse`. And then maybe the firewall.

PR #699, opened 2019-01-25
Create generic code for different types of monitors

lobkovilya:
Create interface `MonitorServer` and implementation `MonitorServerImpl`:
```
type MonitorServer interface {
	Update(entity Entity)
	Delete(entity Entity)

	AddRecipient(recipient Recipient)
	DeleteRecipient(recipient Recipient)
	MonitorEntities(stream grpc.ServerStream) error

	Serve()
}
```

It provides entities monitoring logic. Every objects that implements method `GetId() string` can be considered as `Entity`. That means that all grpc generated entities (`CrossConnect`, `local.Connection`, `remote.Connection`) could easily be monitored by `Recipient`:

```
type Recipient interface {
	SendMsg(msg interface{}) error
}
```

For example every `grpc.ServerStream` can be treat as `Recipient`.

`MonitorServer` accepts Entity and before sending it to `Recipient` wraps it into `Event`:

```
type Event struct {
	EventType string
	Entities  map[string]Entity
}
```

And additionally for each `Entity` we have to implement `EventConverter` interface:
```
type EventConverter interface {
	Convert(event Event) (interface{}, error)
}
```
This is needed for properly convert `Event` to `interface{}` that actually is some real grpc event type (crossconnect.CrossConnectEvent, connection.ConnectionEvent, etc)


edwarnicke:
@lobkovilya Thank you for the excellent writeup, this made it much simpler to review this PR.

PR #710, opened 2019-01-29
Fix Monitor server test stability

haiodo:
Fix Monitor server test stability


haiodo:
@edwarnicke @nickolaev  it seems this will fix master, and it looks a bit better to not rollback @lobkovilya  changes, but just have this test fixed :-)

Issue #697, opened 2019-01-24
Implement AutoHeal functionality


haiodo:
We need to be sure all AutoHeal functionality is working.

- [x] Local NSE die
All good, we could easy recover.
- [x] Remote NSE die
All good, we close local Dataplane and let source NSMD to choose another NSE.

- [x] Local Dataplane die
All good, we could re-program Dataplanes, remote and local one.
- [x] Remote Dataplane die
All good, we let source NSMD to recover and call our Dataplane to heal.

- [x] Remote NSM die
- - [x] Track remote NSMD down
- - [x] NSE should re-connect to new NSMD.
- - [x] Dataplane should re-connect to new NSMD.
- - [x] NSMD should search for new NSE and connect to.

- [x] Local NSMD die
- - [x] If Dataplane is alive, find local NSE and restore from Dataplane active connections.
- - [ ] Both NSMD and Dataplane are die, Client is alive. Client Monitor enforces Request to heal connection.
- - [x] Remote NSMD should track source NSMD die, wait for timeout and if heal not received close NSE connection.

Workflow for implemented parts:

Startup process:
<img width="607" alt="NSM-Startup-Recovery" src="https://user-images.githubusercontent.com/477235/54074888-d7088e80-42ca-11e9-91a7-0d50fd9d8aef.png">

Healing process:
<img width="699" alt="NSM-Healing" src="https://user-images.githubusercontent.com/477235/54074892-da037f00-42ca-11e9-97ba-0fe8d280f7cd.png">

Request/Heal handling
<img width="1182" alt="NSM-Request" src="https://user-images.githubusercontent.com/477235/54074893-dcfe6f80-42ca-11e9-95a2-1dc928d048f5.png">

haiodo:
Dataplane auto heal is a bit more complex, found 3 cases at the moment:

<img width="1100" alt="auto-heal dataplane" src="https://user-images.githubusercontent.com/477235/52044182-17f6e000-2574-11e9-89ae-7d6c301ee39f.png">

haiodo:
Initial diagram of how it working inside.
<img width="1754" alt="explained" src="https://user-images.githubusercontent.com/477235/53025794-31fd5180-3494-11e9-9ac3-e04c5bb286a1.png">

haiodo:
Short update:
* Working on remote NSM die Healing.
* NSM Die from Dataplane recovery healing, required to not have NSE resource leaking.

haiodo:
[...]

PR #722, opened 2019-02-04
NSE/Dataplane auto healing

haiodo:
Implementation of Auto heal functionality for NSE local/remote die and Dataplane local/remote die.

## Description
Auto heal functionality allow to recover NSC <-> NSE operability via requesting new NSE if previous one is lost, and recover from situations if Dataplane pod is die because of implementation errors/development instabilities.

## Motivation and Context
Fixes for issue: #697

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested
- [x] NSE die tested unit
- [x] NSE die tested integration
- [x] Dataplane die tested unit
-  [x] Dataplane die tested integration

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

### DIFF
--- a/controlplane/pkg/model/model.go
+++ b/controlplane/pkg/model/model.go
@@ -76,7 +76,7 @@ type Model interface {
 	GetClientConnection(connectionId string) *ClientConnection
 	GetAllClientConnections() []*ClientConnection
 	UpdateClientConnection(clientConnection *ClientConnection)
-	DeleteClientConnection(connectionId string) *ClientConnection
+	DeleteClientConnection(connectionId string)
 
 	ConnectionId() string
 
@@ -141,12 +141,11 @@ func (i *impl) UpdateClientConnection(clientConnection *ClientConnection) {
 	}
 }
 
-func (i *impl) DeleteClientConnection(connectionId string) *ClientConnection {
+func (i *impl) DeleteClientConnection(connectionId string) {
 	i.Lock()
 	clientConnection := i.clientConnections[connectionId]
 	if clientConnection == nil {
 		i.Unlock()
-		return nil
 	}
 	delete(i.clientConnections, connectionId)
 	i.Unlock()
@@ -154,7 +153,6 @@ func (i *impl) DeleteClientConnection(connectionId string) *ClientConnection {
 	for _, listener := range i.listeners {
 		listener.ClientConnectionDeleted(clientConnection)
 	}
-	return clientConnection
 }
 
 func (i *impl) AddListener(listener ModelListener) {
--- a/controlplane/pkg/model/model.go
+++ b/controlplane/pkg/model/model.go
@@ -141,7 +141,7 @@ func (i *impl) UpdateClientConnection(clientConnection *ClientConnection) {
 	}
 }
 
-func (i *impl) DeleteClientConnection(connectionId string) {
+func (i *impl) DeleteClientConnection(connectionId string) *ClientConnection {
 	i.Lock()
 	clientConnection := i.clientConnections[connectionId]
 	if clientConnection == nil {
@@ -153,6 +153,7 @@ func (i *impl) DeleteClientConnection(connectionId string) {
 	for _, listener := range i.listeners {
 		listener.ClientConnectionDeleted(clientConnection)
 	}
+	return clientConnection
 }
 
 func (i *impl) AddListener(listener ModelListener) {
--- a/controlplane/pkg/model/model.go
+++ b/controlplane/pkg/model/model.go
@@ -2,6 +2,7 @@ package model
 
 import (
 	"fmt"
+	"github.com/networkservicemesh/networkservicemesh/controlplane/pkg/apis/nsm"
 	"strconv"
 	"sync"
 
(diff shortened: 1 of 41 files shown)


edwarnicke:
@haiodo Could we rebase this on master to make sure the VPN CI runs against it?

haiodo:
Yes, will do. I'm still working on Dataplane healing tests, and Fighting dataplane issues :-(

nickolaev:
@haiodo 30 mins to pass the tests is too much and I'm not sure these are really passing and not just silently timeout. Please revert 549b932 and 030ca6a. Let's leave these to a separate PR.

controlplane/pkg/model/model.go
```diff
@@ -21,14 +22,45 @@ type Dataplane struct {
...
+}
+
+func (cc *ClientConnection) GetNetworkService() string {
+	if cc == nil {
+		return ""
+	}
+	return cc.Endpoint.GetNetworkService().GetName()
+}
+
+func (cc *ClientConnection) GetSourceConnection() nsm.NSMConnection {
```
nickolaev: Should the working be swapped here: `GetConnectionSource`?

controlplane/pkg/model/model.go
```diff
@@ -146,10 +178,14 @@ func (i *impl) DeleteClientConnection(connectionId string) {
 	clientConnection := i.clientConnections[connectionId]
 	if clientConnection == nil {
 		i.Unlock()
```
nickolaev: Can we just `defer i.Unlock()` while we're changing here?

controlplane/pkg/nsm/heal_constants.go
```diff
@@ -0,0 +1,8 @@
+package nsm
+
+import "time"
+
+const (
+	HealTimeout          = time.Minute * 10
```
nickolaev: What are these numbers? Please add some comments here and explain. 10 minutes look too much to me.

controlplane/pkg/nsm/nsm.go
```diff
@@ -46,32 +48,86 @@ func NewNetworkServiceManager(model model.Model, serviceRegistry serviceregistry
...
+func (srv *networkServiceManager) Request(ctx context.Context, request nsm.NSMRequest) (nsm.NSMConnection, error) {
+	// Check if we are recovering connection, by checking passed connection Id is known to us.
+	return srv.request(ctx, request, srv.model.GetClientConnection(request.GetConnectionId()))
+}
+
+func create_logid() (uuid string) {
+	b := make([]byte, 4)
+	_, err := rand.Read(b)
+	if err != nil {
+		fmt.Println("Error: ", err)
```
nickolaev: Why `fmt.Println` here?

controlplane/pkg/nsm/nsm.go
```diff
@@ -46,32 +48,86 @@ func NewNetworkServiceManager(model model.Model, serviceRegistry serviceregistry
...
 	}
 
-	var nsmConnection nsm.NSMConnection = nil
-	nsmConnection = srv.cloneConnection(request, nsmConnection)
+	// 1. Create a new connection object.
+	nsmConnection := srv.newConnection(request)
 
-	// Create a ConnectId for the request.GetConnection(), since connections are managed per NSM
-	nsmConnection.SetId(srv.createConnectionId())
+	// 2. Set connection id for new connections.
```
nickolaev: All these comments would make a great `docs/specs/AutoHealing.md`. Please consider it.

test/integration/nse_dataplane_heal_test.go
```diff
@@ -0,0 +1,94 @@
...
+	defer k8s.Cleanup()
+
+	Expect(err).To(BeNil())
+
+	s1 := time.Now()
+	k8s.Prepare("nsmd", "nsc", "nsmd-dataplane", "icmp-responder-nse", "jaeger")
+	logrus.Printf("Cleanup done: %v", time.Since(s1))
+
+	// Deploy open tracing to see what happening.
+	jaeger := k8s.CreatePod(pods.Jaeger())
```
nickolaev: We do not want to test `Jaeger` deployments in our CI. If you need these during the development time, please use env variables toe enable them.

controlplane/pkg/nsm/heal_constants.go
edwarnicke: I'm also curious about why 10 minutes...

controlplane/pkg/model/model.go
haiodo: Yes, thanks.

controlplane/pkg/nsm/heal_constants.go
haiodo: Let's have it 1 minute for now, in case some long pod-restarts.

controlplane/pkg/nsm/nsm.go
haiodo: Yes, planed to work on this as next task.

test/integration/nse_dataplane_heal_test.go
haiodo: Since we have this inside main NSMD codebase, for using Jarger, I think we need to be sure it all working with/without it. So we should have at least some small test with it. I've planed to access API to it, to retrieve a full list of operations performed during Connection establish/heal etc, but not yet able to do so. So lets pod's definition in tests to be here, and use it in some other PR.

haiodo:
> * What are the chances that we are just hiding known dataplane problems like this? Specifically during CI.

I think we just should have more tests, to check dataplane is working fine in different scenarios, also @stepan-anokhin is working for capturing VPP dumps, so should be fine I think.

> * Should we have a disable knob? We can use it in CI and disable autohealing to catch problems during the tests that to not verify the autoheal functionality.

Adding property to not HEAL I think would be a good idea for debugging, testing of our components. Let's me add this as separate PR.

> * Docs. Your comments are a great base for putting these in docs/spec/.

Planed to do as separate PR.

> * Please split the test refactoring in a separate PR and base this one on it.

Very hard, since new tests are based on this refactoring.

controlplane/pkg/model/model.go
haiodo: Nope, defer is not working, since firstly it lock for writing and for listeners lock for reading.

haiodo:
@edwarnicke @fkautz waiting for your approval for merge this changes. Thanks.

PR #726, opened 2019-02-06
Handle Client connection failure

nickolaev:
The SDK client does have wrong behaviour after 10 iterations of retries to connect.

## Description
This patch adds a proper error handling in SDK/client for the case where all iteration retries failed.

## Motivation and Context
A wrong behaviour and crash were observed while the client runs in a volatile environment.
NSC for example just continues and finished its operation even though the Connect failed.

## How Has This Been Tested?
- [X] Tested locally
- [ ] Have not tested

## Types of changes
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

### DIFF
--- a/examples/cmd/nsc/nsc.go
+++ b/examples/cmd/nsc/nsc.go
@@ -32,7 +32,8 @@ func main() {
 		logrus.Fatalf("Unable to create the NSM client %v", err)
 	}
 
-	client.Connect("nsm", "kernel", "Primary interface")
+	if _, err := client.Connect("nsm", "kernel", "Primary interface"); err != nil {
+		logrus.Fatalf("Client connect failed with error: %v", err)
+	}
 	logrus.Info("nsm client: initialization is completed successfully")
-
 }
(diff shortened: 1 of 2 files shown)


examples/cmd/nsc/nsc.go
```diff
@@ -32,7 +32,8 @@ func main() {
 		logrus.Fatalf("Unable to create the NSM client %v", err)
 	}
 
-	client.Connect("nsm", "kernel", "Primary interface")
+	if _, err := client.Connect("nsm", "kernel", "Primary interface"); err != nil {
+		logrus.Fatalf("Client connect failed with error: %v", err)
```
haiodo: @nickolaev should you not print "nsm client: initialization is completed successfully" in case of error?

nickolaev: @haiodo. I believe Fatal exits the proces with 1.

nickolaev:
I can confirm, after manually running the NSC without an ICMP responder. It true 10 times, then dies. K8s restarts the Pod and so on until the service is alive and the connection is established.

Issue #711, opened 2019-01-30
Spec: Readiness probes for nsmd, vppagent-dataplane


edwarnicke:

nickolaev:
#730 merged. Closing.

PR #730, opened 2019-02-07
Add liveness/readiness probes for nsmd and vppagent-dataplane

rdimitrov:
## Description

Add liveness/readiness probes for nsmd and vppagent-dataplane. The changes consist of:
* Created `/liveness` and `/readiness` HTTP handlers for both components
* Each component sets specific flags stating if its dependencies initialized okay or not
* Updated the `kube-testing` description files to take into account these changes
* Updated the Kubernetes configuration files to enable liveness/readiness probing

## Motivation and Context

Related to #711

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

- [X] My change requires a change to the documentation.

### DIFF
--- a/k8s/conf/vppagent-dataplane.yaml
+++ b/k8s/conf/vppagent-dataplane.yaml
@@ -23,6 +23,18 @@ spec:
             - name: workspace
               mountPath: /var/lib/networkservicemesh/
               mountPropagation: Bidirectional
+          livenessProbe:
+            httpGet:
+              path: /liveness
+              port: 5555
+            initialDelaySeconds: 3
+            periodSeconds: 3
+          readinessProbe:
+            httpGet:
+              path: /readiness
+              port: 5555
+            initialDelaySeconds: 5
+            periodSeconds: 3
       volumes:
         - hostPath:
             path: /var/lib/networkservicemesh
--- a/k8s/conf/vppagent-dataplane.yaml
+++ b/k8s/conf/vppagent-dataplane.yaml
@@ -27,14 +27,16 @@ spec:
             httpGet:
               path: /liveness
               port: 5555
-            initialDelaySeconds: 3
-            periodSeconds: 3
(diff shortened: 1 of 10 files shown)


nickolaev:
@rdimitrov I've added a couple of checks in the PR description. It is somehow covered by the existing tests, it;s a new feature and we'll need to add something to the docs to reflect that this check is already something.

k8s/conf/vppagent-dataplane.yaml
```diff
@@ -23,6 +23,18 @@ spec:
...
+            httpGet:
+              path: /liveness
+              port: 5555
+            initialDelaySeconds: 3
+            periodSeconds: 3
+          readinessProbe:
+            httpGet:
+              path: /readiness
+              port: 5555
+            initialDelaySeconds: 5
```
edwarnicke: Question... what motivated this choice of initial delay?

edwarnicke:
@rdimitrov Overall approach is nicer than what I had suggested, good job!

I am curious about the choice of initialDelay values :)

rdimitrov:
@edwarnicke Thank you, Ed! I really appreciate it! 👍 

And related to the values, I set them more or less as a starting point. If it would've failed, I was going to increase it a bit. 
I can do some basic measurement to get a rough idea of what does it take for each component to start. And based on that, experiment a bit to see if it makes sense to tune them better.

rdimitrov:
Updates:
* Created a dedicated file in the `docs/spec/` folder for documenting that feature
* Added measuring of the boot time for both components (printed in logging=debug). Useful for debugging purposes, for example, if a liveness probe fails due to slow starting time.
* Adjusted the timings, so the system is more stable. For example, the default value of `timeoutSeconds` is 1s was not enough as it was causing the liveness probe to trigger a fail and restart the container.

PR #732, opened 2019-02-08
Adding VPN integration tests

nickolaev:
## Description
Add the `secure-intranet-connectivity` integration test.
This also adds testing capabilities to describe NetworkService enabling more complex real-world scenarios.

## Motivation and Context
This test exercises NSM's capabilities to implement NetwrokService CRD, link a couple of Endpoints and create complex deployments. It is setting a direction of increasing the number of integration tests and their coverage.

Currently only the simplest single host scenario is enabled. The remote test can be run  by defining `VPN_REMOTE_ENABLED`.

An option to run is:
VPN_REMOTE_ENABLED=1 make k8s-integration-TestVPN*-test

## How Has This Been Tested?

- [X] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -296,6 +296,21 @@ jobs:
             export TAG="$BUILD_TAG"
             make docker-vppagent-dataplane-build
             make docker-vppagent-dataplane-push
+  build-vppagent-firewall-nse:
+    working_directory: /go/src/github.com/networkservicemesh/networkservicemesh/
+    docker:
+      - image: circleci/golang:1.11
+    steps:
+      - checkout
+      - setup_remote_docker:
+          docker_layer_caching: true
+      - run:
+          command: |
+            export COMMIT="${CIRCLE_SHA1:8:8}"
+            export BUILD_TAG="circle-${CIRCLE_BUILD_NUM}"
+            export TAG="$BUILD_TAG"
+            make docker-vppagent-firewall-nse-build
+            make docker-vppagent-firewall-nse-push
 workflows:
   version: 2
   build-and-test:
@@ -328,6 +343,9 @@ workflows:
       - build-vppagent-dataplane:
           requires:
             - sanity-check
+      - build-vppagent-firewall-nse:
+          requires:
+            - sanity-check
       - packet-deploy:
           requires:
             - sanity-check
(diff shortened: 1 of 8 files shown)


haiodo:
Why revert is required?

nickolaev:
@haiodo I have added #733 to track this. Can you please check if you can reproduce it.
I agree that a full revert is not the solution. The problem is that the patch is huge, os its hard to isolate a small piece that causes it.

test/kube_testing/pods/vpn-gateway-nsc.go
```diff
@@ -0,0 +1,52 @@
...
+
+	pod := &v1.Pod{
+		ObjectMeta: v12.ObjectMeta{
+			Name: name,
+		},
+		TypeMeta: v12.TypeMeta{
+			Kind: "Deployment",
+		},
+		Spec: v1.PodSpec{
+			Containers: []v1.Container{
```
haiodo: You need to add NSC as init container, since it will do infinit restart in this case.

Following is example:
```
Spec: v1.PodSpec{
			RestartPolicy: v1.RestartPolicyNever,
			Containers: []v1.Container{
				{
					Name:            "alpine-img",
					Image:           "alpine:latest",
					ImagePullPolicy: v1.PullIfNotPresent,
					Command: []string{
						"tail", "-f", "/dev/null",
					},
				},
			},
			InitContainers: []v1.Container{
				{
					Name:            "nsc",
					Image:           "networkservicemesh/nsc:latest",
					ImagePullPolicy: v1.PullIfNotPresent,
					Resources: v1.ResourceRequirements{
						Limits: v1.ResourceList{
							"nsm.ligato.io/socket": resource.NewQuantity(1, resource.DecimalSI).DeepCopy(),
						},
						Requests: nil,
					},
					Env: envVars,
				},
			},
		},
```

nickolaev: Right ;) Thanks will fix it.

haiodo:
:thumbsup:

.circleci/config.yml
```diff
@@ -115,6 +115,7 @@ jobs:
             ./scripts/prepare-circle-integration-tests.sh
             mkdir -p ~/junit/
             gotestsum --junitfile ~/junit/integration-tests.xml ./test/...
+          no_output_timeout: 30m
```
nickolaev: I didn't get the impression that this is working as expected.

PR #723, opened 2019-02-05
Admission Webhook

platoff:
## Description
Admission Webhook implementation according to spec #708 

## Motivation and Context
See #708 

## How Has This Been Tested?
- [ ] Covered by existing integration testing
- [ ] Added integration testing to cover
- [x] Tested locally
- [ ] Have not tested

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #735, opened 2019-02-10
WIP: Admission Webhook CI

platoff:
## Description

Add admission webhook to CI, implement client supporting admission webhook annotation

## Motivation and Context

This is a continuation of admission webhook work #723 

## How Has This Been Tested?
- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/sdk/client/clientv2.go
+++ b/sdk/client/clientv2.go
@@ -1,0 +1,100 @@
+// Copyright 2018, 2019 VMware, Inc.
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at:
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package client
+
+import (
+	"context"
+	"github.com/networkservicemesh/networkservicemesh/controlplane/pkg/apis/local/connection"
+	"github.com/networkservicemesh/networkservicemesh/pkg/tools"
+	"github.com/sirupsen/logrus"
+	"os"
+	"strings"
+
+	"github.com/networkservicemesh/networkservicemesh/sdk/common"
+)
+
+const (
+	annotationEnv = "NS_NETWORKSERVICEMESH_IO"
+)
+
+type NsmClientV2 struct {
+	clients     []*NsmClient
+	connections []*connection.Connection
+}
+
+func configFromUrl(configuration *common.NSConfiguration, url *tools.NsUrl) *common.NSConfiguration {
+	var conf common.NSConfiguration
+	if configuration != nil {
+		conf = *configuration
+	}
+	conf.OutgoingNscName = url.NsName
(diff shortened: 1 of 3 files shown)


nickolaev:
@platoff I'm not sure we need to have a `v2` client. Why not merge all functionality in the same code/client and allow the users to select it by supplying different env variables. E.g. if `NS_NETWORKSERVICEMESH_IO` is present config from it, otherwise use the previous behaviour.

platoff:
@nickolaev I'd like to avoid a `v2` client, but need your suggestion: as [mutation admission controller spec](https://docs.google.com/document/d/19VzbWQj3Pgf_VSUIQUXCS5HXV-chNpuXNKjVziSWjio/edit#) says we have comma delimited list of ‘urls’ for the network service, for example: 

```ns.networkservicemesh.io: icmp-responder, secure-intranet-connectivity/eth2?app=firewall&version=2```

I'm assuming this example would connect client to 2 network services `icmp-responder` and `secure-intranet-connectivity`. If I'm correct on this, current client does not support connecting to many services, so I introduced `v2` client which incapsulate N current clients to connect to N services with minimal impact to the codebase.

Any suggestions are much appreciated. Thank you!

**Comment by nickolaev** [comment]
@platoff OK I see what you mean now. So essentially this `v2` is just an overlay providing a `collection` of `NSClient`s. Then maybe we can just keep your proposed code and just rename it to some sort of `NSClientMultiple` or `NSClientCollection` or some suitable name.
@edwarnicke @fkautz any opinions here?

File: sdk/client/clientv2.go [hunk 1]
```diff
@@ -0,0 +1,100 @@
...
+		}
+		labels.WriteString(k)
+		labels.WriteRune('=')
+		labels.WriteString(v[0])
+	}
+	conf.OutgoingNscLabels = labels.String()
+	return &conf
+}
+
+func NewNSMClientV2(ctx context.Context, configuration *common.NSConfiguration) (*NsmClientV2, error) {
```
nickolaev: I don't see this as a second version fo the client. It is a higher abstraction that builds on top of the client. My preference is that this one is renamed to `NewNSMClientList` or similar.

k8s/cmd/admission-webhook/main.go
```diff
@@ -136,20 +136,44 @@ func (whsvr *WebhookServer) mutate(ar *v1beta1.AdmissionReview) *v1beta1.Admissi
 	logrus.Infof("AdmissionReview for Kind=%v, Namespace=%v Name=%v UID=%v patchOperation=%v UserInfo=%v",
 		req.Kind, req.Namespace, req.Name, req.UID, req.Operation, req.UserInfo)
 
-	var deployment appsv1.Deployment
-	if err := json.Unmarshal(req.Object.Raw, &deployment); err != nil {
-		logrus.Errorf("Could not unmarshal raw object: %v", err)
+	var meta *metav1.ObjectMeta
```
nickolaev: I would like to see this file changed in a separate PR with a better explanation of what we do here.
This PR is about adding to the CI and then also chnages the core component.
We were already burned by merging huge patches with too many changes incorporated. Let's try to be more strict.

edwarnicke:
@nickolaev On the issue of the v2 client... how would you feel about augmenting the existing client to simply support the new needed multiplicity and the new env variable pattern?

@platoff tends to be a super courteous coder, and do things like v2 instead of messing with other folks code, which is sometimes awesome.  But in this case, does it make more sense to get back to *one* client (an updated and augmented v1 client)?

nickolaev:
@edwarnicke @platoff I am all about having one client.

Issue #725, opened 2019-02-05
TestNSCAndICMPRemote is failed because of VPP dataplane error


haiodo:
=== RUN   TestNSCAndICMPRemote
time="2019-02-05T04:25:35Z" level=info msg="Cleanup done: 303.362144ms"
time="2019-02-05T04:25:35Z" level=info msg="Checking node: ci-12029-master"
time="2019-02-05T04:25:35Z" level=info msg="Checking node: ci-12029-worker1"
time="2019-02-05T04:25:39Z" level=info msg="Started NSMD/Dataplane: 3.895254482s on node 0"
time="2019-02-05T04:25:43Z" level=info msg="Started NSMD/Dataplane: 8.295653286s on node 1"
time="2019-02-05T04:25:44Z" level=info msg="Starting ICMP Responder NSE on node: 1"
time="2019-02-05T04:25:55Z" level=info msg="ICMP Responder started done: 11.404754838s"
time="2019-02-05T04:25:59Z" level=info msg="NSC started done: 3.998489083s"
time="2019-02-05T04:26:00Z" level=info msg="NSC IP status Ok"
time="2019-02-05T04:26:01Z" level=info msg="NSC Route status, Ok"
time="2019-02-05T04:26:16Z" level=info msg="NSC Ping is success:PING 10.20.1.2 (10.20.1.2): 56 data bytes\n\n--- 10.20.1.2 ping statistics ---\n5 packets transmitted, 0 packets received, 100% packet loss\n"

Network is not configured, it seems it still sometimes could not find a proper network interface, not sure why.

Full logs for test failure.
[full_logs.txt](https://github.com/networkservicemesh/networkservicemesh/files/2832640/full_logs.txt)

Dataplane last log lines:
```
[...]

PR #748, opened 2019-02-15
Bug: Add check of cmdline during resolving namespace by inode

lobkovilya:

Dataplane chose the first netns by inode that it finds in /proc directory. But sometimes it can be init-container's netns, so vppagent can't resolve it later. Now it checks that netns located in pause-container process directory.

## Motivation and Context
[Issue #725](https://github.com/networkservicemesh/networkservicemesh/issues/725)

## How Has This Been Tested?
- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/utils/fs/fsutils.go
+++ b/utils/fs/fsutils.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"strings"
 	"syscall"
 	"unicode"
 )
@@ -52,7 +53,9 @@ func FindFileInProc(inode uint64, suffix string) (string, error) {
 				continue
 			}
 			if tryInode == inode {
-				return filename, nil
+				if cmdline, err := GetCmdline(name); err == nil && strings.Contains(cmdline, "pause") {
+					return filename, nil
+				}
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


utils/fs/fsutils.go
```diff
@@ -79,3 +82,11 @@ func GetAllNetNs() ([]uint64, error) {
 	}
 	return inodes, nil
 }
+
+func GetCmdline(pid string) (string, error) {
```
nickolaev: This looks like a common code. Should we move it to utils, tools or similar?

utils/fs/fsutils.go
```diff
@@ -52,7 +53,9 @@ func FindFileInProc(inode uint64, suffix string) (string, error) {
 				continue
 			}
 			if tryInode == inode {
-				return filename, nil
+				if cmdline, err := GetCmdline(name); err == nil && strings.Contains(cmdline, "pause") {
```
nickolaev: This changes the behaviour of the function. It is no longer a simple FindFileInProc. It checks the cmd line for a magic "pause" string. I don't think this is a reliable way to solve the issue
For example What will happen if we have a pod with 2 apps + 1 init container?
Do we have other options?

lobkovilya: According to kubernetes documentation pod always starts `pause` container with pid=1. Pause container has same lifetime as pod that it works on. So it doesn't matter how many apps and init containers we have in our pod. 
But yes, you are right, I have to choose another name for that function :)

nickolaev:
`TestVPNLocal` fails without connection to this patch. Merging.

PR #761, opened 2019-02-20
Breakup integration testing

edwarnicke:
## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


haiodo:
@edwarnicke I think it would be too complicated to add new tests, since we will need to tweak build environment every time. We should probable split them using different packages.

edwarnicke:
@haiodo If this works, I think you are likely correct, for the moment, just seeing if this solves the problem :)

PR #763, opened 2019-02-20
Admission web hook will check if the pod is already processed

nickolaev:
## Description
This patch adds additional checks if the webhook should be applied. Specifically, it evaluates if there is already an "NSC" initCOntainer injected.

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


k8s/cmd/admission-webhook/main.go
```diff
@@ -76,7 +79,15 @@ func init() {
 	_ = v1.AddToScheme(runtimeScheme)
 }
 
-func getAnnotationValue(ignoredList []string, metadata *metav1.ObjectMeta) (string, bool) {
+func getAnnotationValue(ignoredList []string, metadata *metav1.ObjectMeta, spec *corev1.PodSpec) (string, bool) {
+
+	// check if InitContainer already injected
+	for _, c := range spec.InitContainers {
+		if c.Name == "nsc" {
```
nickolaev: Having "nsc" hardcoded as the name of the annotation container is not great.

nickolaev:
@platoff what do you think here?

PR #765, opened 2019-02-21
Enforce build to fail if code is not go fmt compliant

rdimitrov:
## Description

The changes consist of adding another check in the Circle CI configuration file to run a `go fmt` check and fail the build if it finds something that is not compliant.

The first commit is with the changes related to enabling that kind of check in the CI configuration and the second applies the result from running a `go fmt ./... or make format` locally.

**Update:**
Thought that it would be useful to add an additional `golint` run(not breaking the build if errors are found though).

## Motivation and Context

Using `go fmt` makes the code easier to read and write and overall it helps to keep it in better shape.

Also most of the IDEs already have `go fmt` plugins that run upon saving which forces having the content of the file formatted.
## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,6 +13,11 @@ jobs:
       - run: yamllint -c .yamllint.yml $(git ls-files '*.yaml' '*.yml')
       - run: go version
       - run: "! go fmt ./... 2>&1 | read"
+      - run:
+          name: golint
+          command: |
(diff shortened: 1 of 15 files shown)


nickolaev:
@rdimitrov please merge master to run the tests

edwarnicke:
Thanks for this, its a very very good idea :)

PR #770, opened 2019-02-22
Specify namespace for NSM

### LINKED ISSUE #754: Do not assume "default" namespace
At several places in NSM code we assume that we work in the "default" Kubernetes Namespace.

https://github.com/networkservicemesh/networkservicemesh/blob/master/k8s/pkg/registryserver/nse.go

https://github.com/networkservicemesh/networkservicemesh/blob/master/k8s/cmd/crossconnect-monitor/crossconnect_monitor.go#L96

"default" is also assumed in the integration testing.

Fix this and provide a way to specify the namespace that NSM operates in.

ivanayov:
Until now `default` namespace was used for all NSM resources.
Migrating to own `nsm-system` namespace as a first step to a possible multi-namespace update in future

Signed-off-by: Ivana Yovcheva <iyovcheva@vmware.com>

## Motivation and Context

Closes #754 

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

Tested locally.

Integration tests pass 1 by one, but fail together.

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -112,6 +112,9 @@ jobs:
       gke_project_id:
         type: string
         default: "ci-management"
+      namespace:
+        type: string
+        default: "default"
     working_directory: /go/src/github.com/networkservicemesh/networkservicemesh
     docker:
       - image: circleci/golang
@@ -190,6 +193,7 @@ jobs:
       CLOUD_PROVIDER: "<< parameters.cloud_provider >>"
       KUBECONFIG: /home/circleci/project/data/kubeconfig
       GO111MODULE: "on"
+      NSM_NAMESPACE: "<< parameters.namespace >>"
 
   packet-destroy:
     docker:
@@ -446,6 +450,7 @@ workflows:
             - packet-deploy
       - integration-tests:
           name: "packet-test-basic-usecase"
+          namespace: "nsm-system-integration"
           cloud_provider: "packet"
           cluster_id: "1"
           test_tags: "basic usecase"
@@ -453,6 +458,7 @@ workflows:
             - packet-deploy-k8s-1
       - integration-tests:
           name: "packet-test-recover"
+          namespace: "nsm-system-integration"
           cloud_provider: "packet"
           cluster_id: "2"
           test_tags: "recover"
@@ -460,6 +466,7 @@ workflows:
             - packet-deploy-k8s-2
       - integration-tests:
           name: "packet-test-bench"
+          namespace: "nsm-system-integration"
           cloud_provider: "packet"
           cluster_id: "2"
           test_tags: "bench"
--- a/.mk/k8s.mk
+++ b/.mk/k8s.mk
@@ -28,7 +28,10 @@ DEPLOYS = $(DEPLOY_INFRA) $(DEPLOY_ICMP) $(DEPLOY_VPN)
 
 CLUSTER_CONFIG_ROLE = cluster-role-admin cluster-role-binding cluster-role-view
 CLUSTER_CONFIG_CRD = crd-networkservices crd-networkserviceendpoints crd-networkservicemanagers
-CLUSTER_CONFIGS = $(CLUSTER_CONFIG_ROLE) $(CLUSTER_CONFIG_CRD)
+CLUSTER_CONFIG_NAMESPACE = namespace-nsm
+CLUSTER_CONFIGS = $(CLUSTER_CONFIG_ROLE) $(CLUSTER_CONFIG_CRD) $(CLUSTER_CONFIG_NAMESPACE)
+
+NSM_NAMESPACE = `cat "${K8S_CONF_DIR}/${CLUSTER_CONFIG_NAMESPACE}.yaml" | awk '/name:/ {print $$2}'`
 
 # All of the rules that use vagrant are intentionally written in such a way
 # That you could set the CLUSTER_RULES_PREFIX different and introduce
@@ -294,7 +297,7 @@ k8s-admission-webhook-load-images:  k8s-start $(addsuffix -load-images,$(addpref
 
 .PHONY: k8s-admission-webhook-create-cert
 k8s-admission-webhook-create-cert:
-	./scripts/webhook-create-cert.sh
+	@NSM_NAMESPACE=${NSM_NAMESPACE} ./scripts/webhook-create-cert.sh
 
 .PHONY: k8s-admission-webhook-delete-cert
 k8s-admission-webhook-delete-cert:
--- a/deployments/helm/icmp-responder/templates/icmp-responder-nse.tpl
+++ b/deployments/helm/icmp-responder/templates/icmp-responder-nse.tpl
@@ -45,4 +45,4 @@ spec:
               networkservicemesh.io/socket: 1
 metadata:
   name: icmp-responder-nse
-  namespace: default
+  namespace: nsm-system
--- a/k8s/cmd/crossconnect-monitor/crossconnect_monitor.go
+++ b/k8s/cmd/crossconnect-monitor/crossconnect_monitor.go
@@ -15,6 +15,7 @@ import (
 	"github.com/golang/protobuf/ptypes/empty"
 	"github.com/networkservicemesh/networkservicemesh/controlplane/pkg/apis/crossconnect"
 	"github.com/networkservicemesh/networkservicemesh/k8s/pkg/networkservice/clientset/versioned"
+	"github.com/networkservicemesh/networkservicemesh/k8s/pkg/networkservice/namespace"
 	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -92,8 +93,9 @@ func lookForNSMServers() {
 		logrus.Fatalln("Unable to initialize nsmd-k8s", err)
 	}
 
+	nsmNamespace := namespace.GetNamespace()
 	for !closing {
-		result, err := nsmClientSet.Networkservicemesh().NetworkServiceManagers("default").List(metav1.ListOptions{})
+		result, err := nsmClientSet.Networkservicemesh().NetworkServiceManagers(nsmNamespace).List(metav1.ListOptions{})
 		if err != nil {
 			logrus.Fatalln("Unable to find NSMs", err)
 		}
--- a/k8s/conf/skydive.yaml
+++ b/k8s/conf/skydive.yaml
@@ -4,6 +4,7 @@ metadata:
   name: skydive-analyzer
   labels:
     app: skydive-analyzer
+  namespace: nsm-system
 spec:
   type: NodePort
   ports:
@@ -24,6 +25,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: skydive-analyzer-config-file
+  namespace: nsm-system
 data:
   skydive.yml: |
     storage:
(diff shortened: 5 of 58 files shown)


edwarnicke:
Good start :)

test/kube_testing/k8s.go
```diff
@@ -301,29 +301,29 @@ func (o *K8s) GetNodes() []v1.Node {
 }
 
 func (o *K8s) ListPods() []v1.Pod {
-	podList, err := o.clientset.CoreV1().Pods("default").List(metaV1.ListOptions{})
```
haiodo: Could we use some configurable values here?,
I suppose we could try to run few tests in parallel in case different namespaces will be used for test pod deployment. Parallel would be perfect scenario, since deployment could be not fast.

ivanayov: Thanks @haiodo  This is a good proposal. I can do the multi-namespace approach in a follow up PR.

ivanayov:
For some reason, when I run failing tests separately, they pass (TestDataplaneCrossConnectReconnect, TestExec, TestDataplaneCrossConnectUpdate, TestDataplaneCrossConnectMultiple), but when I run all tests they fail. Only TestDataplaneCrossConnectBasic is always successful.

This looks like a concurrency issue here https://github.com/networkservicemesh/networkservicemesh/blob/master/test/kube_testing/k8s.go#L55

haiodo:
Hi @ivanayov https://github.com/networkservicemesh/networkservicemesh/blob/master/test/kube_testing/k8s.go#L55 looks ok, it is designed to deploy pods in parallel, since it is a bit faster, I think this method call source need to be checked, @stepan-anokhin As I remember it is your tests, could you please take a look. Thanks.

ivanayov:
I got the problem :) Cleanup is deleting the namespace and it takes time to terminate, so when running in parallel, some of the pods try to deploy on terminating namespace

ivanayov:
Requires #875

edwarnicke:
@ivanayov How are things going with this?

ivanayov:
@edwarnicke I'm back on this this week.

@nickolaev said you're having concerns about having unique namespace per test. We discussed that with the tests we'd like to cover possible actual scenarios and my opinion is that if we want to have different use cases operating in the same namespace, this should be a separate test case. Having separated cluster setup and cleanup for each test, but using the same resources concurrently might be problematic.

After rebase I'm having this error, so currently I'm trying to resolve it:
```
Pod nsmgr-kube-master container not started
....
Error starting nsmd service: Failed to get my own NetworkServiceManager: rpc error: code = Unknown desc = an empty namespace may not be set when a resource name is provided\"\npanic: runtime error: invalid memory address or nil pointer dereference
```

ivanayov:
Previous issues are fixed, but still fails, as roles are statically created from yaml and they are namespaced. In order to close close this, they should go to tests code.

ivanayov:
Requires #947 to be merged first

ivanayov:
All tests are passing locally, but in the CI they fail with `"failed to create a namespace (error: Post https://147.75.89.49:6443/api/v1/namespaces: stream error: stream ID 1; INTERNAL_ERROR)"`
Looks like a problem with the K8s server

.circleci/config.yml
```diff
@@ -156,6 +159,7 @@ jobs:
           command: |
             export COMMIT="${CIRCLE_SHA1:8:8}"
             export CONTAINER_TAG="${COMMIT}"
+            export NSM_NAMESPACE="<< parameters.namespace >>"
```
nickolaev: No need to set it locally if you alredy set it globally below. PLease remove.

.circleci/config.yml
```diff
@@ -190,6 +194,7 @@ jobs:
       CLOUD_PROVIDER: "<< parameters.cloud_provider >>"
       KUBECONFIG: /home/circleci/project/data/kubeconfig
       GO111MODULE: "on"
+      NSM_NAMESPACE: "<< parameters.namespace >>"
```
nickolaev: This is a global env for the whole job, all the steps get this. No need to set explicitly inside.

.circleci/config.yml
```diff
@@ -258,6 +263,9 @@ jobs:
     parameters:
       container:
         type: string
+      namespace:
```
nickolaev: Not used, please remove.

.circleci/config.yml
```diff
@@ -373,61 +381,73 @@ workflows:
       - license-scan
       - build-container:
           name: "build-nsmd"
+          namespace: "nsm-system-integration"
```
nickolaev: Nope. You don't need this set here for the build. You want it for the integration tests.

.circleci/config.yml
```diff
@@ -112,6 +112,9 @@ jobs:
       gke_project_id:
         type: string
         default: "ci-management"
+      namespace:
+        type: string
+        default: ""
```
nickolaev: I prefer if the default value is more explicit. I guess `default: "default"` is better here.

.mk/k8s.mk
```diff
@@ -28,7 +28,10 @@ DEPLOYS = $(DEPLOY_INFRA) $(DEPLOY_ICMP) $(DEPLOY_VPN)
 
 CLUSTER_CONFIG_ROLE = cluster-role-admin cluster-role-binding cluster-role-view
 CLUSTER_CONFIG_CRD = crd-networkservices crd-networkserviceendpoints crd-networkservicemanagers
-CLUSTER_CONFIGS = $(CLUSTER_CONFIG_ROLE) $(CLUSTER_CONFIG_CRD)
+CLUSTER_CONFIG_NAMESPACE = namespace-nsm
+CLUSTER_CONFIGS = $(CLUSTER_CONFIG_ROLE) $(CLUSTER_CONFIG_CRD) $(CLUSTER_CONFIG_NAMESPACE)
+
+NSM_NAMESPACE = "nsm-system"
```
nickolaev: Can you `grep`/`sed` this from $(CLUSTER_CONFIG_NAMESPACE) yaml file?

deployments/helm/icmp-responder/templates/icmp-responder-nse.tpl
```diff
@@ -45,4 +45,4 @@ spec:
               networkservicemesh.io/socket: 1
 metadata:
   name: icmp-responder-nse
-  namespace: default
+  namespace: nsm-system
```
nickolaev: I guess for the helm charts we should use parameters instead of hardcoding these. Let's take a note and do this in a follow up PR.

scripts/prepare-circle-integration-tests.sh
```diff
@@ -28,4 +28,6 @@ kubectl apply -f k8s/conf/crd-networkservices.yaml
 kubectl apply -f k8s/conf/crd-networkserviceendpoints.yaml
 kubectl apply -f k8s/conf/crd-networkservicemanagers.yaml
 
+# export NSM_NAMESPACE="nsm-system-integration"
```
nickolaev: If this is just a hint: prepend it with a proper comment when to enable this line.

k8s/cmd/crossconnect-monitor/crossconnect_monitor.go
```diff
@@ -26,6 +26,8 @@ import (
 var closing = false
 var managers = map[string]string{}
 
+const NsmNamespaceEnv = "NSM_NAMESPACE"
```
nickolaev: Can we have this constant somewhere in k8s/pkg and just refer to it here.
There are several other places where you use it so it makes sense to have it in a central place.
I will makre these for completenes.

k8s/pkg/registryserver/registry_cache.go
```diff
@@ -11,6 +13,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const NsmNamespaceEnv = "NSM_NAMESPACE"
```
nickolaev: Constant somewhere in k8s/pkg, please.

test/kube_testing/crds/networkservice.go
```diff
@@ -75,8 +75,9 @@ func NewNSCRD() (*NSCRD, error) {
 	config, err := clientcmd.BuildConfigFromFlags("", path)
 	Expect(err).To(BeNil())
 
+	nsmNamespace := os.Getenv("NSM_NAMESPACE")
```
nickolaev: Constant somewhere in k8s/pkg, please.

test/kube_testing/k8s.go
```diff
@@ -301,6 +303,13 @@ func (o *K8s) deletePodForce(pod *v1.Pod) error {
 	return nil
 }
 
+func (o *K8s) initNamespace() {
+	var err error
+	nsmNamespace := os.Getenv("NSM_NAMESPACE")
```
nickolaev: Constant somewhere in k8s/pkg, please.

test/kube_testing/pods/nsmd.go
```diff
@@ -138,6 +143,10 @@ func NSMgrPodWithConfig(name string, node *v1.Node, config *NSMgrPodConfig) *v1.
 							Name:  "NODE_NAME",
 							Value: nodeName,
 						},
+						v1.EnvVar{
+							Name:  "NSM_NAMESPACE",
```
nickolaev: Constant somewhere in k8s/pkg, please.

k8s/conf/skydive.yaml
```diff
@@ -54,7 +56,7 @@ data:
     ui:
       topology:
         favorites:
-          nsm-filter: "G.V().Has('Type', 'container', 'Docker.Labels.io.kubernetes.pod.namespace', 'default').In('Type', 'netns').Descendants().As('namespaces').G.V().Has('Type', 'host').As('hosts').Select('namespaces', 'hosts')"
+          nsm-filter: "G.V().Has('Type', 'container', 'Docker.Labels.io.kubernetes.pod.namespace', 'nsm-system').In('Type', 'netns').Descendants().As('namespaces').G.V().Has('Type', 'host').As('hosts').Select('namespaces', 'hosts')"
```
nickolaev: Yeah, that is not great, but we don't have a better way to parametrise this. I guess @matrohon can have an idea of how to be more namespace flexible here. That should be a follow-up PR, though.

test/kube_testing/k8s.go
nickolaev: So what if the env is empty or does not exist?

ivanayov: In this case `nsmNamespace` will be empty and `default` namespace will be used

test/kube_testing/k8s_forward.go.go
```diff
@@ -103,10 +104,14 @@ func (p *PortForward) getFreePort() (int, error) {
 
 // Create an httpstream.Dialer for use with portforward.New
 func (p *PortForward) dialer() (httpstream.Dialer, error) {
+	podname := ""
+	if p.pod != nil {
+		podname = p.pod.Name
+	}
```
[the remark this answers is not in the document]
ivanayov: This will happen only if `p.pod` is nil. May be it's better to return an informative error in that case

.circleci/config.yml
```diff
@@ -153,6 +156,7 @@ jobs:
           no_output_timeout: 40m
       - run:
           name: Running integration tests - <<parameters.test_tags>>
+          namespace: "nsm-system-integration"
```
nickolaev: i'm sorry I probably was not clear. This is not needed here. It is needed in the workflow (at the bottom of the ffile) where you invoke `integration-tests` for the different scenarios and cloud providers. As we agreed all `integration-tests` for packet will have "nsm-system-integration" while the rest will not be set and will default to "default".

k8s/pkg/networkservice/clientset/versioned/constants.go
```diff
@@ -0,0 +1,18 @@
+// Copyright 2018 VMware, Inc.
```
nickolaev: 2019

nickolaev: The `clientset` is auto-generated code. maybe better to move it somewhere else. Why not create a new package under `k8s/pks/namespace` and have only this file for a start. I'm sure there will be more code in the future if we wan to get to a more advanced namespace support.

k8s/pkg/registryserver/registry_cache.go
```diff
@@ -173,3 +177,11 @@ func (rc *registryCacheImpl) Stop() {
 		stopFunc()
 	}
 }
+
+func getNamespace() string {
```
nickolaev: For example, this can be moved to such `k8s/pkg/namespace` package.

Issue #766, opened 2019-02-21
Split integration testing


nickolaev:
Our integration testing is a really important part of the release process. We run the tests in the CI here:
https://github.com/networkservicemesh/networkservicemesh/blob/master/.circleci/config.yml#L125

Since we add more and more tests, this step is taking a long time and is hard to track what is going on.
Therefore we can split the test execution into parts, differentiating by test's function:
 * basic - simple k8s tests (like [exect_test.go](https://github.com/networkservicemesh/networkservicemesh/blob/master/test/integration/exec_test.go)), basic nsmd deployment etc.
 * recover - all the healing tests ([death_handling_test.go](https://github.com/networkservicemesh/networkservicemesh/blob/master/test/integration/death_handling_test.go), [nse_dataplane_heal_test.go](https://github.com/networkservicemesh/networkservicemesh/blob/master/test/integration/nse_dataplane_heal_test.go), [nse_heal_test.go](https://github.com/networkservicemesh/networkservicemesh/blob/master/test/integration/nse_heal_test.go))
 * usecase - all the icmp and vpn tests
 
These are just proposed categories, one can come op with another split. The testfiles can be marked with [go build tags](https://golang.org/pkg/go/build/) and then each category can be run in a separate CircleCI `run` clause. Note we do not want separate jobs here, otherwise, we will end up in parallel execution and create problems. The test category selection happens by adding a flag `-tags basic` to the `go test` or `gotestsum` command line.

So each new test added has to choose a category and mark its file e.g. `// +build basic`. We can add a simple README in the test folder or in doc/spec if that is preferred, explaining the marking strategy.

New categories can be added or even split. I assume if we get enough usecase tests we can do `usecase_simple` and `usecase_complex` or similar.

edwarnicke:
I like very much looking at breaking up by thematic groups... using build tags for them is an interesting way to automate :)

edwarnicke:
@rdimitrov Was this resolved by #781 ?  If so, should we close this issue?

rdimitrov:
@edwarnicke Indeed, we should close it. It's already resolved by that PR 👍

PR #781, opened 2019-02-26
Split integration tests by context using build tags

rdimitrov:

## Description

* Added build tags to each integration test - either `basic`, `recover` or `usecase`
* Trigger a separate run step for each test group in Circle CI config.yml
* Added separate junit report files for each group (at the end the whole directory is archived as a build artifact, so it shouldn't be a problem if it's one or multiple junit files)

## Motivation and Context

Related to #766
## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


.circleci/config.yml
```diff
@@ -112,24 +112,45 @@ jobs:
...
+          command: |
+            gotestsum --junitfile ~/junit/integration-tests-basic.xml -f standard-verbose ./test/... -timeout 30m -failfast -tags=basic
+          no_output_timeout: 40m
+          environment:
+            KUBECONFIG: /home/circleci/project/data/kubeconfig
+            GO111MODULE: "on"
+      - run:
+          name: Running integration tests - recover
+          command: |
+            gotestsum --junitfile ~/junit/integration-tests-recover.xml -f standard-verbose ./test/... -timeout 30m -failfast -tags=recover
```
haiodo: Are we sure it do not overwrite file in second run?

haiodo: I suppose we need to use different files for different parts.

rdimitrov: Thanks for your feedback! Do you refer to the junit files, as they are indeed different for each test group?

PR #782, opened 2019-02-26
Cluster deployment

BoutheinaDab:
## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


BoutheinaDab:
A new way to deploy a k8s cluster, composed by a master and a worker VM, without using vagrant. Then, NSM will be deployed on that cluster.

.k8s.mk
```diff
@@ -41,7 +41,7 @@ include .vagrant.mk
 include .packet.mk
 
 # .null.mk allows you to skip the vagrant machinery with:
-# export CLUSTER_RULES_PREFIX=null
+ export CLUSTER_RULES_PREFIX=null
```
nickolaev: I don't think we want to default to this. Let's stick with Vagrant for the time being.

scripts/Two_VM_Deployment/.kube/config
```diff
@@ -0,0 +1,19 @@
...
+    server: https://172.31.16.96:6443
```
nickolaev: How is this IP generated?

scripts/Two_VM_Deployment/create-kubernetes-cluster.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash -x
```
nickolaev: Are these file copied from scripts/terraform?

.k8s.mk
BoutheinaDab: Yes, but for VM were nested virtualization can not be authorized, vagrant can not be used to create VM. That was the case for me when I started the set up of NSM, deploying vagrant VMs was not possible. That is way I was creating an other k8s cluster in tow separate VMs, and then deploy NSM on the master.
I think it is a novel way more generic to deploy NSM without using vagrant.

scripts/Two_VM_Deployment/.kube/config
BoutheinaDab: It is the IP generated in the file /etc/kubernetes/admin.conf.
Indeed, we the file .kube/config is a copy of /etc/kubernetes/admin.conf, in order to enable the k8s cluster.

scripts/Two_VM_Deployment/create-kubernetes-cluster.sh
BoutheinaDab: They are similar in some parts to files scripts/terraform, particularly install-kubernetes (in some parts), and download-worker-images.sh.
But the other files are different. the script create-kubernetes-cluster.sh keeps the same name, but the content is different. I
Do you suggest to change the name, or to remove the similarities? 
In fact, scripts in scripts/terraform or scripts/vagrant can not be used alone, in their current form, to deploy a non-vagrant 2-VMs K8s cluster (master+worker). That's the main insight of scripts/Two_VM_Deployment.

scripts/Two_VM_Deployment/.kube/config
nickolaev: But is that stable and reproducible? We have the option to use VBox, KVM and VMware Worksation/Fusion. Are they all enabling the same IP?

scripts/Two_VM_Deployment/create-kubernetes-cluster.sh
nickolaev: I would prefer to have a single set of scripts that can initialize *any* cluster for us. The script itself can get arguments when called so it modifies its behaviour accordingly.

scripts/Two_VM_Deployment/.kube/config
BoutheinaDab: I don't think that the IP would be the same.
I think it is better to remove the file .kube/config, because already the script copies its content from /etc/kubernetes/admin.conf. The latter depends on the VM on which k8s is installed I think.

.k8s.mk
nickolaev: Well, this particular line just changes the default `CLUSTER_RULES_PREFIX` from `vagrant` to `null`. We would prefer to keep `vagrant` as the default for now and allow interested people to explicitly `export CLUSTER_RULES_PREFIX=null` if they need to.

scripts/Two_VM_Deployment/create-kubernetes-cluster.sh
BoutheinaDab: Ok, so do you propose that I create a new set of scripts for both vagrant and non-vagrant k8s clusters? or to modify the current scripts/vagrant to do that?

nickolaev:
So in general, K8s cluster deployment is not NSM's core task here. There are better projects like https://kubespray.io/ that do that. We only needed these deployment scripts to init our development Vagrant environment and our CI/CD pipeline.

Essentially one can deploy the K8s cluster with whatever tools they see fit and the just:
```shell
export CLUSTER_RULES_PREFIX=null
make k8s-infra-deploy
```

This will download the images off the official docker hub and deploy to where `kubectl` is pointing. Of course, this method does not allow you to do local development and then deploy your images to that random cluster, but that is why we have Vagrant.

@edwarnicke can you please share your opinion here? Maybe we need a separate doc about external clusters explaining the above.

.k8s.mk
BoutheinaDab: Ok, I get it!

edwarnicke:
@BoutheinaDab I have a question... you have these scripts for configuring VMs once they are running to form a K8s cluster, but I don't see any hooks for *starting* VMs.  Thinking a bit about this we currently have two levels:

1)  I don't have a K8s cluster, and want a way to stand that up soup to nuts.  Vagrant is an example of this.  packet.mk was trying to be another
2)  I have a K8s cluster, the 'null' case handles this

It feels to me like you have an intermediate stage here where you have the VMs... you just need them setup as a proper K8s cluster.  Is that correct?  Out of curiosity, how are you getting your VMs?  I understand why Vagrant is a bad fit for you... I'm curious what you are using :)

BoutheinaDab:
> @BoutheinaDab I have a question... you have these scripts for configuring VMs once they are running to form a K8s cluster, but I don't see any hooks for _starting_ VMs. Thinking a bit about this we currently have two levels:
> 
> 1. I don't have a K8s cluster, and want a way to stand that up soup to nuts.  Vagrant is an example of this.  packet.mk was trying to be another
> 2. I have a K8s cluster, the 'null' case handles this
>
@edwarnicke Thank you for feedback. Indeed, yes. I am in the first case, I don't have k8s cluster. I have two ubuntu VMs that I want to use to deploy a k8s cluster. Using Vagrant was not possible, because I was deploying NSM on a ubuntu VM, where the nested virtualization is not enabled. 
> 
It feels to me like you have an intermediate stage here where you have the VMs... you just need them setup as a proper K8s cluster. Is that correct? Out of curiosity, how are you getting your VMs? I understand why Vagrant is a bad fit for you... I'm curious what you are using :)

Right. I skipped the intermediate stage of starting the VMs. Indeed, the VMs are already created and started, and I just use them to setup a proper k8s cluster. It is correct!
Regarding the VMs, I am using two VMware VMs (ubuntu 16.04).

edwarnicke:
@BoutheinaDab OK... so question... why not use the existing scripts in scripts/vagrant/scripts ... those aren't specific to Vagrant really... they just install K8s on two VMs... however they have been stood up.

nickolaev:
Well, I did not want to block the PR. I still think we do not need this. At least not in the proposed format. We already support Vagrant on VMware Workstation, VBox and KVM. We have recently added support for Kind as a cluster provider. We do support Helm Charts now. I think we have enough tools to cover development and basic deployment tasks.

I'll let @edwarnicke and @fkautz have the final say.

Please note that the PR has failing checks and conflicts, which renders it unmergeable as it is.

PR #786, opened 2019-02-28
Implement method in registry for updating NSMgr

lobkovilya:

## Description
Add method UpdateNSM to registry. Cache is also supported.

## Motivation and Context
In case of NSMgr dies and restarts it should update info about itself in registry (because URL could changed)

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

- [X] My change requires a change to the documentation.


controlplane/pkg/apis/registry/registry.proto
```diff
@@ -64,4 +64,5 @@ service NetworkServiceRegistry {
 
 service NetworkServiceDiscovery {
     rpc FindNetworkService (FindNetworkServiceRequest) returns (FindNetworkServiceResponse);
+    rpc UpdateNSM (NetworkServiceManager) returns (NetworkServiceManager);
```
edwarnicke: Why is this in the NetworkServiceDiscovery rather than NetworkServiceRegistration above?

edwarnicke: Ah... I think I see... because its not something that you would want the client from the NSE using.

Perhaps UpdateNSM should be in its own 'service' ?

Issue #715, opened 2019-02-01
Spec: Metrics in MonitorCrossConnects/MonitorConnections


matrohon:

matrohon:
https://docs.google.com/document/d/1hMoNRyMQvJd1IuAduG4lTbhRU5NdyYM82WpKz-XTkSI/edit?usp=sharing

edwarnicke:
Marked this as pinned so that stale-bot will leave it alone :)

haiodo:
I think this should be closed already.

PR #802, opened 2019-03-05
WIP: Metrics in MonitorCrossConnects

platoff:
Implementation of Metrics in MonitorCrossConnects/MonitorConnections #715 

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


controlplane/pkg/monitor/monitor_server.go
```diff
@@ -12,6 +14,11 @@ const (
 	INITIAL_STATE_TRANSFER = "INITIAL_STATE_TRANSFER"
 )
 
+const (
+	src_prefix = "SRC-"
+	dst_prefix = "DST-"
```
nickolaev: Where these come from? We should maintain common constant definitions across all packages we generate. Please consider refactoring the other modules if needed.

controlplane/pkg/monitor/monitor_server.go
```diff
@@ -21,6 +28,11 @@ type Event struct {
 	Entities  map[string]Entity
 }
 
+type Stat struct {
```
nickolaev: Can we have a more elaborate name of the structure here? `Stat` is pretty short. Also do we need this type exported?

controlplane/pkg/monitor/monitor_server.go
```diff
@@ -33,6 +45,8 @@ type MonitorServer interface {
 	Update(entity Entity)
 	Delete(entity Entity)
 
+	UpdateStat(name string, statistics *interfaces.InterfacesState_Interface_Statistics)
```
nickolaev: I believe this should be plural: `UpdateStats` or even better `UpdateStatistics`

platoff: @nickolaev what do you think about bad abstractions here. I'd prefer to remove monitor server and use concrete implementation instead. Any thoughts on this?

controlplane/pkg/monitor/monitor_server.go
edwarnicke: If these are related to the prefixes we happened to have selected in vppagent-dataplane... I'd prefer *not* to presume upon them, and they may at some point change...

controlplane/pkg/monitor/monitor_server.go
edwarnicke: Or possibly UpdateMetrics ? (brainstorming :) )

nickolaev:
#980 supersedes this one. Closing.

PR #813, opened 2019-03-07
Fixes for CircleCI and Packet

nickolaev:
## Description
An attempt to optimize (and speed up?) the CI Packet deployment.

## Motivation and Context
Terraform on Packet was saying there are deprecated options. Fixing this, went also through the state of the circleci.yaml and made it use the latest golang image. It is the same approach as when we build the files, so better consistency.
Packet instances are explicitly named to t1.small.x86 and the k8s installation script is simplified.

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


.circleci/config.yml
```diff
@@ -42,40 +42,9 @@ jobs:
     environment:
       GO111MODULE: "on"
```
edwarnicke: Why cut out all the sanity checks?

edwarnicke:
Switching from Ubuntu 16.04 to Ubuntu 18.04 takes us out fo the 'fast deploy' scenario for Packet... which means an extra 2+ minutes on the packet-deploy job and thus the overall CI time...

.circleci/config.yml
nickolaev: @edwarnicke there are 2 separate jobs `sanity-check` and `build`. Apparently, the latter is never referenced and looks like a stale historical piece just laying around. I propose to remove it.

nickolaev:
OK so 18.04 is reverted.

PR #818, opened 2019-03-08
parametrize docker image ligato/vpp-agent to satisfy also ARM64 platform

stanislav-chlebec:
Signed-off-by: Stanislav Chlebec <stanislav.chlebec@pantheon.tech>

Where the docker image is used ligato/vpp-agent in different Dockerfiles it was replaced by parameter.
The parameter is going to be set up in the file .mk/docker.mk 

## Description
Parameter VPP_AGENT is used to define proper docker image according to detected platform x86_64 or aarch64

## Motivation and Context
This change is required if we want to have docker building process on ARM64 platform executed by command: make k8s_save

## How Has This Been Tested?
- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:
It should be somewhere mentioned where is the point where it is possible to change version of ligato/vpp-agent docker image


.mk/docker.mk
```diff
@@ -20,13 +20,19 @@ RUN_CONTAINERS=$(BUILD_CONTAINERS)
 KILL_CONTAINERS=$(BUILD_CONTAINERS)
 LOG_CONTAINERS=$(KILL_CONTAINERS)
 ORG=networkservicemesh
+ifeq ($(shell uname -m), x86_64)
+  VPP_AGENT=ligato/vpp-agent:v1.8.1
+endif
+ifeq ($(shell uname -m), aarch64)
```
nickolaev: Can we define a global so I can override it if I decide to:
```Makefile
ARCH ?= $(shell uname -m)
```

.mk/docker.mk
```diff
@@ -86,7 +92,7 @@ docker-devenv-attach:
 docker-%-debug:
 	@docker exec -ti $$(docker container ls | grep networkservicemesh/devenv | cut -c1-12) /go/src/github.com/networkservicemesh/networkservicemesh/scripts/debug.sh $*
 
-.PHONY: docker-%-push
+.PHONY: docker-push-%
```
nickolaev: That is wrong. Please revert it.

.mk/docker.mk
```diff
@@ -20,13 +20,19 @@ RUN_CONTAINERS=$(BUILD_CONTAINERS)
 KILL_CONTAINERS=$(BUILD_CONTAINERS)
 LOG_CONTAINERS=$(KILL_CONTAINERS)
 ORG=networkservicemesh
+ifeq ($(shell uname -m), x86_64)
+  VPP_AGENT=ligato/vpp-agent:v1.8.1
+endif
+ifeq ($(shell uname -m), aarch64)
+  VPP_AGENT=ligato/vpp-agent-arm64:dev
```
nickolaev: Interesting, why the `:dev` here? Are we going to get reproducible results like this?

edwarnicke:
@rdimitrov could you check this to see if it runs?

rdimitrov:
@edwarnicke I will try it 👍

.mk/docker.mk
stanislav-chlebec: I modified according your suggestion

.mk/docker.mk
stanislav-chlebec: Reverted.

.mk/docker.mk
stanislav-chlebec: You are right, no reproducible result.
But the ligato team is now working on version 2 which is more prospective
So I should maybe use https://github.com/ligato/vpp-agent/tree/v2.0.0-beta 
Uses
VPP_REPO_URL=https://github.com/FDio/vpp.git
VPP_COMMIT=a8e3001e68d8f5ea6cf526b131c92f5993597f81 # branch stable/1810

stanislav-chlebec: Actually the same version of VPP as 1.8.1 ....

stanislav-chlebec: docker image ligato/vpp-agent-arm64:dev is built on the branch dev of repo https://github.com/ligato/vpp-agent

stanislav-chlebec: I set the version for arm64 also to v1.8.1

rdimitrov:
@stanislav-chlebec @edwarnicke @nickolaev 
Built the images on both ARM64 and x86-64 and it went successful, so okay from my side. 
Nice! 👍 

PS. 
Just noticed a couple of small things:
* Passing a `build-arg` introduces a warning when the arg is not used, for ex. for each non-vppagent related build:
```[Warning] One or more build-args [VPP_AGENT] were not consumed```
* For ARM64, the vppagent image doesn't have the kafka config file and we try to delete it, so during build time we have:
```rm: cannot remove '/opt/vpp-agent/dev/kafka.conf': No such file or directory```
* Resolve the failed DCO check

nickolaev:
@stanislav-chlebec are you canceling this PR?

stanislav-chlebec:
Well it seems that yes
It happened due to my inexperience with git☺
I attempted to sign-off one of those commit but I did it in a bad way
I am not sure if it is possible to revert
If not I will reconstruct it and prepare a new PR.
Sorry.
Stan

nickolaev:
@stanislav-chlebec ` git checkout -b recover 5fa131c`?

rdimitrov:
Here it is - https://pastebin.com/vzKL661Q

stanislav-chlebec:
Thanks
It helped.
S.

PR #821, opened 2019-03-10
Make the kind guide slightly more generic

fkautz:


nickolaev:
It's a message that is auto generated. Let's merge it though.

Issue #801, opened 2019-03-05
NSMDp should support >= 10 clients.


haiodo:
Failed to deploy NSE or Client happened because of:

Warning  UnexpectedAdmissionError  27s   kubelet, kube-worker  Update plugin resources failed due to requested number of devices unavailable for networkservicemesh.io/socket. Requested: 1, Available: 0, which is unexpected.

It seems DeviceAPI plugin could not give a workspace sometimes.

* [Feature](?template=spec.md)
* [Bug](?template=bug.md)

Once switched to the correct template, keep in mind that switching templates will remove all already entered data within this issue.

Thank you for reaching out!

haiodo:
Changed logic in PR #795 as experiment, to return always more >= 10 devices in compare to allocated ones, but still had same issue happening for

haiodo:
Will try tomorrow to use: 
https://kubernetes.io/docs/concepts/extend-kubernetes/compute-storage-net/device-plugins/#monitoring-device-plugin-resources to check for resources usage, also this seems should help to track died Clients/NSEs.

haiodo:
Also it seems we need to monitor kubelet restart situations.

haiodo:
This one is done. @edwarnicke  @nickolaev  could you please resolve. Thanks.

PR #823, opened 2019-03-10
#801 Fixes NSMDp

haiodo:
Fixed NSMDp if more of 10 devices are required.

## Description
Fixed NSMDp if more of 10 devices are required.

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/k8s/cmd/nsmdp/nsmdp.go
+++ b/k8s/cmd/nsmdp/nsmdp.go
@@ -43,8 +43,11 @@ const (
 	// ServerSock defines the name of NSM client socket
 	ServerSock = "networkservicemesh.io.sock"
 
-	// A number of devices we have in buffer for use.
+	// A number of devices we have in buffer for use, so we hold extra DeviceBuffer count of deviceids send to kubelet.
 	DeviceBuffer = 30
+
+	// Send device ids to kubelet every N seconds.
+	KubeletNotifyDelay = 30 * time.Second
 )
 
 type nsmClientEndpoints struct {
@@ -154,11 +157,11 @@ func (n *nsmClientEndpoints) ListAndWatch(e *pluginapi.Empty, s pluginapi.Device
 
 	// Restore state from NSMD
 	for {
-		n.reciveWorkspaces()
+		n.receiveWorkspaces()
 		n.sendDeviceUpdate()
 
-		//
-		time.Sleep(30 * time.Second)
+		// Sleep before next notification.
+		time.Sleep(KubeletNotifyDelay)
 	}
 }
 
@@ -262,7 +265,7 @@ func (n *nsmClientEndpoints) sendDeviceUpdate() {
 	}
 }
 
-func (n *nsmClientEndpoints) reciveWorkspaces() {
+func (n *nsmClientEndpoints) receiveWorkspaces() {
 	for {
 		reply, err := enumWorkspaces(n.serviceRegistry)
 		if err != nil {
--- a/k8s/cmd/nsmdp/nsmdp.go
+++ b/k8s/cmd/nsmdp/nsmdp.go
@@ -17,19 +17,20 @@ package main
 
 import (
 	"fmt"
+	"github.com/networkservicemesh/networkservicemesh/controlplane/pkg/apis/nsmdapi"
 	"net"
 	"os"
 	"os/signal"
 	"path"
+	"sync"
 	"syscall"
 	"time"
 
 	"github.com/grpc-ecosystem/grpc-opentracing/go/otgrpc"
-	"github.com/networkservicemesh/networkservicemesh/controlplane/pkg/apis/nsmdapi"
 	"github.com/networkservicemesh/networkservicemesh/controlplane/pkg/nsmd"
 	"github.com/networkservicemesh/networkservicemesh/controlplane/pkg/serviceregistry"
 	"github.com/networkservicemesh/networkservicemesh/pkg/tools"
-	opentracing "github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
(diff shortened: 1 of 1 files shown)


haiodo:
@nickolaev @edwarnicke  Could you take a look Please, thanks.

k8s/cmd/nsmdp/nsmdp.go
```diff
@@ -41,10 +42,18 @@ const (
 	resourceName = "networkservicemesh.io/socket"
 	// ServerSock defines the name of NSM client socket
 	ServerSock = "networkservicemesh.io.sock"
+
+	// A number of devices we have in buffer for use.
+	DeviceBuffer = 30
```
nickolaev: Is this an arbitrary number? What if we need 31?

k8s/cmd/nsmdp/nsmdp.go
```diff
@@ -141,34 +150,14 @@ func indexOf(slice []string, value string) int {
...
-				Health: pluginapi.Healthy,
-			})
-		}
-		if err := s.Send(resp); err != nil {
-			logrus.Errorf("Failed to send response to kubelet: %v\n", err)
-		}
+		n.reciveWorkspaces()
+		n.sendDeviceUpdate()
+
+		//
```
nickolaev: nit

k8s/cmd/nsmdp/nsmdp.go
```diff
@@ -231,6 +226,78 @@ func NewNSMDeviceServer(serviceRegistry serviceregistry.ServiceRegistry) error {
...
+}
+
+func (n *nsmClientEndpoints) reciveWorkspaces() {
+	for {
+		reply, err := enumWorkspaces(n.serviceRegistry)
+		if err != nil {
+			logrus.Errorf("Error receive devices from NSM. %v", err)
+			// Make a fast delay to faster startup of NSMD.
+			<-time.Tick(100 * time.Millisecond)
+			continue
```
nickolaev: What are the chances we end up in an endless loop here?

k8s/cmd/nsmdp/nsmdp.go
```diff
@@ -231,6 +226,78 @@ func NewNSMDeviceServer(serviceRegistry serviceregistry.ServiceRegistry) error {
...
+		}
```
nickolaev: you actually can remove the `continue` and move the end of the for cycle with `break` here. The rest of the code is not repeated in any case.

k8s/cmd/nsmdp/nsmdp.go
```diff
@@ -231,6 +226,78 @@ func NewNSMDeviceServer(serviceRegistry serviceregistry.ServiceRegistry) error {
...
+	defer nsm.mutext.Unlock()
+
+	for {
+		nsm.clientId++
+
+		id := fmt.Sprintf("nsm-%d", nsm.clientId)
+		if _, ok := nsm.devs[id]; ok {
+			// Item already exists, increment to new one.
+			continue
+		}
```
nickolaev: We can close the for here.

nickolaev: and can we prevent the endless loops.

k8s/cmd/nsmdp/nsmdp.go
```diff
@@ -231,6 +226,78 @@ func NewNSMDeviceServer(serviceRegistry serviceregistry.ServiceRegistry) error {
...
+	if n.pluginApi != nil {
+		api := *n.pluginApi
+		logrus.Infof("Send device plugins intfo %v", n.resp)
+		if err := api.Send(n.resp); err != nil {
+			logrus.Errorf("Failed to send response to kubelet: %v\n", err)
+		}
+	}
+}
+
+func (n *nsmClientEndpoints) reciveWorkspaces() {
```
rdimitrov: s/recive/receive

PS.  And probably something a bit more cosmetic, but acronyms like API, NSM or NSMD should be all uppercase based on Go's naming convention. There's a couple of places that you can update to resolve that both here and in the PR before this one 👍

k8s/cmd/nsmdp/nsmdp.go
haiodo: On every allocation, it adds +1 device. It is buffer to hold +30 available ones.

k8s/cmd/nsmdp/nsmdp.go
haiodo: Thanks, nsm here is name of internal structure, so I could not make it uppercase.

k8s/cmd/nsmdp/nsmdp.go
haiodo: We could not close here, assume after restart we recieved nsm-1,...nsm-30, nsm-40 and all already used, so we need to add more items line nsm-41,nsm-42.

k8s/cmd/nsmdp/nsmdp.go
haiodo: Nope, If we remove continue here, it will be delay 30seconds before kubelet will receive a initial list of devices, since all of previous ones could be already used by nsmd by restore. This will slowdown all test in case NSMD will not yet start. So this continue allow to wait for NSMD startup, like 3-5 seconds before we will do a huge delay.

k8s/cmd/nsmdp/nsmdp.go
haiodo: If NSMD will not respond properly, nothing will work in any case, so it is ok.

k8s/cmd/nsmdp/nsmdp.go
rdimitrov: @haiodo I was referring to the case when the acronym is in the middle of the namings, for ex. `deployNsmdAndDataplane` should be `deployNSMDAndDataplane`

PR #832, opened 2019-03-12
WIP: Adapt vppagent-nsc to admission webhook changes

matrohon:

vppagent-nsc doesn't work anymore with the introduction of admission-webhook

## Description
This change makes the vppagent-nsc working

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/examples/cmd/vppagent-nsc/nsc.go
+++ b/examples/cmd/vppagent-nsc/nsc.go
@@ -70,7 +70,7 @@ func main() {
 		vppAgentEndpoint: defaultVPPAgentEndpoint,
 	}
 
-	client, err := client.NewNSMClient(nil, nil)
+	clientList, err := client.NewNSMClientList(nil, nil)
 	if err != nil {
 		logrus.Fatalf("Unable to create the NSM client %v", err)
 	}
@@ -80,15 +80,16 @@ func main() {
 		logrus.Fatalf("Unable to create the backend %v", err)
 	}
 
-	var outgoingConnection *connection.Connection
-	outgoingConnection, err = client.Connect("if1", "mem", "Primary interface")
+	err = clientList.Connect("if1", "mem", "Primary interface")
 	if err != nil {
 		logrus.Fatalf("Unable to connect %v", err)
 	}
 
-	err = backend.Connect(outgoingConnection)
-	if err != nil {
-		logrus.Fatalf("Unable to connect %v", err)
+	for _, conn := range clientList.GetConnections() {
+		err = backend.Connect(conn)
+		if err != nil {
+			logrus.Fatalf("Unable to connect %v", err)
+		}
 	}
 
 	logrus.Info("nsm client: initialization is completed successfully, wait for Ctrl+C...")
--- a/k8s/conf/admission-webhook.yaml
+++ b/k8s/conf/admission-webhook.yaml
@@ -22,6 +22,9 @@ spec:
             - name: webhook-certs
               mountPath: /etc/webhook/certs
               readOnly: true
+          env:
+            - name: INITCONTAINER
+              value: vppagent-nsc
       volumes:
         - name: webhook-certs
           secret:
--- a/sdk/client/clientlist.go
+++ b/sdk/client/clientlist.go
@@ -98,3 +98,7 @@ func (nsmc *NsmClientList) Connect(name, mechanism, description string) error {
 	}
 	return nil
 }
+
+func (nsmc *NsmClientList) GetConnections() []*connection.Connection {
+	return nsmc.connections
+}


k8s/conf/admission-webhook.yaml
```diff
@@ -22,6 +22,9 @@ spec:
             - name: webhook-certs
               mountPath: /etc/webhook/certs
               readOnly: true
+          env:
+            - name: INITCONTAINER
+              value: vppagent-nsc
```
matrohon: This is just to easily test this patch. It will be removed before merging

examples/cmd/vppagent-nsc/nsc.go
```diff
@@ -70,7 +70,7 @@ func main() {
 		vppAgentEndpoint: defaultVPPAgentEndpoint,
 	}
 
-	client, err := client.NewNSMClient(nil, nil)
+	clientList, err := client.NewNSMClientList(nil, nil)
```
nickolaev: Why not calling the simple Client? Are you expecting the NSC will have to hanlde multiple connections?

examples/cmd/vppagent-nsc/nsc.go
```diff
@@ -80,15 +80,16 @@ func main() {
...
-	outgoingConnection, err = client.Connect("if1", "mem", "Primary interface")
+	err = clientList.Connect("if1", "mem", "Primary interface")
 	if err != nil {
 		logrus.Fatalf("Unable to connect %v", err)
 	}
 
-	err = backend.Connect(outgoingConnection)
-	if err != nil {
-		logrus.Fatalf("Unable to connect %v", err)
+	for _, conn := range clientList.GetConnections() {
```
nickolaev: Again, I don't get it why we need all this? I would expect we have a single connection.

sdk/client/clientlist.go
```diff
@@ -98,3 +98,7 @@ func (nsmc *NsmClientList) Connect(name, mechanism, description string) error {
 	}
 	return nil
 }
+
+func (nsmc *NsmClientList) GetConnections() []*connection.Connection {
```
nickolaev: That is interesting. shall we just extend Connect to return this list alongside the error?

examples/cmd/vppagent-nsc/nsc.go
matrohon: ditto

examples/cmd/vppagent-nsc/nsc.go
matrohon: This nsc can be used to replace the ["standard"](https://github.com/networkservicemesh/networkservicemesh/blob/master/examples/cmd/nsc/nsc.go) nsc. It should be able to handle multiple "annotations", don't you think?

examples/cmd/vppagent-nsc/nsc.go
matrohon: Your proposal makes sense, since sdk.client.Connect() returns a Connection, sdk.clientList.Connect() should return a slice of Connection.

I'll update my patch this way.

examples/cmd/vppagent-nsc/nsc.go
nickolaev: Ehm, I think this is an example of a `smart` Client, i.e. one that does not use an external init Container. I am not sure if it is even possible that we can have an initContainer with MemIf.
That said - It can be perfectly fine having multiple connections as lomng as we find a way to demo/test it. For example, can we add this container in our tests under usecases? I know we barely test anythign related to MemIf today - only the Firewall is there. Every other test uses nsc and icmp-responders with Kernel interfaces.

edwarnicke:
I'm a bit confused.  The vppagent-nsc is 'smart' ... its got a vpp running inside it and the vpp is what connects to the networkservice.  Its not an init container, or a 'non-nsm-aware' app that needs things set up for it by an NSC.  So I'm not entirely clear on what the web hook would do for it.

nickolaev:
@edwarnicke exactly my point https://github.com/networkservicemesh/networkservicemesh/pull/832#discussion_r265034784

sdk/client/clientlist.go
matrohon: Your proposal makes sense, since sdk.client.Connect() returns a Connection, sdk.clientList.Connect() should return a slice of Connection.

I'll update my patch this way.

matrohon:
@edwarnicke I'll remove the part that makes vppagent-nsc used as an InitContainer. But this patch still makes sense in order the vppagent-nsc to be used with the admission webhook changes, am I wrong?

examples/cmd/vppagent-nsc/nsc.go
matrohon: This mean this vppagent-nsc is not tested in the CI, since the usage of memif is hardcoded in it?

matrohon:
Actually, the question behind this patch would be : do we want nsm aware application to ask for a network service the same way nsm unaware application does, through an [annotation](https://github.com/networkservicemesh/networkservicemesh/blob/master/k8s/conf/vpn-gateway-nsc.yaml#L22) instead of dedicated [env variables](https://github.com/networkservicemesh/networkservicemesh/blob/master/k8s/conf/vppagent-nsc.yaml#L20) ?

nickolaev:
@matrohon that is a good question, which I was consudering treating by changing the SDK to leverage `https://github.com/spf13/viper`. It allows us to define ENV, config map and possibly CLI configuration means all handled at once. Looks great.
The big problem as always is `naming things`. Our current ENV names are far from great. Ideas?

matrohon:
@nickolaev viper is definitely a good lib to consider. Do we want to integrate it for v1?

Concerning the ENV used by the sdk, I totally agree that for now their name are not explicit. For instance, as a client, I would just want to say what network service I want to consume. For now it is specified as OUTGOING_NSC_NAME, am I wrong? If I'm right, something like NSM_NETWORK_SERVICE would be more appropriate, don't you think?

matrohon:
since the admission-webhook process is about adding an initContainer, which is not needed for vppagent-nsc that directly uses the sdk (annotation/initContainer not needed in this context), and since the SDK configuration is now handled by #866 , I think we can close this PR

Issue #855, opened 2019-03-19
VPP Agent Dataplane is failed to create Kernel interface connection


haiodo:
## Expected Behavior

VPP Agent Dataplane is failed to create kernel interface based but return success status to NSMD.

## Current Behavior

VPP Agent Dataplane is failed to create kernel interface based but return success status to NSMD.

## Steps to Reproduce

Execute integration tests inside GKE.

## Failure Logs

Failed case:
```
[...]

PR #857, opened 2019-03-19
VPP Agent Dataplane VETH mode is not working

haiodo:

Fix VPPAgent Dataplane VETH mode.

## Description

The veth devices are virtual Ethernet devices.  They can act as tunnels between network namespaces to create a bridge to a physical network device in another namespace, but can also be used as standalone network devices.

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/dataplane/vppagent/pkg/converter/kernel_connection_converter.go
+++ b/dataplane/vppagent/pkg/converter/kernel_connection_converter.go
@@ -2,15 +2,16 @@ package converter
 
 import (
 	"fmt"
-	"os"
-
 	linux_interfaces "github.com/ligato/vpp-agent/plugins/linux/model/interfaces"
 	"github.com/ligato/vpp-agent/plugins/linux/model/l3"
 	"github.com/ligato/vpp-agent/plugins/vpp/model/interfaces"
 	"github.com/ligato/vpp-agent/plugins/vpp/model/rpc"
 	"github.com/networkservicemesh/networkservicemesh/controlplane/pkg/apis/local/connection"
 	"github.com/sirupsen/logrus"
+	"os"
 )
+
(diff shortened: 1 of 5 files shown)


haiodo:
@nickolaev Could you please review, latest failure are caused by #855

dataplane/vppagent/pkg/converter/kernel_connection_converter.go
```diff
@@ -110,7 +111,7 @@ func (c *KernelConnectionConverter) ToDataRequest(rv *rpc.DataRequest, connect b
 			IpAddresses: ipAddresses,
 			HostIfName:  tmpIface,
 			Veth: &linux_interfaces.LinuxInterfaces_Interface_Veth{
-				PeerIfName: m.GetParameters()[connection.InterfaceNameKey],
+				PeerIfName: c.conversionParameters.Name,
```
edwarnicke: @haiodo So this was the root issue?  It looks like the rest of the chain is adding testing (which is good!).

nickolaev:
I am a bit worried about 'DATAPLANE_ALLOW_VHOST' being added just for testing purposes. However there does not seem to be a better way forward.
I'll let the tests pass and merge.

dataplane/vppagent/pkg/converter/kernel_connection_converter.go
haiodo: Yes, this is root cause for GKE for not working at all, but still we had #855 and it still not working in 50% of times.

haiodo:
@nickolaev Yes, DATAPLANE_ALLOW_VHOST is used only for tests right now to check all scenarios in CI. Potentially it could be used in case vhost is not working properly, but not sure it will be.

nickolaev:
The well-known not-passing tests are failing. Looking at the code all seems fine. Merging.

haiodo:
Thanks.

Issue #860, opened 2019-03-19
Device Plugin leads to tests failure


lobkovilya:
Following tests are not stable:
* `TestNSMDDP`
* `TestNSMDRecoverNSE` 
* `TestNSMHealRemoteDieNSMD_NSE` 

All they do something like:
```
k8s.DeletePods(nodes_setup[1].Nsmd)
k8s.DeletePods(icmpPod)

nodes[0].Nsmd = k8s.CreatePod(pods.NSMgrPodWithConfig(nsmdName, nodes[0].Node, &pods.NSMgrPodConfig{})) // Recovery NSEs
icmpPod = nsmd_test_utils.DeployICMP(k8s, nodes[0].Node, "icmp-responder-nse-2", defaultTimeout)
```

Fail happens during attempt to deploy `icmp-responder-nse-2`, because there is no resource for it.

But adding sleep solves the problem:
```
k8s.DeletePods(nodes_setup[1].Nsmd)
k8s.DeletePods(icmpPod)

time.Sleep(10 * time.Second)

nodes[0].Nsmd = k8s.CreatePod(pods.NSMgrPodWithConfig(nsmdName, nodes[0].Node, &pods.NSMgrPodConfig{})) // Recovery NSEs
icmpPod = nsmd_test_utils.DeployICMP(k8s, nodes[0].Node, "icmp-responder-nse-2", defaultTimeout)
```

artem-belov:
Pod deleting with GracePeriod=0 does not guarantee of pod full termination.

> warning: Immediate deletion does not wait for confirmation that the running resource has been terminated. The resource may continue to run on the cluster indefinitely.

Removing that delete option resolve the problem

nickolaev:
@lobkovilya @artem-belov is there a PR coming?
Also, is there a chance that the tests hang for stale pod not terminating?

artem-belov:
@nickolaev Pull request is created #861 

> Also, is there a chance that the tests hang for stale pod not terminating?

Tests will continue working with warning after timeout. The most likely all required resources will be freed in timeout even if pod is not fully deleted.

edwarnicke:
@artem-belov @lobkovilya Is this issue still occurring?

lobkovilya:
I believe it is not occurring (at least at packet.net), fixed by @artem-belov in https://github.com/networkservicemesh/networkservicemesh/pull/861

PR #861, opened 2019-03-20
Fix nsmd pod deletion in tests (#860)

artem-belov:
## Description
If you are deleting pod with Grace Period 0 - the Pod may not be terminated instantly.
It may return error next time, when nsmdp will try to allocate not freed resource.

## Motivation and Context
Following tests are not stable:

- TestNSMDDP
- TestNSMDRecoverNSE
- TestNSMHealRemoteDieNSMD_NSE

With the fix test will wait while previous nsmd pod will be terminated before creating new one. 

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/test/kube_testing/k8s.go
+++ b/test/kube_testing/k8s.go
@@ -24,7 +24,7 @@ import (
 
 const (
 	podStartTimeout  = 1 * time.Minute
-	podDeleteTimeout = 1 * time.Minute
+	podDeleteTimeout = 15 * time.Second
 	podExecTimeout   = 1 * time.Minute
 	podGetLogTimeout = 1 * time.Minute
 )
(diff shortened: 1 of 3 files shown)


test/kube_testing/k8s.go
```diff
@@ -386,6 +413,21 @@ func (l *K8s) CreatePod(template *v1.Pod) *v1.Pod {
 	return results[0]
 }
 
+func (l *K8s) DeletePodsSafe(timeout int, pods ...*v1.Pod) {
```
nickolaev: Please remove this function. Instead, change `DeletePods` to always pass a const timeout, defined appropriately in `const.go`.

artem-belov: I am using podDeleteTimeout const which was already exists in k8s.go
Now force delete will be started in 15 seconds timeout after soft delete attempt for all pods

test/kube_testing/k8s.go
```diff
@@ -268,21 +268,45 @@ func NewK8s() (*K8s, error) {
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// Delete POD with completion check
+// Make force delete on timeout
 func (o *K8s) deletePods(pods ...*v1.Pod) error {
 	for _, pod := range pods {
```
nickolaev: Can we try spawning `goroutines` for each of the pod to be deleted? Then, of course, wait for all fo them to finish. I hope this can speed up the termination a bit.

artem-belov: Done. Now we are deleting all pods in parallel

nickolaev:
Comparing the test performance results:

|Test|c05a368|master(b88d9ee02a3445a6f00e9a93586a614193aa7b94)| Increase (%)|
|-|-|-|-|
|basic| 10.38 | 5.21 | 199% |
|recover| 12.37 | 7.26 | 170% |
|usecase| 6.28 | 4.52| 139% |

I think we should seriously reconsider these numbers before we merge this PR.

lobkovilya:
Probably we should split work in 2 PRs, the first one will fix `deletePods` only for 3 tests (mentioned in issue https://github.com/networkservicemesh/networkservicemesh/issues/860) and the second will set default `deletePods` method to "soft-mode" for the rest of the tests

artem-belov:
@nickolaev Test duration Increasing is necessary for proper test working. We have to wait while resources are not freed by previous pod. We can keep such deletion way for nsmd pods only but I think we should keep it for all pods to avoid that issue in another tests.

nickolaev:
@artem-belov can you please rebase this one on the latest master. Let's see how this plays out with the parallelised test run.

edwarnicke:
@artem-belov +1 to rebase and lets see how it goes :)

PR #880, opened 2019-03-26
Bump cleanup signal handling before anything else

rdimitrov:

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/dataplane/vppagent/cmd/vppagent-dataplane.go
+++ b/dataplane/vppagent/cmd/vppagent-dataplane.go
@@ -25,8 +25,6 @@ import (
 )
 
 func main() {
-	go common.BeginHealthCheck()
-
 	// Capture signals to cleanup before exiting
 	c := make(chan os.Signal, 1)
 	signal.Notify(c,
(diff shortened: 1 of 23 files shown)


lobkovilya:
Probably order of health checks should be changed also in https://github.com/networkservicemesh/networkservicemesh/blob/master/controlplane/cmd/nsmd/nsmd.go, because `logrus.Fatalf` calls `os.Exit(1)`

rdimitrov:
@lobkovilya Thanks for the feedback 👍 
Do you suggest swapping the `Fatalf` with `Warnf` or to put the health check after everything fatal-related?
Basically, even now in case the `Fatalf` exits, the probing server will die causing the probes checking to fail as well.

lobkovilya:
As for me, I believe that behavior of `Fatalf` is not obvious enough, so I'd prefer `Warnf` or `Errorf` and simple `return`.
Also to avoid duplicates I think it's better to call `defer nsmd.SetAPIServerFailed()` at the beginning

dataplane/vppagent/cmd/vppagent-dataplane.go
```diff
@@ -35,6 +33,8 @@ func main() {
 		syscall.SIGTERM,
 		syscall.SIGQUIT)
 
+	go common.BeginHealthCheck()
```
edwarnicke: Thanks for fixing this... I'm sensitive to it because I've been bitten by it before ;)

Issue #319, opened 2018-10-08
NSM CI on Azure


edwarnicke:
[Cross Cloud-CI](https://github.com/crosscloudci/cross-cloud) runs on Azure.  As an extension of #316 it would be good to get NSM CI working on Azure as well.

stepan-anokhin:
@edwarnicke I'll take a look.

nickolaev:
#887 merged. Closing.

PR #887, opened 2019-03-29
Add Azure Support

stepan-anokhin:
Signed-off-by: Stepan Anokhin <stepan.anokhin@gmail.com>

Add Azure Support

## Description
1. Create/Destroy AKS clusters
2. Build NSM with Azure Pipeline CI
3. Make sure all tests pass on Azure CI

## Motivation and Context
See #319 

## How Has This Been Tested?
- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


.mk/k8s.mk
```diff
@@ -158,7 +159,7 @@ k8s-config: $(addsuffix -config,$(addprefix k8s-,$(CLUSTER_CONFIGS)))
 .PHONY: k8s-start
 k8s-start: $(CLUSTER_RULES_PREFIX)-start
 
-.PHONY: k8s-start
+.PHONY: k8s-restart
```
edwarnicke: Nice catch :)

scripts/azure/env.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+# get scripts location
+export KUBECONFIG="$( cd "$(dirname "$0")" ; pwd -P )/.kube/config"
```
edwarnicke: You will also probably need to do something like [this](https://github.com/networkservicemesh/networkservicemesh/blob/master/.env/gke.env) in terms of setting various env variables that will cause the make machinery to use the azure env.

stepan-anokhin: This is already [done](https://github.com/networkservicemesh/networkservicemesh/blob/e12916a3be66a9614b34aedf12150859056ce1d2/.env/azure.env)

stepan-anokhin:
Only a licensing check failed during the last build. 
Also I've created an issue #961 to fix and enable remote-nsmd-healing tests currently disabled for AKS and GKE

test/integration/recover_remote_nse_nsm_heal_test.go
```diff
@@ -0,0 +1,136 @@
+// +build recover,!azure
```
nickolaev: That's fine, but we really have to pursue fixing these later. Having a problem in one cloud probably means that is hidden in the others.

test/kube_testing/k8s.go
```diff
@@ -25,7 +25,7 @@ import (
 )
 
 const (
-	podStartTimeout  = 1 * time.Minute
+	podStartTimeout  = 3 * time.Minute
```
nickolaev: OK, we move these timeouts now and then.  I hope this does not make tests too slow.

test/integration/recover_nse_nsm_heal_test.go
```diff
@@ -1,10 +1,9 @@
-// +build recover,!gke
+// +build recover
```
nickolaev: GKE is fixed, great :)

test/integration/const.go
```diff
@@ -4,6 +4,6 @@ import "time"
 
 const (
 	ciDelayCoefficient = 1
-	defaultTimeout     = 60 * time.Second * ciDelayCoefficient
+	defaultTimeout     = 2 * time.Minute * ciDelayCoefficient
```
nickolaev: Oh, bumping up timeouts. Not great.

test/integration/nsmd_test_utils/test_utils.go
```diff
@@ -523,7 +523,7 @@ func IsMemifNsePinged(k8s *kube_testing.K8s, from *v1.Pod) (result bool) {
 	}
 	logrus.Infof("nse ip: %v", nseIp)
 	logrus.Infof(" %v trying vppctl ping to %v", from.Name, nseIp)
-	for attempts := 30; attempts > 0; <-time.Tick(300 * time.Millisecond) {
+	for attempts := 60; attempts > 0; <-time.Tick(300 * time.Millisecond) {
```
nickolaev: Ugh, we need more attempts in order to succeed. Should not hurt in general, but is there a bigger problem hiden here? Or is it just the instances are slow?

docs/azure-quickstart.md
```diff
@@ -0,0 +1,190 @@
...
+for more details.
+
+Using `make` machinery:
+```bash
+export AZURE_RESOURCE_GROUP=nsm-ci
+export AZURE_CLUSTER_NAME=nsm-ci-cluster
+export AZURE_SERVICE_PRINCIPAL=1fe55163-6f8c-4592-8e9f-5b9cab7e39f4
+export AZURE_SERVICE_PRINCIPAL_SECRET=f0d6d3ce-b72e-430a-972f-025b2cc7279e
+make azure-start
+```
```
nickolaev: Thanks for taking the time to write this detailed description. Looks great!

test/kube_testing/k8s.go
stepan-anokhin: In current approach Azure pulls all required images from docker hub. But this affects only the first deploy as images are getting cached on AKS nodes by docker as usual.

test/integration/const.go
stepan-anokhin: Tests on Azure are a little bit slower than on packet. So I were seeing timeout-failures from time to time. Probably we'll need to scale AKS nodes in future. Currently we use [Standard_B2s](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/sizes-general#b-series).

test/integration/nsmd_test_utils/test_utils.go
stepan-anokhin: Effectively this is yet another increased timeout. The reason is the same as for the `defaultTimeout`. So nothing new here. 

On the other hand setting up memif interface takes surprisingly long time on all platforms. I think we should take a look at the memif creation performance. Denis Tingajkin is working on memif performance tests, as I know.

test/integration/recover_remote_nse_nsm_heal_test.go
stepan-anokhin: Yep. There is definitely a problem that simply has a greater chance to reproduce on Azure.  I've created #961 to track that.

Issue #841, opened 2019-03-13
Add exclude_prefix check to SDK


edwarnicke:
## Overview

This is a sub-issue of #838, see that issue for more context.

The SDK should have a check to ensure that IPAM has not allocated IPs or Routes from the exclude_prefix list in the connection context.  This check should be added to all of the examples in NSM.

nickolaev:
My assumption is the change should happen somewhere along these lines:
https://github.com/networkservicemesh/networkservicemesh/blob/master/sdk/endpoint/composite/ipam.go#L56

edwarnicke:
Yes.  Probably best to write some handy helper functions that could be used in all kinds of IPAM Composables... 

Note, net.IPNet has some helpful functions for testing whether something is in a given Prefix.

edwarnicke:
@rdimitrov How are we doing getting this resolved?

rdimitrov:
@edwarnicke It's WIP as it turned out that the IP allocation algorithm needs to be updated as well.

lobkovilya:
We need more carefully choose prefix that will be split. Candidates could be categorized in three types:
1. Without any intersection with `excludePrefixes`
2. Have some excluded prefixes as a part
3. Fully contained in one of `excludePrefixes`

I suppose Type1 is most preferable for splitting. Type2 is the second priority and Type3 is immediately discarded. In case of absent of Type1 we dial with Type2. And here is the most complicated part - we should carefully check is it possible to find prefix of required size in type2-candidate:

Let's assume that type2-candidate has size of prefix equal to `C`, we trying to extract prefix with size equal `R` we know that `R > C`. Also we have excluded prefixes with sizes `E1`, `E2` ... some of them bigger than R, some smaller. Always can think about prefixes as nodes in tree:
![image](https://user-images.githubusercontent.com/12110261/55904778-d8dfad80-5bfa-11e9-90b3-97898184051a.png)

Excluded prefix eliminates the whole subtree (red color nodes on image):
![image](https://user-images.githubusercontent.com/12110261/55905669-aafb6880-5bfc-11e9-8906-2348df4f3f1d.png)

Easy to understand that candidate has `2^(R-C)` prefixes of required size `R`, in example - `2^2 = 4` prefixes of size `R` in candidate of size `C`. If there are some excluded prefixes with sizes smaller then `R` (`E1` in example) they will exclude `2^(R-E)` prefixes of size `R`, in example - `2^1 = 2`. 

[...]

PR #888, opened 2019-03-29
Take into account the excludePrefix in the SDK

rdimitrov:
## Motivation and Context

Issue #841

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


rdimitrov:
@lobkovilya I was wondering if you can review my changes as based on the integration tests, I'm not sure if it's better to return directly or raise an error and handle it differently when a interesect fault occurs?

lobkovilya:
@rdimitrov I suppose we can pass `excludedPrefixes` even deeper, right to the `ExtractPrefixes()` function. This approach let us affect on algorithm of extracting prefix and not "give up" so fast and return error :) What do you think about it?

controlplane/pkg/prefix_pool/prefixpool.go
```diff
@@ -47,14 +49,15 @@ func NewPrefixPool(prefixes ...string) (PrefixPool, error) {
 	}, nil
 }
 
-func (impl *prefixPool) Extract(connectionId string, family connectioncontext.IpFamily_Family, requests ...*connectioncontext.ExtraPrefixRequest) (srcIP *net.IPNet, dstIP *net.IPNet, requested []string, err error) {
+func (impl *prefixPool) Extract(connectionId string, family connectioncontext.IpFamily_Family, excludedPrefixes []string, requests ...*connectioncontext.ExtraPrefixRequest) (srcIP *net.IPNet, dstIP *net.IPNet, requested []string, err error) {
```
lobkovilya: Probably it's better to create new method
```
ExtractWithExcludedPrefix(connectionId string, family connectioncontext.IpFamily_Family, excludedPrefixes []string, requests ...*connectioncontext.ExtraPrefixRequest) (srcIP *net.IPNet, dstIP *net.IPNet, requested []string, err error)
```
And leave `Extract` as it is:
```
Extract(connectionId string, family connectioncontext.IpFamily_Family, requests ...*connectioncontext.ExtraPrefixRequest) (srcIP *net.IPNet, dstIP *net.IPNet, requested []string, err error) {
    return ExtractWithExcludedPrefix(connectionId, family, []string{}, requests...)
}
```

lobkovilya:
And yeah, test TestExcludePrefixCheck should be changed. For example we could grep logs from icmp-responder and wait for error in logs "IPAM: Allocated IP intersects with excludedPrefix" instead of waiting logs in nsc. Or for that test we can deploy "dump" icmp-responder without excluded prefix check and create one more test to check logic that you have provided to sdk

controlplane/pkg/prefix_pool/prefixpool.go
```diff
@@ -201,13 +205,44 @@ func ExtractPrefixes(prefixes []string, requests ...*connectioncontext.ExtraPref
...
+	}
+
+	if intersect {
+		return fmt.Errorf("IPAM: Allocated IP intersects with excludedPrefix")
+	}
+
+	return nil
+}
+
 func ExtractPrefix(prefixes []string, prefixLen uint32) (string, []string, error) {
```
lobkovilya: This is function where choosing of appropriate prefix happen. So we need to pass excludedPrefixes here, to affect on this algorithm.

There is a cycle over all available prefixes. Firstly it tries to find prefix with length equal to `prefixLen`. In case of absent of such prefix we find a candidate to split in 2 subnet. So please provide additional logic of choosing that candidate - it mustn't be equal to one of excludedPrefix or be included in one of excludedPrefix

lobkovilya:
It looks good, but I assume that we still have situations when we have available prefix, but algorithm returns error. 

Imagine that we call:
```
ExtractPrefixes(string[]{"10.20.0.0/16"}, 30, string[]{"10.20.0.0/24"})
```
Cycle https://github.com/networkservicemesh/networkservicemesh/pull/888/files#diff-32a094c1e510859d935b4e71d8de9103R278 goes down the tree of prefixes and always choose left branch (subnet with last prefix bit equals to 0). Cycle stops when size of `rootCIDRNet` will equal to `prefixLen`. So we easily achieve `10.20.0.0/30` but final check at https://github.com/networkservicemesh/networkservicemesh/pull/888/files#diff-32a094c1e510859d935b4e71d8de9103R301 possibly says that `10.20.0.0/30` intersects with `10.20.0.0/24` and it leads to error of `ExtractPrefixes` function. But obviously we could find subnet of `10.20.0.0/16` with size 30 that is not intersected with `excludedPrefixes`.

I haven't had a chance to check it in running program, but I suspect that provided example is valid. Could you please check it?

rdimitrov:
Appreciate your feedback, @lobkovilya 👍 I created a new test for that case and it does fail, so I'll update it 👍

controlplane/pkg/prefix_pool/prefixpool.go
```diff
@@ -201,13 +205,38 @@ func ExtractPrefixes(prefixes []string, requests ...*connectioncontext.ExtraPref
...
+	}
+
+	if intersect {
+		return fmt.Errorf("IPAM: Allocated IP intersects with excludedPrefix")
+	}
+
+	return nil
+}
+
+func ExtractPrefix(prefixes []string, prefixLen uint32, excludedPrefixes []string) (string, []string, error) {
```
rdimitrov: @lobkovilya Sorry for taking a bit longer to reply. I did try the test you suggested and although currently it is handled as a false negative (stating that it cannot allocate the desired network when in reality it is possible), digging deeper seems that the address allocation algorithm (the one with the subnet splitting) doesn't proceed to check the range from 10.20.1.0 to 10.20.255.255 as it should, but it ends at the 10.20.0.0 one.

controlplane/pkg/prefix_pool/prefixpool.go
```diff
@@ -221,24 +250,29 @@ func ExtractPrefix(prefixes []string, prefixLen uint32) (string, []string, error
...
 			resultPrefixes := append(prefixes[:idx], prefixes[idx+1:]...)
+			/* Verify the allocated addresses does not intersect with one from the excludedPrefix list */
+			err := verifyExcluding(resultPrefix, excludedPrefixes)
+			if err != nil {
+				return "", prefixes, err
+			}
 			// Lets remove from list and return
 			return resultPrefix, resultPrefixes, nil
 		} else {
 			// Update minimal root for split
```
haiodo: Hi, We also need to choose a candidate to split here with verification subnet is fit based on 
verifyExcluding(resultPrefix, excludedPrefixes)

controlplane/pkg/prefix_pool/prefixpool.go
```diff
@@ -221,24 +250,29 @@ func ExtractPrefix(prefixes []string, prefixLen uint32) (string, []string, error
...
 		return "", prefixes, fmt.Errorf("Failed to find room to have prefix len %d at %v", prefixLen, prefixes)
 	}
 
-	resultPrefixRoot := prefixes[max_prefix_idx]
-	right_parts := []string{}
+	resultPrefixRoot := prefixes[maxPrefixIndex]
+	rightParts := []string{}
 
 	_, rootCIDRNet, _ := net.ParseCIDR(resultPrefixRoot)
 	for {
```
haiodo: Here we need to check and choose a direction we need to split based on 
verifyExcluding(resultPrefix, excludedPrefixes) to select left or right subtree, so we will not split just to left one.

controlplane/pkg/prefix_pool/prefixpool.go
```diff
@@ -256,15 +290,18 @@ func ExtractPrefix(prefixes []string, prefixLen uint32) (string, []string, error
...
-	resultPrefixes = append(resultPrefixes, reverse(right_parts)...)
-	resultPrefixes = append(resultPrefixes, prefixes[max_prefix_idx+1:]...)
-
-	// return result
+	var resultPrefixes []string
+	resultPrefixes = append(resultPrefixes, prefixes[:maxPrefixIndex]...)
+	resultPrefixes = append(resultPrefixes, reverse(rightParts)...)
+	resultPrefixes = append(resultPrefixes, prefixes[maxPrefixIndex+1:]...)
+	/* Verify the allocated addresses does not intersect with one from the excludedPrefix list */
+	err := verifyExcluding(rootCIDRNet.String(), excludedPrefixes)
```
haiodo: This check is not required since we will already choose a matched subtree.

nickolaev:
@haiodo can you please confirm the current state if the PR is in line with your vision of it?

.circleci/config.yml
```diff
@@ -307,23 +307,23 @@ workflows:
           requires:
             - packet-deploy
       - integration-tests:
-          name: "packet test basic usecase"
+          name: "packet-test-basic-usecase"
```
nickolaev: Makes sense but please do it in a separate PR.

controlplane/pkg/nsm/nsm.go
```diff
@@ -389,8 +390,8 @@ func (srv *networkServiceManager) isLocalEndpoint(endpoint *registry.NSERegistra
 }
 
 /**
-	ctx - we assume it is big enought to perform connection.
- */
+ctx - we assume it is big enought to perform connection.
```
nickolaev: Please put these in a separate PR.

controlplane/pkg/prefix_pool/prefixpool.go
nickolaev: @rdimitrov, please share your approach here. Why is it better?

controlplane/pkg/prefix_pool/prefixpool_test.go
```diff
@@ -127,12 +140,29 @@ func TestExtract2_ipv6(t *testing.T) {
...
 
+func TestExtractExclude(t *testing.T) {
+	RegisterTestingT(t)
+
+	prefix, _, err := ExtractPrefix([]string{"10.20.0.0/16"}, 30, []string{"10.20.0.0/24"})
+	Expect(err).To(BeNil())
+	Expect(prefix).To(Equal("10.20.128.0/30"))
+}
+
+func TestExtractExcludeIntersect(t *testing.T) {
```
nickolaev: Are we sure we are covering all the needed cases? Maybe we need more tests?

controlplane/pkg/tests/nsmd_remote_test.go
```diff
@@ -215,27 +215,27 @@ func TestNSMDDelayRemoteMechanisms(t *testing.T) {
 
 	type Response struct {
 		nsmResponse *connection.Connection
-		err error
+		err         error
```
nickolaev: The changes in this file are not relevant to the patch.Please consider a separate PR.

dataplane/pkg/common/egress_interface.go
```diff
@@ -36,9 +36,9 @@ type EgressInterface interface {
```
nickolaev: Not relevant. Conseder a seaprate PR.

dataplane/pkg/common/egress_interface_test.go
```diff
@@ -1,4 +1,4 @@
-package  common
```
nickolaev: Not relevant. Please consider a separate PR.

dataplane/vppagent/pkg/converter/acl_converter.go
```diff
@@ -105,7 +105,7 @@ func getIcmp(parsed map[string]string) (*vpp_acl.ACL_Rule_IpRule_Icmp, error) {
 	if err != nil {
```
nickolaev: Separate PR?

dataplane/vppagent/pkg/converter/kernel_connection_converter.go
```diff
@@ -91,11 +91,11 @@ func (c *KernelConnectionConverter) ToDataRequest(rv *configurator.Config, conne
 			Enabled:     true,
```
nickolaev: Separate PR?

dataplane/vppagent/pkg/converter/memif_interface_converter.go
```diff
@@ -20,9 +20,9 @@ import (
 	"github.com/ligato/vpp-agent/api/models/vpp"
```
nickolaev: Separate PR?

dataplane/vppagent/pkg/converter/remote_connection_converter.go
```diff
@@ -82,7 +82,7 @@ func (c *RemoteConnectionConverter) ToDataRequest(rv *configurator.Config, conne
 			Vxlan: &vpp_interfaces.VxlanLink{
```
nickolaev: Separate PR?

dataplane/vppagent/pkg/vppagent/vppagent.go
```diff
@@ -16,15 +16,16 @@ package vppagent
```
nickolaev: Separate PR?

examples/cmd/vppagent-firewall-nse/vppagent.go
```diff
@@ -58,7 +58,7 @@ func (vxc *vppAgentXConnComposite) crossConnecVppInterfaces(ctx context.Context,
 	}
```
nickolaev: Separate PR?

examples/cmd/vppagent-icmp-responder-nse/vppagent.go
```diff
@@ -45,9 +45,9 @@ func (ns *vppagentComposite) CreateVppInterface(ctx context.Context, nseConnecti
 		return err
```
nickolaev: Separate PR?

examples/cmd/vppagent-nsc/vppagent.go
```diff
@@ -70,8 +70,7 @@ func Reset(vppAgentEndpoint string) error {
 	client := configurator.NewConfiguratorClient(conn)
```
nickolaev: Separate PR?

pkg/tools/tools.go
```diff
@@ -126,7 +126,7 @@ func WaitForPortAvailable(ctx context.Context, protoType string, registryAddress
 		default:
```
nickolaev: Separate PR?

scripts/aws/create-kubernetes-cluster.go
```diff
@@ -22,15 +22,15 @@ import (
 )
```
nickolaev: Separate PR?

scripts/aws/cubernetes-cluster.go
```diff
@@ -5,26 +5,26 @@ import (
 	"os"
```
nickolaev: Separate PR?

scripts/aws/destroy-kubernetes-cluster.go
```diff
@@ -45,24 +45,26 @@ func DeleteEksRole(iamClient *iam.IAM, eksRoleName *string) {
```
nickolaev: Separate PR?

scripts/prepare-circle-integration-tests.sh
```diff
@@ -17,7 +17,7 @@
 
 set -xe
 
-KUBECTL_VERSION=v1.13.3
+KUBECTL_VERSION=v1.13.4
```
nickolaev: That does not belong here.

test/kube_testing/pods/nsmd.go
```diff
@@ -118,7 +118,7 @@ func NSMgrPodWithConfig(name string, node *v1.Node, config *NSMgrPodConfig) *v1.
 					Image:           "networkservicemesh/nsmdp",
```
nickolaev: Separate PR.

test/kube_testing/pods/vppagent_dataplane.go
```diff
@@ -89,7 +89,7 @@ func createVPPDataplanePod(name string, node *v1.Node, liveness, readiness *v1.P
 					},
```
nickolaev: Separate PR.

lobkovilya:
Seems like current implementation still has "false negative" behavior. 

Algorithm has 2 phases: election of appropriate candidate for splitting and splitting itself. Now election phase uses so called "best-fit" strategy:
```go
for idx, prefix := range prefixes {
    ...
    if uint32(parentLen) < prefixLen && (parentLen > maxPrefix || maxPrefix == 0) {
	maxPrefix = parentLen
	maxPrefixIndex = idx
    }
}
```
Since there is no check whether candidate intersects with `excludePrefixes` or not, we can't unambiguously say that it is possible to extract prefixes from it. In other words another candidate could be "wider" (has shorter prefix) but without intersections with `excludePrefixes`. 

But simple reelection of candidate after Phase 2 ("splitting") failed will not help. Turns out splitting is harder than we thought before. Checking which subtree intersects with `excludePrefixes` and going down to opposite subtree will not work in case of `excludePrefixes` consist of several non-overlapping prefixes with different sizes. Imagine that candidate has prefix `/16`, `excludePrefixes` contains couple tiny `/30` subnets. One of subnet included to left `/17` subtree and other to right. Algorithm will go down to the left subtree and may fail to find appropriate prefix while right subtree could contain required prefix.

I leave some draft description of possible algorithm in related issue, hope it would be helpful.

controlplane/pkg/prefix_pool/prefixpool.go
```diff
@@ -47,6 +51,121 @@ func NewPrefixPool(prefixes ...string) (PrefixPool, error) {
...
+
+		/* 1. Check if each excluded entry overlaps with the available prefix */
+		for _, prefix := range copyPrefixes {
+			intersecting := false
+			excludedIsBigger := false
+			_, subnetPrefix, _ := net.ParseCIDR(prefix)
+			intersecting, excludedIsBigger = intersect(subnetExclude, subnetPrefix)
+			/* 1.1. If intersecting, check which one is bigger */
+			if intersecting {
+				/* 1.1.1. If excluded is bigger, we remove the original entry */
```
lobkovilya: @rdimitrov Hm, probably I am mistaken, but if we simply remove original entry in that case, then during releasing we increase the original size of prefix pool. I mean the following test most likely will fail:
```go
pool, err := NewPrefixPool("10.20.0.0/16", "2.20.0.0/16")
Expect(err).To(BeNil())
excludedPrefix := []string{"2.20.0.0/8"}

excluded, err := pool.ExcludePrefixes(excludedPrefix)

Expect(err).To(BeNil())
Expect(pool.GetPrefixes()).To(Equal([]string{"10.20.0.0/16"}))

err = pool.ReleaseExcludedPrefixes(excluded)
Expect(err).To(BeNil())
Expect(pool.GetPrefixes()).To(Equal([]string{"10.20.0.0/16", "2.20.0.0/16"}))
```

sdk/endpoint/ipam.go
```diff
@@ -51,12 +51,24 @@ func (ice *IpamEndpoint) Request(ctx context.Context, request *networkservice.Ne
 		return nil, err
 	}
 
+	/* Exclude the prefixes from the pool of available prefixes */
+	excludedPrefixes, err := ice.prefixPool.ExcludePrefixes(request.Connection.Context.ExcludedPrefixes)
```
lobkovilya: May be it is better to save copy of prefixPool before excluding prefixes

sdk/endpoint/ipam.go
```diff
@@ -51,12 +51,24 @@ func (ice *IpamEndpoint) Request(ctx context.Context, request *networkservice.Ne
...
+	}
+
 	//TODO: We need to somehow support IPv6.
 	srcIP, dstIP, prefixes, err := ice.prefixPool.Extract(request.Connection.Id, connectioncontext.IpFamily_IPV4, request.Connection.Context.ExtraPrefixRequest...)
 	if err != nil {
 		return nil, err
 	}
 
+	/* Release the actual prefixes that were excluded during IPAM */
+	err = ice.prefixPool.ReleaseExcludedPrefixes(excludedPrefixes)
```
lobkovilya: And restore that copy right here? 
Then we can avoid implementing `ReleaseExcludePrefixes` method. What do you think about it?

rdimitrov: Make sense, but if we restore the original state of `prefixes` at this point - this happens right after we do the IPAM using the altered copy with the excluded prefixes. If successful, this results in extracting prefixes from it, so if we do that, in the end, we should have the troubles to take into account that change and translate it back to the original, unaltered copy.

controlplane/pkg/prefix_pool/prefixpool.go
rdimitrov: Thanks, @lobkovilya! 👍 I've updated the tests with that use case and added a fix for it.

rdimitrov: By the way, I found this one to be really useful for verifying what to expect - https://www.techzoom.net/Tools/IPAddressCalculator 👍

rdimitrov:
@nickolaev I assume the "requested changes" part was already covered

sdk/endpoint/ipam.go
lobkovilya: Yep, you're absolutely right!

Issue #884, opened 2019-03-28
TestDataplaneHealRemote test is failed because NSM:(5.1-XXX) Failed to select mechanism. No matched mechanisms found.


haiodo:
## Expected Behavior

Test should pass.
## Current Behavior

Test is failed with message: 
NSM:(5.1-4B049BF9) Failed to select mechanism. No matched mechanisms found...

Not able to reproduce locally.

## Steps to Reproduce

1. Step 1
2. Step 2
3. You get the idea...

## Context

* Kubernetes Version:
* etc.

## Failure Logs

```
=== RUN   TestDataplaneHealRemote
time="2019-03-28T18:27:20Z" level=info msg="Cleanup done: 639.05209ms"
time="2019-03-28T18:27:20Z" level=info msg="Checking node: ci-22049-master"
time="2019-03-28T18:27:20Z" level=info msg="Checking node: ci-22049-worker1"
time="2019-03-28T18:27:33Z" level=info msg="Started NSMgr/Dataplane: 12.517439302s on node ci-22049-master"
time="2019-03-28T18:27:44Z" level=info msg="Started NSMgr/Dataplane: 11.060748293s on node ci-22049-worker1"
time="2019-03-28T18:27:44Z" level=info msg="Starting ICMP Responder NSE on node: ci-22049-worker1"
time="2019-03-28T18:27:46Z" level=info msg="ICMP Responder icmp-responder-nse-1 started done: 2.190776149s"
time="2019-03-28T18:27:46Z" level=info msg="Starting NSC nsc-1 on node: ci-22049-master"
time="2019-03-28T18:27:51Z" level=info msg="NSC started done: 5.000566803s"
time="2019-03-28T18:27:51Z" level=info msg="Executing: [ip addr] in pod nsc-1:alpine-img"
time="2019-03-28T18:27:52Z" level=info msg="NSC IP status Ok"
time="2019-03-28T18:27:52Z" level=info msg="Executing: [ip route] in pod nsc-1:alpine-img"
time="2019-03-28T18:27:53Z" level=info msg="NSC Route status, Ok"
time="2019-03-28T18:27:53Z" level=info msg="Executing: [ping 10.20.1.2 -c 5] in pod nsc-1:alpine-img"
[...]

PR #896, opened 2019-04-03
Wait for dataplane mechanisms before using it

artem-belov:
Wait for dataplane mechanisms before using it
nsmd starts working with dataplane right after registration. I cause errors with missing mechanisms (issue #884)

## Description
Waiting for mechanisms configuration added.
Component test added.

## Motivation and Context
Issue #884 

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/controlplane/pkg/nsmd/dataplanes.go
+++ b/controlplane/pkg/nsmd/dataplanes.go
@@ -89,6 +89,7 @@ func dataplaneMonitor(model model.Model, dataplaneName string) {
 		// TODO: this is not good -- direct model changes
 		dataplane.RemoteMechanisms = updates.RemoteMechanisms
 		dataplane.LocalMechanisms = updates.LocalMechanisms
+		dataplane.RemoteConfigured = true
 	}
 }
 
--- a/controlplane/pkg/nsmd/dataplanes.go
+++ b/controlplane/pkg/nsmd/dataplanes.go
@@ -89,7 +89,7 @@ func dataplaneMonitor(model model.Model, dataplaneName string) {
 		// TODO: this is not good -- direct model changes
 		dataplane.RemoteMechanisms = updates.RemoteMechanisms
 		dataplane.LocalMechanisms = updates.LocalMechanisms
-		dataplane.RemoteConfigured = true
+		dataplane.MechanismsConfigured = true
 	}
 }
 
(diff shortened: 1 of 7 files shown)


controlplane/pkg/nsmd/serviceregistry.go
```diff
@@ -241,6 +241,17 @@ func (impl *nsmdServiceRegistry) WaitForDataplaneAvailable(model model.Model, ti
 	st := time.Now()
 	for ; true; <-time.After(100 * time.Millisecond) {
 		if dp, _ := model.SelectDataplane(); dp != nil {
+
+			// Wait for mechanisms configuration
+			if !dp.RemoteConfigured {
```
lobkovilya: Is it possible to check `dp.RemoteMechanisms == nil` or `len(dp.RemoteMechanisms) == 0` instead of creating new flag `RemoteConfigured`?

artem-belov: Dataplane may not have RemoteMechanisms after Mechanisms Update

controlplane/pkg/nsmd/dataplanes.go
```diff
@@ -89,6 +89,7 @@ func dataplaneMonitor(model model.Model, dataplaneName string) {
 		// TODO: this is not good -- direct model changes
 		dataplane.RemoteMechanisms = updates.RemoteMechanisms
 		dataplane.LocalMechanisms = updates.LocalMechanisms
+		dataplane.RemoteConfigured = true
```
nickolaev: Hmm what `RemoteConfigured` means here?

nickolaev: Maybe should change it to RemoteMechanismChnaged, and clear the flag in NSM?

artem-belov: We need that flag only to wait while nsmd will receive package with dataplane mechanisms parameters first time. So we don't need to clear the flag.
For example to fix situation, when we are trying to heal state with new registered dataplane, but we haven't received package with information about mechanisms yet.  Nsm will emit updateMechanism and return error because it cannot select mechanism.

artem-belov: So RemoteConfigured means that we received at least one package with mechanisms and nsm can register connection

nickolaev: OK, but those mechanisms can be "local" only, right?

artem-belov: Yes

artem-belov: I will rename it to MechanismsConfigured.

Issue #886, opened 2019-03-29
GKE VXLan tunnel is not working


haiodo:
## Expected Behavior

VXLAN_TUNNEL should work

## Current Behavior

VXLAN_TUNNEL is not working inside GKE. Working well with VETH mode forced inside Vagrant.

## Steps to Reproduce

Execute any of remote integration tests inside GKE.

## Failure Logs

Dataplane logs for both nodes:

[dp1.log](https://github.com/networkservicemesh/networkservicemesh/files/3022470/dp1.log)
[dp2.log](https://github.com/networkservicemesh/networkservicemesh/files/3022471/dp2.log)

SRC dataplane:
```
root@nsmd-dataplane-gke-dev-default-pool-8cdecbc6-4gw9:~# vppctl show int
              Name               Idx    State  MTU (L3/IP4/IP6/MPLS)     Counter          Count     
host-SRC-1-veth                   3      up          9000/0/0/0     rx packets                   433
                                                                    rx bytes                   18798
                                                                    drops                        433
host-eth0                         1      up          9000/0/0/0     rx packets                    26
                                                                    rx bytes                    1624
                                                                    tx packets                     1
                                                                    tx bytes                      42
                                                                    drops                         25
                                                                    ip4                            3
                                                                    ip6                            7
local0                            0     down          0/0/0/0       
vxlan_tunnel0                     2      up           0/0/0/0       tx packets                   433
                                                                    tx bytes                   34386
```

NSC:
```
kubectl exec -ti nsc-1 ip addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN qlen 1000
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
    inet 127.0.0.1/8 scope host lo
       valid_lft forever preferred_lft forever
    inet6 ::1/128 scope host 
       valid_lft forever preferred_lft forever
3: eth0@if112: <BROADCAST,MULTICAST,UP,LOWER_UP,M-DOWN> mtu 1460 qdisc noqueue state UP 
    link/ether 0a:58:0a:28:00:6d brd ff:ff:ff:ff:ff:ff
    inet 10.40.0.109/24 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::1866:2fff:fe71:b431/64 scope link 
       valid_lft forever preferred_lft forever
[...]

PR #899, opened 2019-04-03
Fix for #886

haiodo:

If Dataplane pods had different nets we need to have routes between them inside vpp.

## Description
Adds proper routes between data planes in case they have different network prefixes.

## Motivation and Context
To support this kind of scenarios.

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/controlplane/pkg/model/model.go
+++ b/controlplane/pkg/model/model.go
@@ -189,7 +189,10 @@ func (i *impl) GetAllClientConnections() []*ClientConnection {
 
 func (i *impl) UpdateClientConnection(clientConnection *ClientConnection) {
 	i.Lock()
-	i.clientConnections[clientConnection.ConnectionId] = clientConnection
+	_, ok := i.clientConnections[clientConnection.ConnectionId]
+	if ok {
+		i.clientConnections[clientConnection.ConnectionId] = clientConnection
+	}
 	i.Unlock()
 
 	for _, listener := range i.listeners {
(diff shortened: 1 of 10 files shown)


edwarnicke:
Question... why are we changing the remoteConnection converters etc?  We only need to fix the routes  here for the vppagent-dataplane's Pod interface...

edwarnicke:
Also not clear why we are altering the NSEs around this issue...

examples/cmd/vppagent-firewall-nse/vppagent.go
```diff
@@ -49,6 +49,7 @@ func (vxc *vppAgentXConnComposite) crossConnecVppInterfaces(ctx context.Context,
 
 	conversionParameters := &converter.CrossConnectConversionParameters{
 		BaseDir: baseDir,
+		Routes: &converter.ExtraRoutesParameters{},
```
edwarnicke: Is this an unrelated fix?

controlplane/pkg/model/model.go
```diff
@@ -204,6 +207,7 @@ func (i *impl) DeleteClientConnection(connectionId string) {
 		i.Unlock()
 		return
 	}
+	clientConnection.ConnectionState = ClientConnection_Closed
```
edwarnicke: Is this an unrelated fix?

examples/cmd/vppagent-firewall-nse/vppagent.go
haiodo: Thanks. Pushed fix.

controlplane/pkg/model/model.go
haiodo: This fixes concurrency issue, some tests failed on build.

haiodo: Fix for concurrency issue and moved marking of connection is closed into this method.

PR #900, opened 2019-04-03
Run supervisord as process with PID 1

### LINKED ISSUE #894: vppagent-firewall-nse doesn't unregister after deletion
## Expected Behavior

After deleting the deployment k8s/conf/vppagent-firewall-nse.yaml, it shouldn't be seen with 

```
kubectl describe networkserviceendpoints.networkservicemesh.io
```

## Current Behavior

The endpoint is still listed as a valid endpoint in the registry

## Steps to Reproduce

1. kubectl apply -f k8s/conf/secure-intranet-connectivity.yaml
2. kubectl apply -f k8s/conf/vppagent-firewall-nse.yaml
3. kubectl describe networkserviceendpoints.networkservicemesh.io
4. kubectl delete -f k8s/conf/vppagent-firewall-nse.yaml
5. kubectl describe networkserviceendpoints.networkservicemesh.io

## Context

I've used a vagrant deployment.

matrohon:
Fixes #894

## Description

Supervisord needs to be run as PID 1 to receive kill signals from kubelet and forward them to the processes it manages. Once SIGTERM received, NSM endpoint will properly unregister themselves.

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/docker/Dockerfile.vppagent-firewall-nse
+++ b/docker/Dockerfile.vppagent-firewall-nse
@@ -18,3 +18,6 @@ COPY --from=build /go/bin/vppagent-firewall-nse /bin/vppagent-firewall-nse
 RUN rm /opt/vpp-agent/dev/etcd.conf /opt/vpp-agent/dev/kafka.conf; echo 'Endpoint: "0.0.0.0:9112"' > /opt/vpp-agent/dev/grpc.conf; echo "disabled: true" > /opt/vpp-agent/dev/linux-plugin.conf
 COPY dataplane/vppagent/conf/vpp/startup.conf /etc/vpp/vpp.conf
 COPY examples/conf/vppagent-firewall-nse/supervisord.conf /etc/supervisord/supervisord.conf
+
+RUN rm -f /dev/shm/db /dev/shm/global_vm /dev/shm/vpe-api
+ENTRYPOINT ["/usr/bin/supervisord","-c","/etc/supervisord/supervisord.conf"]
--- a/docker/Dockerfile.vppagent-icmp-responder-nse
+++ b/docker/Dockerfile.vppagent-icmp-responder-nse
@@ -18,3 +18,6 @@ COPY --from=build /go/bin/vppagent-icmp-responder-nse /bin/vppagent-icmp-respond
 RUN rm /opt/vpp-agent/dev/etcd.conf /opt/vpp-agent/dev/kafka.conf; echo 'Endpoint: "0.0.0.0:9112"' > /opt/vpp-agent/dev/grpc.conf; echo "disabled: true" > /opt/vpp-agent/dev/linux-plugin.conf
 COPY dataplane/vppagent/conf/vpp/startup.conf /etc/vpp/vpp.conf
 COPY examples/conf/vppagent-icmp-responder-nse/supervisord.conf /etc/supervisord/supervisord.conf
+
+RUN rm -f /dev/shm/db /dev/shm/global_vm /dev/shm/vpe-api
+ENTRYPOINT ["/usr/bin/supervisord","-c","/etc/supervisord/supervisord.conf"]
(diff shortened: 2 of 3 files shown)


docker/Dockerfile.vppagent-firewall-nse
```diff
@@ -18,3 +18,6 @@ COPY --from=build /go/bin/vppagent-firewall-nse /bin/vppagent-firewall-nse
 RUN rm /opt/vpp-agent/dev/etcd.conf /opt/vpp-agent/dev/kafka.conf; echo 'Endpoint: "0.0.0.0:9112"' > /opt/vpp-agent/dev/grpc.conf; echo "disabled: true" > /opt/vpp-agent/dev/linux-plugin.conf
 COPY dataplane/vppagent/conf/vpp/startup.conf /etc/vpp/vpp.conf
 COPY examples/conf/vppagent-firewall-nse/supervisord.conf /etc/supervisord/supervisord.conf
+
+RUN rm -f /dev/shm/db /dev/shm/global_vm /dev/shm/vpe-api
```
nickolaev: Can you explain why we need this step? Is it related to the issue?

matrohon: I've mimic the [CMD](https://github.com/ligato/vpp-agent/blob/485e388d53a46f6789fe0b0fca100f184cfbaee1/docker/dev/Dockerfile#L125) of the original image, which does this rm in addition to running supervisord

matrohon: see https://github.com/moby/moby/issues/5147

edwarnicke:
Nice catch!  The fact that it takes a long time to shut down the containers for some of our Pods has been driving me nuts, does this fix that?

edwarnicke:
Is the main difference between this and the CMD in the [base vpp-agent container](https://github.com/ligato/vpp-agent/blob/master/docker/prod/Dockerfile#L67) that you in the base image:

```
CMD rm -f /dev/shm/db /dev/shm/global_vm /dev/shm/vpe-api && \
    /usr/bin/supervisord -c /etc/supervisord/supervisord.conf
```

Results in the ```rm``` picking up pid 1?  

Is there a reason you picked entrypoint here instead of a new use of CMD ?  CMD has the handy feature that it is a tiny bit less opinionated than Endpoint in ways that sometimes make it easier to do some debugging activities.

docker/Dockerfile.vppagent-icmp-responder-nse
```diff
@@ -18,3 +18,6 @@ COPY --from=build /go/bin/vppagent-icmp-responder-nse /bin/vppagent-icmp-respond
 RUN rm /opt/vpp-agent/dev/etcd.conf /opt/vpp-agent/dev/kafka.conf; echo 'Endpoint: "0.0.0.0:9112"' > /opt/vpp-agent/dev/grpc.conf; echo "disabled: true" > /opt/vpp-agent/dev/linux-plugin.conf
 COPY dataplane/vppagent/conf/vpp/startup.conf /etc/vpp/vpp.conf
 COPY examples/conf/vppagent-icmp-responder-nse/supervisord.conf /etc/supervisord/supervisord.conf
+
+RUN rm -f /dev/shm/db /dev/shm/global_vm /dev/shm/vpe-api
+ENTRYPOINT ["/usr/bin/supervisord","-c","/etc/supervisord/supervisord.conf"]
```
edwarnicke: Could we make this CMD instead of entrypoint?

matrohon:
with CMD, the commands will run in a shell which will have PID 1 and receive signals from kubelet/docker. See Entrypoint  section in [docker documentation](https://docs.docker.com/engine/reference/builder/#usage).

To my knowlegde, the only way to have supervisord managing signals is to run it as an ENTRYPOINT.

I agree this is less handy... as an alternative, we can run supervisord and the firewall in two distinct containers, in the same pod. The firewall binary will be run as an Entrypoint, receive SIGTERM and send its unregiter message. It is working, I have a dedicated [branch](https://github.com/matrohon/networkservicemesh/tree/bug%23894-firewall-in-container) for that.

docker/Dockerfile.vppagent-icmp-responder-nse
matrohon: if we want supervisord to receive signals, this is the simplest way

matrohon:
We can also still overwrite the ENTRYPOINT for debugging purposes

PR #917, opened 2019-04-08
Initial GKE enabled build

haiodo:

GKE enabled building & testing.

## Description
GKE enabled building & testing.

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -109,6 +109,12 @@ jobs:
         type: string
       test_tags:
         type: string
+      gke_compute_zone:
+        type: string
+        default: "us-west2-a"
+      gke_project_id:
+        type: string
+        default: "ci-management"
     working_directory: /go/src/github.com/networkservicemesh/networkservicemesh
     docker:
       - image: circleci/golang
@@ -131,17 +137,19 @@ jobs:
             sudo chown -R circleci:circleci /cncf/
       - restore_cache:
           key: terraform-{{.Environment.CIRCLE_WORKFLOW_ID}}
-      - run: cp -a /home/circleci/project/terraform/terraform/* ./scripts/terraform/
-      - run: ssh-add /home/circleci/project/terraform/sshkey
+      - run: cp -a /home/circleci/project/terraform/terraform/* ./scripts/terraform/ || true
+      - run: ssh-add /home/circleci/project/terraform/sshkey || true
       - restore_cache:
           key: cncf-data-<< parameters.cloud_provider >>-<< parameters.cluster_id >>-{{.Environment.CIRCLE_WORKFLOW_ID}}
-      - run: cp /home/circleci/project/terraform/sshkey* /cncf/data
+      - run: cp /home/circleci/project/terraform/sshkey* /cncf/data || true
       - run:
           name: Downloading go deps
           command: |
             export COMMIT="${CIRCLE_SHA1:8:8}"
             export CONTAINER_TAG="${COMMIT}"
             # export CONTAINER_FORCE_PULL="true"
+            ./scripts/install-kubectl.sh
+            cat /home/circleci/project/data/kubeconfig
             ./scripts/prepare-circle-integration-tests.sh
             mkdir -p ~/junit/
             ./scripts/go-mod-download.sh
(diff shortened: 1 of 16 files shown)


nickolaev:
WiP I guess. Yet please align the test targets with the parametrised 9nes we have for packet. Please check master for reference.

.circleci/config.yml
```diff
@@ -228,105 +228,343 @@ jobs:
...
+          no_output_timeout: 40m
+      - run:
+          name: Running integration tests - basic
+          command: |
+            export COMMIT="${CIRCLE_SHA1:8:8}"
+            export CONTAINER_TAG="${COMMIT}"
+            export CONTAINER_REPO=gcr.io/$(gcloud config get-value project)
+            export KUBECONFIG=~/data/config
+            export GO111MODULE="on"
+            gotestsum --junitfile ~/junit/integration-tests-basic.xml -f standard-verbose ./test/... -timeout 30m -failfast -tags="azure basic"
```
nickolaev: Why `azure` here?

.circleci/config.yml
```diff
@@ -170,12 +172,19 @@ jobs:
...
+            if [ "x<< parameters.cloud_provider >>" == "xgke" ]; then
+              .circleci/cleanup-gke.sh "$GCLOUD_SERVICE_KEY" "${CIRCLE_SHA1:8:8}" "${GOOGLE_PROJECT_ID}" "${GOOGLE_COMPUTE_ZONE}" "dev-testing-${CIRCLE_SHA1:8:8}-<< parameters.cluster_id >>"
+            else
+              .circleci/cleanup-<< parameters.cloud_provider >>.sh
+            fi
     environment:
       PACKET_CLUSTER_ID: "<< parameters.cluster_id >>"
       KUBECONFIG: /home/circleci/project/data/kubeconfig
       GO111MODULE: "on"
+      GOOGLE_COMPUTE_ZONE: "europe-west3-a"
```
nickolaev: Can you make GKE specifics parameters with default values too?

.circleci/config.yml
```diff
@@ -330,3 +384,61 @@ workflows:
           filters:
             branches:
               only: master
+      - gke-create-cluster:
+          cluster_id: "1"
+          name: "gke-cluster1"
+          requires:
+            - build-nsmd
```
nickolaev: Can we leverage GKE build?

.circleci/cleanup-gke.sh
```diff
@@ -0,0 +1,12 @@
...
+export GCLOUD_SERVICE_KEY=$1
+export COMMIT=$2
+export GOOGLE_PROJECT_ID=$3
+export GOOGLE_COMPUTE_ZONE=$4
+export GKE_CLUSTER_NAME=$5
+
+export CONTAINER_TAG="${COMMIT}"
+
+.circleci/init-gke.sh "$GCLOUD_SERVICE_KEY" "$COMMIT" "$GOOGLE_PROJECT_ID" "$GOOGLE_COMPUTE_ZONE" "$GKE_CLUSTER_NAME"
+make gke-destroy
```
nickolaev: Shall we keep the cluster-specific files in their own subfolders, like `.circleci/gke` for example? We can do this for packet too.

haiodo: I think this would be good idea.

.circleci/config.yml
nickolaev: So you can do smething like:
```yaml
  integration-tests:
    parameters:
      cloud_provider:
        type: string
      cluster_id:
        type: string
      test_tags:
        type: string
      gke_compute_zone:
        type: string
        default: "europe-west3-a"
      gke_project_id:
        type: string
        default: "ci-management"
....
    environment:
...
	      GOOGLE_COMPUTE_ZONE:  "<< parameters. gke_compute_zone >>"
             GOOGLE_PROJECT_ID: "<< parameters. gke_project_id >>"
```

haiodo: Thanks, will do.

.circleci/config.yml
haiodo: We can, but it was slower in my experiments and require few more tricks and configurations on GKE, since I've reached few times some quotas.

.circleci/config.yml
```diff
@@ -131,17 +137,19 @@ jobs:
             sudo chown -R circleci:circleci /cncf/
       - restore_cache:
           key: terraform-{{.Environment.CIRCLE_WORKFLOW_ID}}
-      - run: cp -a /home/circleci/project/terraform/terraform/* ./scripts/terraform/
-      - run: ssh-add /home/circleci/project/terraform/sshkey
+      - run: cp -a /home/circleci/project/terraform/terraform/* ./scripts/terraform/ || true
+      - run: ssh-add /home/circleci/project/terraform/sshkey || true
```
nickolaev: Interesting why we need the `|| true` here?

.mk/k8s.mk
```diff
@@ -54,6 +53,8 @@ include .mk/null.mk
 
 include .mk/docker.mk
 
+include .mk/gke.mk
+
```
nickolaev: Why we need to move it after docker?

.circleci/config.yml
haiodo: There is no terraform stuff in GKE case, alternative is to have if [] or move this logic into packet, script, probable it could be better idea to do.

.mk/k8s.mk
haiodo: To be able do gke-push, not used right now on CI, to reuse all defined containers from make file.

.circleci/config.yml
nickolaev: We can have a command. I'll optimise this after the merge. It is fine for now.

PR #960, opened 2019-04-16
Add sorting for prefixes so the unit test results are more consistent

rdimitrov:

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


controlplane/pkg/prefix_pool/prefixpool.go
```diff
@@ -34,6 +35,8 @@ type prefixPool struct {
 }
 
 func (impl *prefixPool) GetPrefixes() []string {
+	/* Sort the prefixes, so their order is consistent during unit testing */
```
haiodo: Would it be better to do on split to store sorted list of prefixes? As I remember original algorithm was created a list of sorted prefixes by design?

rdimitrov: Thanks, I've moved the sorting to be applied during the release phase 👍

PR #1002, opened 2019-04-25
Add arp entries to vppagent transaction

artem-belov:
Configure arp entries for vpp

Signed-off-by: Artem Belov <belov3004@gmail.com>

## Description
Add system arp entries to vppagent transaction

## Motivation and Context
AWS needs configured arp entries to reach gateway

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/dataplane/pkg/common/egress_interface.go
+++ b/dataplane/pkg/common/egress_interface.go
@@ -25,6 +25,12 @@ import (
 	"strings"
 )
 
+type ARPEntry struct {
+	Interface    	string
+	IpAddress       string
+	PhysAddress     string
+}
+
 type EgressInterface interface {
 	SrcIPNet() *net.IPNet
 	DefaultGateway() *net.IP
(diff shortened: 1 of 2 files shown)


edwarnicke:
@nickolaev Do you want to hold for that nit?  Or go ahead and have @artem-belov push a fix for it subsequently?

dataplane/pkg/common/egress_interface.go
```diff
@@ -105,6 +150,11 @@ func NewEgressInterface(srcIp net.IP) (EgressInterface, error) {
 		return nil, err
 	}
 
+	arpEntries, err := getArpEntries()
+	if err != nil {
```
haiodo: @artem-belov we should check if there is no file, we should not fail.

haiodo:
@nickolaev Yes, we already have parsing of proc/net/route inside it to detect default gateway, I suppose this interface is intended to provide all necessary information base on host system, to be able to access all stuff we need. Please let us know if you would like to move some parts to another package/modules. Thanks.

dataplane/pkg/common/egress_interface.go
artem-belov: Done

nickolaev:
All looks good. Merging.

PR #1013, opened 2019-04-29
Common NSE image

Bolodya1997:
## Description

1. Adds single docker image for the testing NSEs.
2. Adds *--dirty* flag to **icmp-responder-nse** application.

## Motivation and Context

Now there are a lot of docker images used in the CI testing. Every new container is being added as a new docker image. So it takes more time for building all these images every time we run CI tests.

The possible solution is using single docker image with multiple applications inside. Application can be selected with the command set in the container description file.

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


examples/cmd/nse/nse_run.sh
```diff
@@ -0,0 +1,33 @@
...
+trap sighup HUP
+trap sigint INT
+trap sigterm TERM
+trap sigquit QUIT
+
+echo Starting NSE: "${NSE_IMAGE}"
+"/bin/${NSE_IMAGE}" &
+
+NSE=$!
+wait "${NSE}"
```
Bolodya1997: > Why 2 times `wait` here?

When the container stops, the shell process receives the signal, breaks wait execution, sends the received signal to the NSE process with the kill and then it should wait for the NSE process to complete all shutdown hooks.
If we omit this second wait, the NSE process will also receive the signal but the container would be destroyed before the shutdown hooks got executed.

examples/cmd/nse/dirty-nse/nse.go
```diff
@@ -0,0 +1,27 @@
+package main
+
+import (
+	"time"
+
+	"github.com/networkservicemesh/networkservicemesh/sdk/endpoint"
+	"github.com/sirupsen/logrus"
+)
+
+func main() {
```
Bolodya1997: > I am not sure why is this needed?
> Also, please add signal handling.

In this context **dirty** means that NSE doesn't perform any cleanup routines. There are a lot of cases when the process can suddenly stop and never continue the execution, so it is to simulate such cases.
So I believe there is no need to add some signal handling in here. The process is broken and so it should be killed by the signal received.

haiodo:
@Bolodya1997 I would prefer name 'test-nse' to not confuse about it is some kind of generic NSE.

docs/guide-debug.md
```diff
@@ -84,7 +84,7 @@ Another examples are:
 
 * `make docker-debug-nsmd` - debug *controlplane/cmd/nsmd*
 * `make docker-debug-nsc` - debug *examples/cmd/nsc*
-* `make docker-debug-icmp-responder-nse` - debug *examples/cmd/icmp-responder-nse*
+* `make docker-debug-icmp-responder-nse` - debug *examples/cmd/nse/icmp-responder-nse*
```
haiodo: Is it correct?

Bolodya1997: Actually there was an issue:
    make docker-icmp-responder-nse**-debug**
Instead of:
    make docker**-debug**-icmp-responder-nse
And so for the other examples.

examples/cmd/nse/nse_run.sh
Bolodya1997: Outdated

examples/cmd/nse/dirty-nse/nse.go
Bolodya1997: Outdated

nickolaev:
I am still not sure about names. We have `nsc` which is not great name of the container because we use it in various sccenarios. Now we're creating `nse` which to me makes the problem worse. For example the existing `icmp-responder` is used as a sidecar container in `vpn-gateway` pod.
I agree we should find better names here, but `nse` is too generic IMHO.

The other thing is using env variables vs command line arguments, what if we just have a singoe binary which can im-lement different functionalities just by passing different args to it? Why we should keep all these shell scripts, env variables etc?

edwarnicke:
@nickolaev I'm not a huge fan of a uni-binary.  I can see the logic in 'build everything once and then choose what you are going to execute' as an approach.

I totally agree on the naming, but I think it also falls into a 'fix things in easy stages'.  Do you think that we could do the naming subsequently?

Bolodya1997:
@nickolaev shell script and env variables have been removed in the last commits. The current solution is to pass startup command to the container - this could be used as for single complex binary with multiple args, so for multiple simple binaries.

edwarnicke:
@Bolodya1997 Would it make sense to collapse the build of various NSE containers into a single CircleCI job?

nickolaev:
OK using `--dirty` makes sense. @edwarnicke please check the prpose dcode now, it is literally a slight divergence in the behavior when you pass that flag. It is not about having various functionalities in the same binary, e.g. `busybox` and the likes.

I will approve the code changes as they are now.

Bolodya1997:
@edwarnicke It does make sense. I can make the subsequent PR to merge all existing NSEs to the **test-nse** image.

Issue #811, opened 2019-03-06
Check for IPNeighbor without HW address in vppagent-dataplane and return error


edwarnicke:
## Expected Behavior

#810 fixes an issue where a [IPNeighbor](https://github.com/networkservicemesh/networkservicemesh/blob/master/controlplane/pkg/apis/connectioncontext/connectioncontext.proto#L24) without a HW Address was being returned from an NSE.

The vppagent-dataplane needs to properly detect and return an error should this happen in the future.
I suggest writing an [IsValid](https://github.com/networkservicemesh/networkservicemesh/blob/master/controlplane/pkg/apis/connectioncontext/helpers.go) for ConnectionContext, and ensuring that's being called in vppagent-dataplane.

A way to test this would be to imitate the issue that revealed #810, and feed it an IPNeighbor with an ip of ::1 and no HWAddress.

## Current Behavior

Currently if an IPNeighbor does not have a HW Address, dataplane provisioning silently fails.

## Steps to Reproduce

1. Step 1
2. Step 2
3. You get the idea...

## Context

* Kubernetes Version:
* etc.

edwarnicke:
@stepan-anokhin Is this bug resolved?  If so, by which PR?

stepan-anokhin:
@edwarnicke it's still pending. I hadn't had a chance to take a look yet.

denis-tingajkin:
In progress

denis-tingajkin:
Seems to me this can be closed

PR #1019, opened 2019-04-29
Check for IPNeighbor without HW address in vppagent-dataplane and return error

denis-tingajkin:

## Description

https://github.com/networkservicemesh/networkservicemesh/issues/811

## How Has This Been Tested?

- [X] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/controlplane/pkg/apis/connectioncontext/helpers.go
+++ b/controlplane/pkg/apis/connectioncontext/helpers.go
@@ -5,7 +5,7 @@ import (
 	"net"
 )
 
-func (c *ConnectionContext) IsComplete() error {
+func (c *ConnectionContext) IsValid() error {
 	if c == nil {
 		return fmt.Errorf("ConnectionContext should not be nil...")
 	}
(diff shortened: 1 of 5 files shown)


controlplane/pkg/apis/remote/connection/helpers.go
```diff
@@ -54,7 +54,7 @@ func (c *Connection) SetContext(newContext *connectioncontext.ConnectionContext)
 	c.Context = newContext
 }
 
-// IsComplete - Have I been told enough to actually give you what you asked for
+// IsValid - Have I been told enough to actually give you what you asked for
```
haiodo: In case of rename, comment should be also changed.

denis-tingajkin: oops! Thanks, fixed.

controlplane/pkg/apis/connectioncontext/helpers.go
```diff
@@ -5,7 +5,7 @@ import (
 	"net"
 )
 
-func (c *ConnectionContext) IsComplete() error {
+func (c *ConnectionContext) IsValid() error {
```
fkautz: Can we rename this to something not starting with Is? Convention makes it sound like this returns a boolean.

Perhaps Validate() or something similar?

denis-tingajkin: Yes, you are right. ÌsValid`should return boolean by name convention.  I'd not changed this method name because we already have similar methods in the project. For example ÌsValid() for `ExtraPrefixRequest`.  Also in the proposal (#811 ) for this issue talking about creation ìsValid()` method. 

Can I also change all similar methods which have a signature `IsValid() error`?

PR #1021, opened 2019-04-30
[WIP] Stabilization of TestSimpleMemifConnection on GKE

lobkovilya:

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -439,48 +439,48 @@ workflows:
             - sanity-check
 
 # packet
-      - packet-deploy:
-          requires:
-            - sanity-check
-      - packet-deploy-k8s:
-          name: "packet-deploy-k8s-1"
-          cluster_id: "1"
-          requires:
-            - packet-deploy
-      - packet-deploy-k8s:
-          name: "packet-deploy-k8s-2"
-          cluster_id: "2"
-          requires:
-            - packet-deploy
-      - integration-tests:
-          name: "packet-test-basic-usecase"
-          namespace: "nsm-system-integration"
-          cloud_provider: "packet"
-          cluster_id: "1"
-          test_tags: "basic usecase"
-          requires:
-            - packet-deploy-k8s-1
-      - integration-tests:
-          name: "packet-test-recover"
-          namespace: "nsm-system-integration"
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -589,9 +589,9 @@ workflows:
...
-            - packet-destroy
-          filters:
-            branches:
-              only: master
+#      - docker-push-latest:
+#          requires:
+#            - packet-destroy
+#          filters:
+#            branches:
+#              only: master
```
edwarnicke: I presume you are narrowing the jobs pat here to speed checking your fix and will uncomment when you have it working?   If so: super smart :)

lobkovilya: Yes, seems like it fails only on gke and I try to get more logs :)

denis-tingajkin: @lobkovilya  you can also try run bench tests on GKE. I think it should help to reproduce the problem.

test/integration/basic_excluded_prefixes_test.go
```diff
@@ -33,7 +33,7 @@ func TestExcludePrefixCheck(t *testing.T) {
 	nodesCount := 1
 
 	variables := map[string]string{
-		nsmd.ExcludedPrefixesEnv: "10.20.1.0/24",
+		nsmd.ExcludedPrefixesEnv:     "200.20.1.0/24",
```
edwarnicke: Could we use a prefix from 172.16/12 (as noted in [RFC 1918](https://tools.ietf.org/html/rfc1918)).

I think we probably want to continue using private IPs here, but it looks like Google is only drawing from 10.0.0.0/8, so we could avoid the collision that way.  Thoughts?

lobkovilya: That's good point, but I'm not sure about other clouds. Since our tests are fine on Azure, probably they don't use 10.0.0.0/8. But I don't mind to try 172.16/12 and figure out :)

edwarnicke: Could we have the ipam pick based on avoiding the exclude prefixes?

Issue #895, opened 2019-04-02
vpp-agent latency monotomically increasing for scale testing


lobkovilya:
For testing purposes I created "greedy" nsc, that requests 50 connections.
First 45 connection established in ~300ms, but after that time of connection rapidly increases upto 7s and continue grows for the following requests.

nickolaev:
It's funny you found that. I have seen this with the examples:
https://github.com/networkservicemesh/examples/blob/master/examples/proxy/proxy-nsc/cmd/nsc.go

It appears that eventually the connections never get deleted at the NSE side and the dataplane gets overloaded.

lobkovilya:
Error happens in vpp-agent, when it tries to add ~45 ARP entries:
```
2019-04-18 04:21:35,751 DEBG 'agent' stdout output:
time="2019-04-18 04:21:35.75113" level=error msg="failed to add linux ARP entry: no buffer space available" loc="descriptor/arp.go(224)" logger=linux-l3plugin.arp-descriptor
``` 
Such amount of ARP entries comes from IPAM endpoint, which adds all IpNeighbors to Connection. Probably we should remove this logic from SDK.

*Note:* I removed this logic locally and 50 connections are easily created

edwarnicke:
@lobkovilya Could we get this reported to the vpp-agent guys as a bug?   Its a really really good catch that this seems to be the root of the problem... but vpp-agent shouldn't be having this issue.

lobkovilya:
Linux has limitation on the size of arp table, it could be increased by adding the following lines to `/etc/sysctl.conf`:
```
net.ipv4.neigh.default.gc_thresh1 = 4096
net.ipv4.neigh.default.gc_thresh2 = 8192
net.ipv4.neigh.default.gc_thresh3 = 8192
```

edwarnicke:
@lobkovilya Have you checked to see if raising the arp table limit avoids the issue?

How many arp entries are we creating in the SDK?  Do we need any of them, or can we allow normal arp behavior?

edwarnicke:
@lobkovilya Another question... is the arp table limit global, or per network namespace?

lobkovilya:
Yes, I have checked it with updated limit value and everything works fine. 

Ipam Endpoint sets Neighbors to Request.Context, so 45th request will get at least 44 neighbor. I set limit global on the node. But I can check, probably there are possibility to set it per namespace

lobkovilya:
[...]

PR #1026, opened 2019-04-30
Endpoints for setting Routes and Neighbors to connection

lobkovilya:

## Motivation and Context
Issue https://github.com/networkservicemesh/networkservicemesh/issues/895

## How Has This Been Tested?

- [X] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


haiodo:
@lobkovilya Should we also remove default route 8.8.8.8 as part of this SDK cleanup?

lobkovilya:
@haiodo I think we should parametrize Ipam Endpoint with routes. I planned to do it later because it doesn't affect on issue with 45 connections limit.

haiodo:
@lobkovilya should we in this case parametrise SDK with ARP entries? 
I think we could pass a function to SDK to process entries, routes, etc?

@nickolaev Thoughts?

nickolaev:
OK so the problem is, that the IPAM we provide with the SDK is kind of bare-bones and I think it should stay very simple. In theory IPAM is a function on its own and is not part of NSM per se. We use it heavily in our examples and tests (CI), that is why it landed in the SDK. We have two approaches here:
 1. state that the IPAM is a simplified example; an MVP, and we keep it as simple as possible to run our tests. Anyone that needs more complex functions shall implement them as an Endpoint
 2. add the Routing and ARP functions as extra composable layers, and people may choose to add them and/or omit them when they create their endpoints.

I would vote for the latter (implement the two new composites) if we have the resources to implement. In this case instead of removing the code, we move it to another file and just insert the ARP. Same for the routes. Of course these layer can be configurable.

lobkovilya:
@nickolaev I like the second approach as well. I think I can implement it within this PR

sdk/endpoint/neighbor.go
```diff
@@ -0,0 +1,65 @@
...
+		return nil, err
+	}
+
+	newConnection, err := ne.GetNext().Request(ctx, request)
+	if err != nil {
+		logrus.Errorf("Next request failed: %v", err)
+		return nil, err
+	}
+
+	addrs, err := net.Interfaces()
```
haiodo: I think this is useless to have, since this is just was used as example for capturing list of arp entries, I think better to provide function for users or list of arp entries to pass.

lobkovilya: It seems like we need more real-world scenarios for neighbors endpoint, because now we don't know how it will be used. So I'd prefer to leave endpoint as it is until use cases become clearer. The main idea of that PR - exclude setting neighbors from widely used IPAM endpoint.

sdk/common/configuration.go
```diff
@@ -45,6 +46,7 @@ type NSConfiguration struct {
 	TracerEnabled      bool
 	MechanismType      string
 	IPAddress          string
+	Routes             []string
```
nickolaev: This also needs an update in `sdk/README.md`.

sdk/endpoint/neighbor.go
nickolaev: I can agree to keep the code as it is here. It can be used as an example and if we need we can move improve it in a follow-up PR.

nickolaev:
@lobkovilya please also update the `Pre-defined composites` at the end of the `sdk/README.md`.

sdk/endpoint/neighbor.go
haiodo: SDK should not have any logic this way, since it was written only to check ARP entries and i'm mostly sure any real NSE will not do it this way :-)

nickolaev:
@haiodo I'd like this one merged. Can you please approve the changes if you agree with them? I'm rerunung the CI now.

PR #1031, opened 2019-05-01
GKE: Set podCIDR explicitly during cluster creation

lobkovilya:

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/.mk/gke.mk
+++ b/.mk/gke.mk
@@ -5,6 +5,7 @@ endif
 .PHONY: gke-start
 gke-start: gcloud-check
 	@if ! (gcloud container clusters list | grep -q ^${GKE_CLUSTER_NAME}); then \
+	    gcloud compute networks subnets list; \
 		time gcloud container clusters create ${GKE_CLUSTER_NAME} --machine-type=n1-standard-2 --num-nodes=2 -q --cluster-ipv4-cidr=10.128.0.0/19; \
 		gcloud container clusters get-credentials ${GKE_CLUSTER_NAME}; \
 		kubectl create clusterrolebinding cluster-admin-binding \
--- a/.mk/gke.mk
+++ b/.mk/gke.mk
@@ -5,7 +5,7 @@ endif
 .PHONY: gke-start
 gke-start: gcloud-check
 	@if ! (gcloud container clusters list | grep -q ^${GKE_CLUSTER_NAME}); then \
-		time gcloud container clusters create ${GKE_CLUSTER_NAME} --machine-type=n1-standard-2 --num-nodes=2 -q; \
+		time gcloud container clusters create ${GKE_CLUSTER_NAME} --machine-type=n1-standard-2 --num-nodes=2 -q --cluster-ipv4-cidr=10.128.0.0/19; \
 		gcloud container clusters get-credentials ${GKE_CLUSTER_NAME}; \
 		kubectl create clusterrolebinding cluster-admin-binding \
 			--clusterrole cluster-admin \
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -439,84 +439,84 @@ workflows:
             - sanity-check
 
 # packet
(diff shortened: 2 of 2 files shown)


.mk/gke.mk
```diff
@@ -5,7 +5,8 @@ endif
 .PHONY: gke-start
 gke-start: gcloud-check
 	@if ! (gcloud container clusters list | grep -q ^${GKE_CLUSTER_NAME}); then \
-		time gcloud container clusters create ${GKE_CLUSTER_NAME} --machine-type=n1-standard-2 --num-nodes=2 -q; \
+	    gcloud compute networks subnets list; \
+		time gcloud container clusters create ${GKE_CLUSTER_NAME} --machine-type=n1-standard-2 --num-nodes=2 -q --cluster-ipv4-cidr=10.128.0.0/19; \
```
edwarnicke: Question: Why are we setting this instead of adapting to what GKE assigns?

lobkovilya: I thought it will be quick fix, but after fifth unsuccessful attempt seems like it is not so easy. 
So you think we need monitoring podCIDR and serviceCIDR from test and wisely choose subnet for NSE?
Or we can create smart NSE that will do it itself? It can accept only size of required subnet from ENV.

edwarnicke: @lobkovilya So I think we have several things we need to test:

1.  That we correctly identify the PodCIDR and serviceCIDR
2.  That we correctly express them in the Request
3.  That the SDK correctly issues non-conflicting IPAM in the NSEs
4.  That if it does issue conflicting IPAM we reject it and close the connection and try a differnet NSE.

Its not clear to me all of those have to be checked in integration testing, but they are all a part of the contract we want to make sure we are doing correct.

Another modifier is: is GKE testing currently broken and breaking CI for unrelated patches?  If so a fix quickly and then a more thoughtful solution may be warranted.

lobkovilya: From logs that leads to test failing:
*nsmd-k8s*:
```
time=\"2019-04-30T13:58:19Z\" level=info msg=\"Receive resource: name gke-dev-testing-566dcde9-default-pool-d157b11e-stl1, subnet 10.20.0.0/24\"
time=\"2019-04-30T13:58:19Z\" level=info msg=\"Receive resource: name gke-dev-testing-566dcde9-default-pool-d157b11e-1hst, subnet 10.20.1.0/24\"
```
This is PodCIDR from 2 nodes, and after merging we have `10.20.0.0/23`. Exactly the same CIDR we can find in Request to NSE:
```
time=\"2019-04-30T13:58:31Z\" level=info msg=\"NSM:(7.2.6.2-CD138CD7) Requesting NSE with request connection:<id:\\\"2\\\" network_service:\\\"icmp-responder\\\" context:<src_ip_required:true dst_ip_required:true excluded_prefixes:\\\"10.20.0.0/23\\\" excluded_prefixes:\\\"10.23.240.0/20\\\" > labels:<key:\\\"app\\\" value:\\\"icmp\\\" > > mechanism_preferences:<type:MEM_INTERFACE > mechanism_preferences:<type:KERNEL_INTERFACE > \"
```
So it's predictable that NSE with subnet `10.20.1.0/24` returns error after such request:
```
time=\"2019-04-30T13:59:02Z\" level=error msg=\"NSM:(7.1.8-4725999E) NSE respond with error: rpc error: code = Unknown desc = IPAM: The available address pool is empty, probably intersected by excludedPrefix \"
```
And since there are no more NSEs in that test, it fails.

As I figured out from GKE documentation in auto mode it configures subnets for each region. Probably that's why we have unstable test.

lobkovilya: I think the quickest fix is change subnet for NSE. Probably chose one which is not mentioned here https://tools.ietf.org/html/rfc1918 :)

PR #1039, opened 2019-05-06
Fix empty exclude prefix fault 

artem-belov:
Fix empty exclude prefix fault 

## Description
Skip empty subnets on prefix pool formation

## Motivation and Context
icmp-responder returns fault on AWS cluster. Cause of the fault is request from nsmd with empty pod subnet. 

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/controlplane/pkg/nsm/nsm_prefix_collector.go
+++ b/controlplane/pkg/nsm/nsm_prefix_collector.go
@@ -130,7 +130,15 @@ func (l *subnetStreamListener) listen() {
 			serviceSubnet = extendResponse.Subnet
 		}
 
-		pool, err := prefix_pool.NewPrefixPool(append(l.additionalPrefixes, podSubnet, serviceSubnet)...)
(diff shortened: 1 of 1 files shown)


controlplane/pkg/nsm/nsm_prefix_collector.go
```diff
@@ -130,7 +130,15 @@ func (l *subnetStreamListener) listen() {
 			serviceSubnet = extendResponse.Subnet
 		}
 
-		pool, err := prefix_pool.NewPrefixPool(append(l.additionalPrefixes, podSubnet, serviceSubnet)...)
+		var pool prefix_pool.PrefixPool
+		if len(podSubnet) == 0 {
+			pool, err = prefix_pool.NewPrefixPool(append(l.additionalPrefixes, serviceSubnet)...)
```
haiodo: Let's do append and only then construct NewPrefixPool

Issue #1024, opened 2019-04-30
nsm-admission-webhook is not getting cleaned up on TestNSCAndICMPWebhookLocal restart


Bolodya1997:
## Expected Behavior
1. **nsm-admission-webhook** deployment is being removed from cluster on `TestNSCAndICMPWebhookLocal` restart.
2. Test successes.

## Current Behavior
1. **nsm-admission-webhook** deployment stays on cluster after `TestNSCAndICMPWebhookLocal` restart.
2. Test fails.

## Steps to Reproduce

1. Start `TestNSCAndICMPWebhookLocal` test.
2. Kill it with a signal after the **nsm-admission-webhook** deployment.
3. Start `TestNSCAndICMPWebhookLocal` test.
4. Test fails...

## Context

* Kubernetes Version:
* etc.

nickolaev:
#1040 merged and issues fixed.

PR #1040, opened 2019-05-06
Cleanup more resources on test startup

Bolodya1997:
## Description
Cleans up k8s services, deployments, mutating webhook configurations, secrets on test startup.

## Motivation and Context
Issue #1024 

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/test/kube_testing/k8s.go
+++ b/test/kube_testing/k8s.go
@@ -283,8 +283,12 @@ func NewK8sWithoutRoles(prepare bool) (*K8s, error) {
 		start := time.Now()
 		client.Prepare("nsmgr", "nsmd", "vppagent", "vpn", "icmp", "nsc", "source", "dest")
 		client.CleanupCRDs()
+		client.CleanupServices("nsm-admission-webhook-svc")
+		client.CleanupDeployments()
+		client.CleanupMutatingWebhookConfigurations()
+		client.CleanupSecrets("nsm-admission-webhook-certs")
 		client.CleanupConfigMaps()
-		nsmrbac.DeleteAllRoles(client.clientset)
+		_ = nsmrbac.DeleteAllRoles(client.clientset)
 		logrus.Printf("Cleanup done: %v", time.Since(start))
 	}
 	return &client, nil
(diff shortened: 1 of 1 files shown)


test/kube_testing/k8s.go
```diff
@@ -283,8 +283,12 @@ func NewK8sWithoutRoles(prepare bool) (*K8s, error) {
 		start := time.Now()
 		client.Prepare("nsmgr", "nsmd", "vppagent", "vpn", "icmp", "nsc", "source", "dest")
 		client.CleanupCRDs()
+		client.CleanupServices("nsm-admission-webhook-svc")
+		client.CleanupDeployments()
+		client.CleanupMutatingWebhookConfigurations()
+		client.CleanupSecrets("nsm-admission-webhook-certs")
```
nickolaev: Hmm, in general, that is right. On the other hand aren't we adding too many corner cases here?
For example, the admission webhook is used only in 2 tests. Can we just have a `defer`  to clean up those 2 specific tests and not put the service removal code here for each and every test we run?

Bolodya1997: There is already used `defer` to clean up all these resources.
The described scenario for the issue means that there is some reason for `defer` not being executed - it can be either a hung test or a test killed by the signal.

haiodo: I'm ok if this not involves more time to cleanup.

Bolodya1997: It seems to increase cleanup time for ~0.1-0.5 second for test in different clouds.
Added a bool param to `NewK8s()` to enable / disable "extra cleanup".

test/integration/basic_dataplane_standalone_test.go
```diff
@@ -139,7 +139,7 @@ func createFixture(test *testing.T, timeout time.Duration) *standaloneDataplaneF
 		test:    test,
 	}
 
-	k8s, err := kube_testing.NewK8s(true)
+	k8s, err := kube_testing.NewK8s(true, false)
```
haiodo: Let's by default cleanup all, why we need to pass it

Bolodya1997: OK, removed the corresponding commit.

haiodo:
I'm ok for extra cleanups, @nickolaev your final decision. Thanks.

PR #1065, opened 2019-05-08
ci: always install latest kubectl

nickolaev:

## Description
Do not fix the `kubectl` version in our test scripts, use the latest stable instead.

## Motivation and Context
Just stay up with the current version.

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -268,7 +268,6 @@ jobs:
             export COMMIT="${CIRCLE_SHA1:8:8}"
             export CONTAINER_TAG="${COMMIT}"
             # export CONTAINER_FORCE_PULL="true"
-            ./scripts/install-kubectl.sh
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -268,7 +268,6 @@ jobs:
             export COMMIT="${CIRCLE_SHA1:8:8}"
             export CONTAINER_TAG="${COMMIT}"
             # export CONTAINER_FORCE_PULL="true"
-            ./scripts/install-kubectl.sh
```
haiodo: Is this line removal is required?

nickolaev: There is no point for calling the script here since it is the first thing called in `./scripts/prepare-circle-integration-tests.sh` a line later.

PR #1087, opened 2019-05-13
Add infinity number of attempts for create/update NSM

lobkovilya:

## Motivation and Context

Since k8s-registry doesn't support transactions and doesn't have kind of *createOrUpdate* method we can't atomically figure out resource existence and call appropriate (create/update) method. That's why we always can get error. I provided additional attempts during error handling

## How Has This Been Tested?

- [X] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/k8s/pkg/registryserver/registry_cache.go
+++ b/k8s/pkg/registryserver/registry_cache.go
@@ -146,7 +146,7 @@ func (rc *registryCacheImpl) CreateOrUpdateNetworkServiceManager(nsm *v1.Network
 		if existingNsm == nil {
 			logrus.Infof("Creating NSM: %v", nsm)
 			newNsm, err := rc.addNetworkServiceManager(nsm)
-			if err == nil || (err != nil && !apierrors.IsAlreadyExists(err)) {
+			if err == nil || !apierrors.IsAlreadyExists(err) {
 				return newNsm, err
 			}
 
@@ -160,7 +160,7 @@ func (rc *registryCacheImpl) CreateOrUpdateNetworkServiceManager(nsm *v1.Network
 			updNsm := nsm.DeepCopy()
 			updNsm.ObjectMeta = existingNsm.ObjectMeta
 			updNsm, err := rc.updateNetworkServiceManager(updNsm)
-			if err == nil || (err != nil && !apierrors.IsConflict(err)) {
+			if err == nil || !apierrors.IsConflict(err) {
 				return updNsm, err
 			}
 
--- a/k8s/pkg/registryserver/registry_cache.go
+++ b/k8s/pkg/registryserver/registry_cache.go
@@ -19,7 +19,7 @@ type RegistryCache interface {
 	GetNetworkService(name string) (*v1.NetworkService, error)
 
 	CreateOrUpdateNetworkServiceManager(nsm *v1.NetworkServiceManager) (*v1.NetworkServiceManager, error)
-	GetNetworkServiceManager(name string) *v1.NetworkServiceManager
+	GetNetworkServiceManager(name string) (*v1.NetworkServiceManager, error)
 
 	AddNetworkServiceEndpoint(nse *v1.NetworkServiceEndpoint) (*v1.NetworkServiceEndpoint, error)
 	DeleteNetworkServiceEndpoint(endpointName string) error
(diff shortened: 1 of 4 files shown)


k8s/pkg/registryserver/registry_cache.go
```diff
@@ -135,24 +135,36 @@ func (rc *registryCacheImpl) GetEndpointsByNsm(nsmName string) []*v1.NetworkServ
...
-	}
-
-	logrus.Infof("Creating NSM: %v", nsm)
-	createNsm, err := rc.addNetworkServiceManager(nsm)
-	if err != nil || apierrors.IsAlreadyExists(err) {
-		existingNsm, err = rc.clientset.NetworkservicemeshV1().NetworkServiceManagers(rc.nsmNamespace).Get(nsm.Name, metav1.GetOptions{})
-		if err != nil {
-			return existingNsm, err
+	attempt := 0
+	for {
```
haiodo: I suppose we should still have some timeout, line 1 minute, and don' t try doing this infinite.

nickolaev: +1 Infinity is a looong time. Let's try to avoid it :)

lobkovilya: In practice maximum possible amount of attempts is 2, more attempts is really low probability event. And we don't need timeout, second attempt should be done as soon as possible, while we know the most recent state of resource in registry.

Also it is not really infinite loop, if there are errors that we don't expect to get (i.e. connection refused) we will break

haiodo: Ok, in this case it is fine with me.

k8s/pkg/registryserver/registry_cache.go
```diff
@@ -134,27 +134,43 @@ func (rc *registryCacheImpl) GetEndpointsByNsm(nsmName string) []*v1.NetworkServ
 	return rc.networkServiceEndpointCache.GetByNetworkServiceManager(nsmName)
 }
 
+const maxAllowedAttempts = 10
```
nickolaev: Should this go to the beginning of the file?

k8s/pkg/registryserver/registry_cache.go
```diff
@@ -134,27 +134,43 @@ func (rc *registryCacheImpl) GetEndpointsByNsm(nsmName string) []*v1.NetworkServ
...
-	}
-
-	logrus.Infof("Creating NSM: %v", nsm)
-	createNsm, err := rc.addNetworkServiceManager(nsm)
-	if err != nil || apierrors.IsAlreadyExists(err) {
-		existingNsm, err = rc.clientset.NetworkservicemeshV1().NetworkServiceManagers(rc.nsmNamespace).Get(nsm.Name, metav1.GetOptions{})
-		if err != nil {
-			return existingNsm, err
+	attempt := 0
+	for attempt < maxAllowedAttempts {
```
nickolaev: Why not `for attempt := 0; attempt < mazAllowedAttempts; attempt ++`?

k8s/pkg/registryserver/registry_cache.go
```diff
@@ -134,27 +134,43 @@ func (rc *registryCacheImpl) GetEndpointsByNsm(nsmName string) []*v1.NetworkServ
...
+			logrus.Infof("Updating existing NSM: %v with %v", existingNsm, nsm)
+			updNsm := nsm.DeepCopy()
+			updNsm.ObjectMeta = existingNsm.ObjectMeta
+			updNsm, err := rc.updateNetworkServiceManager(updNsm)
+			if err == nil || !apierrors.IsConflict(err) {
+				return updNsm, err
+			}
+
+			logrus.Infof("There is no NSM with name %v", nsm.GetName())
+			existingNsm = nil
```
nickolaev: whole this `if .. else ...` looks very strange. I am sure it can be rewritten in a clearer fashion.
So for example here you are in the branche where there is `existingNsm != nil` and then you find that there is no NSMgr with that name. Is that even possible?
Please rethink the logic here.

test/integration/basic_nsmd_k8s_test.go
```diff
@@ -74,9 +74,6 @@ func TestNSMDDRegistryNSE(t *testing.T) {
 	registryClient, err := serviceRegistry.NseRegistryClient()
 	Expect(err).To(BeNil())
 
-	// Cleanup all registered stuff
-	k8s.CleanupCRDs()
-
```
nickolaev: Interesting?

lobkovilya: Moved here: https://github.com/networkservicemesh/networkservicemesh/blob/master/test/kube_testing/k8s.go#L293

k8s/pkg/registryserver/registry_cache.go
lobkovilya: Yes, that exactly what happened sometimes in tests. Previous test sent DELETE to k8s and it was handled by k8s with some delay. So we have NSM in cache, but k8s-registry already empty, so Update will fail. 

About rewritting, actually I found this way of implement that logic the most readable. But I'll think about others ways more carefully

k8s/pkg/registryserver/registry_cache.go
lobkovilya: Absolutely agree, I was adding attempts later and accidently do it in such stupid way :) Will fix

k8s/pkg/registryserver/registry_cache.go
lobkovilya: Probably yes, seems like we keep such style in project. Will fix

PR #979, opened 2019-04-19
Added NSE monitoring to NSM

Bolodya1997:
## Description

Added NSE monitoring to NSM. If connection to NSE dies, NSE is getting deleted from the registry.
If connection to NSE is not getting alive in 10 seconds, NSE is getting deleted from the registry.

## Motivation and Context

NSMD should periodically check NSE without connections are alive and update registry since field. #836

## How Has This Been Tested?

- [ ] Covered by existing integration testing
- [x] Added integration testing to cover
- [x] Tested locally
- [ ] Have not tested

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1122, opened 2019-05-17
Fix linter issues

Bolodya1997:
## Description
Fix general linter issues for #979

## Motivation and Context
1. Fix general linter issues for #979
2. Issue #1124

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


.golangci.yml
```diff
@@ -103,5 +104,11 @@ issues:
     - path: _test\.go
       linters:
         - golint
+    - path: localconnectionmonitor/local_connection_event\.go
```
denis-tingajkin: can we fix duplicaiton?

Bolodya1997: No, because there is a need to cast map from abstract type to package-special and the same for enum values.

haiodo:
I’m not like to rename packages with underscores to this long lines, could we try disable linter for this cases?

Bolodya1997:
> I’m not like to rename packages with underscores to this long lines, could we try disable linter for this cases?

@haiodo I have renamed this:
```
- monitor
-- localconnectionmonitor
-- remoteconnectionmonitor
-- crossconnectmonitor
```
to that:
```
- monitor
-- local
-- remote
-- crossconnect
```
so I believe there is no more problem with linter for this case

PR #1158, opened 2019-05-24
Model refactoring

lobkovilya:
## Description

Now model always returns and stores copies of passed objects. Its interface didn't change. Also covered the model with tests.

## Motivation and Context

There were many places in the code where we indirectly changed fields of structures from the model. 

## How Has This Been Tested?

- [X] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


controlplane/pkg/model/client_connection.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+type clientConnectionDomain struct {
+	baseDomain
+	inner sync.Map
+}
+
+func (d *clientConnectionDomain) AddClientConnection(cc *ClientConnection) {
+	d.inner.Store(cc.ConnectionId, cc.Clone())
+	d.resourceAdded(cc.Clone())
```
Bolodya1997: There can be a case when model is being modified concurrently from 2 different threads and so we can get the following order of execution:
```
T1: AddClientConnection(cc)                  | T2: DeleteClientConnection(cc.GetId())
-------------------------------------------------------------------------------------------
 d.inner.Store(cc.ConnectionId, cc.Clone())  | *
 *                                           | v := d.GetClientConnection(id)
 *                                           | if v == nil {
 *                                           |     return
 *                                           | }
 *                                           | d.inner.Delete(id)
 *                                           | d.resourceDeleted(v)
 d.resourceAdded(cc.Clone())                 | *
-------------------------------------------------------------------------------------------
```
And so all listeners will first receive `delete` event and then `add` event. I believe this can lead to some error state.

controlplane/pkg/services/client_connection_manager.go
```diff
@@ -33,68 +33,65 @@ func (m *ClientConnectionManager) GetNsmName() string {
...
 
-func (m *ClientConnectionManager) UpdateClientConnectionDstStateDown(clientConnection *model.ClientConnection, nsmdDie bool) {
-	logrus.Info("ClientConnection dst state is down")
-	if clientConnection.Xcon.GetLocalDestination() != nil {
-		clientConnection.Xcon.GetLocalDestination().State = connection.State_DOWN
-	} else if clientConnection.Xcon.GetRemoteDestination() != nil {
-		clientConnection.Xcon.GetRemoteDestination().State = remote_connection.State_DOWN
-	}
-	m.model.UpdateClientConnection(clientConnection)
+func (m *ClientConnectionManager) RemoteDestinationDown(cc *model.ClientConnection, nsmdDie bool) {
```
Bolodya1997: This can be used as for remote destination down so for local destination down (if flag is set as `true`). Do we need this **Remote** prefix here?

lobkovilya: I have checked, this method is called only from `remotePeerConnectionMonitor`, so I think **Remote** prefix is ok

controlplane/pkg/model/client_connection.go
lobkovilya: Yes, you are right, I will fix it

controlplane/pkg/model/client_connection.go
```diff
@@ -0,0 +1,175 @@
...
+	})
+	return rv
+}
+
+func (d *clientConnectionDomain) DeleteClientConnection(id string) {
+	v := d.GetClientConnection(id)
+	if v == nil {
+		return
+	}
+	d.inner.Delete(id)
```
haiodo: v := Delete(id) or use lock.

controlplane/pkg/model/model.go
```diff
@@ -150,266 +80,111 @@ type Model interface {
...
+	clientConnectionDomain
+
+	lastConnectionID uint64
+	mtx              sync.RWMutex
+	selector         selector.Selector
+	nsm              *registry.NetworkServiceManager
+	listeners        map[ModelListener]func()
+}
+
+func (m *model) AddListener(listener ModelListener) {
```
haiodo: Let's move listeners to separate file.

PR #1160, opened 2019-05-24
Rename 'examples' to '/test/applications'

Bolodya1997:
## Description
Rename `examples` directory to `/test/applications`.

## Motivation and Context
There is already `networkservicemesh/examples` repo for examples. Applications from `examples` directory are widely used and modified for testing, so this naming becomes confusing.

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


nickolaev:
We will have to think about the 'nsc' which is used as an init container. Do we want it to live in a 'side-cars' folder or something?

haiodo:
> We will have to think about the 'nsc' which is used as an init container. Do we want it to live in a 'side-cars' folder or something?

I suppose yes, we should have "init container" and I hope in future "monitor-container" as side-cars somehow, since they will be part of distribution and put example CI container as destribution one is not good. @Bolodya1997 Could you have put

Bolodya1997:
@nickolaev 
> Is there a need to fix documentation here?
> What about the `helm` charts?

This PR is mostly about renaming and moving files - all links on these renamed files have been updated as well, so there should not be any problems with documentation and docker images. `helm` charts, `make` targets and `k8s` configurations depends only on docker images, so there should not be any problems at all.

> Shall we remove the whole `examples` concept?

`examples` has been finally splitted up in `test/applications` and `side-cars` - what do you mean by "removing whole examples concept"?

> Are targets like `k8s-vpn-deploy` still relevant?
> Do we need the `k8s-icmp-check` scripts still?

`k8s-icmp/vpn-deploy`, `k8s-check` targets are still relevant because they are used in CI to test NSM deployments.

nickolaev:
@Bolodya1997 OK, that makes sense. Still, I would like someone (@lobkovilya ?) to confirm that helm still works in this branch.

Bolodya1997:
@nickolaev 
There is an issue with helm charts: #1201.
With this fix applied everything is OK in this branch.

PR #1189, opened 2019-06-03
[WIP] Refactor Model::ApplyClientConnectionChanges

Bolodya1997:
## Description
1. Add `CompareAndSwapClientConnection`.
2. Make `Delete*`, `Add*`, `Apply*Changes` methods return `error`.
3. Rename `Update*` to `AddOrUpdate*`.
4. Refactor `request` and healing.

## Motivation and Context
Existing approach with `ApplyClientConnectionChanges` looks not so good.

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/controlplane/pkg/model/base.go
+++ b/controlplane/pkg/model/base.go
@@ -1,8 +1,9 @@
 package model
 
 import (
-	"github.com/sirupsen/logrus"
 	"sync"
+
+	"github.com/sirupsen/logrus"
 )
 
 // ModificationHandler aggregates handlers for particular events
@@ -39,7 +40,7 @@ func (b *baseDomain) load(key string) (interface{}, bool) {
 	return v.clone(), true
 }
 
-func (b *baseDomain) store(key string, value cloneable) {
+func (b *baseDomain) store(key string, value cloneable, update bool) bool {
 	b.mtx.Lock()
 	defer b.mtx.Unlock()
 
@@ -47,22 +48,30 @@ func (b *baseDomain) store(key string, value cloneable) {
 	if !exist {
 		b.innerMap[key] = value.clone()
 		b.resourceAdded(value)
-		return
+		return true
+	}
+
+	if !update {
+		return false
 	}
 
 	b.innerMap[key] = value.clone()
 	b.resourceUpdated(old, value)
+	return true
 }
 
-func (b *baseDomain) delete(key string) {
+func (b *baseDomain) delete(key string) bool {
 	b.mtx.Lock()
 	defer b.mtx.Unlock()
+
 	v, exist := b.innerMap[key]
 	if !exist {
-		return
+		return false
 	}
+
 	delete(b.innerMap, key)
 	b.resourceDeleted(v)
+	return true
 }
 
 func (b *baseDomain) applyChanges(key string, changeFunc func(interface{})) interface{} {
@@ -80,6 +89,20 @@ func (b *baseDomain) applyChanges(key string, changeFunc func(interface{})) inte
 	b.innerMap[key] = upd.clone()
 	b.resourceUpdated(old, upd)
 	return upd
+}
+
+func (b *baseDomain) compareAndSwap(key string, value cloneable, compareFunc func (interface{}) bool) bool {
+	b.mtx.Lock()
+	defer b.mtx.Unlock()
+
+	old, exists := b.innerMap[key]
+	if !exists || !compareFunc(old) {
+		return false
+	}
+
+	b.innerMap[key] = value.clone()
+	b.resourceUpdated(old, value)
+	return true
 }
 
 func (b *baseDomain) kvRange(f func(key string, v interface{}) bool) {
--- a/controlplane/pkg/model/client_connection.go
+++ b/controlplane/pkg/model/client_connection.go
@@ -33,14 +33,15 @@ const (
 
 // ClientConnection struct in model that describes cross connect between NetworkServiceClient and NetworkServiceEndpoint
 type ClientConnection struct {
-	ConnectionID            string
 	Request                 networkservice.Request
 	Xcon                    *crossconnect.CrossConnect
 	RemoteNsm               *registry.NetworkServiceManager
 	Endpoint                *registry.NSERegistration
 	DataplaneRegisteredName string
-	ConnectionState         ClientConnectionState
 	DataplaneState          DataplaneState
+
+	id              string
+	connectionState ClientConnectionState
 }
 
 // GetID returns id of clientConnection
@@ -48,7 +49,12 @@ func (cc *ClientConnection) GetID() string {
 	if cc == nil {
 		return ""
 	}
-	return cc.ConnectionID
+	return cc.id
+}
+
+// GetConnectionState returns state of clientConnection
+func (cc *ClientConnection) GetConnectionState() ClientConnectionState {
+	return cc.connectionState
 }
 
 // GetNetworkService returns name of networkService of clientConnection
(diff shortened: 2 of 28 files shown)


controlplane/pkg/model/base.go
```diff
@@ -57,28 +57,35 @@ func (b *baseDomain) store(key string, value cloneable) {
...
 
-func (b *baseDomain) applyChanges(key string, changeFunc func(interface{})) interface{} {
+func (b *baseDomain) applyChanges(key string, value cloneable, changeFunc func(interface{})) interface{} {
 	b.mtx.Lock()
 	defer b.mtx.Unlock()
 
 	old, exist := b.innerMap[key]
 	if !exist {
-		return nil
+		old = value
```
lobkovilya: I think it is a bit unobvious for customers of ApplyClientConnectionChanges method. Passed pointer to structure will be correctly updated even if there is no more such entity in the model. Probably it makes sense to return some boolean flag? What do you think?

denis-tingajkin:
Looks like It is not a part of your pull request but can you consider refactoring `baseDomain` to use `SyncMap` instead using `map` and `sync.RWMutex`?

Bolodya1997:
@denis-tingajkin
There is an issue if we use `SyncMap` instead using `map` and `sync.RWMutex` : https://github.com/networkservicemesh/networkservicemesh/pull/1158#discussion_r287937791

denis-tingajkin:
@Bolodya1997 I thought `baseDomain` has only simple operations on the `map`. Yep, you are right we can't use `sync.Map` in this place

controlplane/pkg/model/base.go
Bolodya1997: I have changed `Add*` [1], `Delete*` [2], `Apply*Changes` [3] methods to return `error` if:
1. `key` exists in domain
2, 3. `key` doesn't exist in domain
5c5ca2f

controlplane/pkg/model/client_connection.go
```diff
@@ -134,20 +138,19 @@ func (d *clientConnectionDomain) GetAllClientConnections() []*ClientConnection {
...
-func (d *clientConnectionDomain) UpdateClientConnection(cc *ClientConnection) {
-	d.store(cc.ConnectionID, cc)
+func (d *clientConnectionDomain) DeleteClientConnection(id string) error {
+	if ok := d.delete(id); ok {
+		return nil
+	}
+	return fmt.Errorf("trying to delete client connection by not existing id: %v", id)
 }
 
-func (d *clientConnectionDomain) ApplyClientConnectionChanges(id string, f func(*ClientConnection)) *ClientConnection {
```
haiodo: Both variants are actually not so good, we need to track after call if anyone changed our model value.

haiodo: Probable compare and swap should be applied.

Bolodya1997:
Depends on #979.

controlplane/pkg/model/client_connection.go
Bolodya1997: I guess we still need `ApplyClientConnectionChanges` for some rare cases when we only need to update some parameter without any knowledge about what do we have in model.
Also I have added `CompareAndSwapClientConnection` : 88ced90

controlplane/pkg/model/base.go
```diff
@@ -39,30 +40,38 @@ func (b *baseDomain) load(key string) (interface{}, bool) {
 	return v.clone(), true
 }
 
-func (b *baseDomain) store(key string, value cloneable) {
+func (b *baseDomain) store(key string, value cloneable, update bool) bool {
 	b.mtx.Lock()
 	defer b.mtx.Unlock()
 
 	old, exist := b.innerMap[key]
 	if !exist {
```
lobkovilya: What if `update = true`, should we return `false` at that point?

controlplane/pkg/nsm/nsm_heal.go
```diff
@@ -10,32 +10,31 @@ import (
...
 	healID := create_logid()
-	logrus.Infof("NSM_Heal(1-%v) %v", healID, connection)
+	logrus.Infof("NSM_Heal(%v) %v", healID, connection)
 
 	cc := connection.(*model.ClientConnection)
-	if cc.ConnectionState != model.ClientConnectionReady {
-		//means that we already closing/healing
+	cc.ConnectionState = model.ClientConnectionHealing
+	if ok := srv.model.CompareAndSwapClientConnection(cc, func(connection *model.ClientConnection) bool {
+		return connection.ConnectionState == model.ClientConnectionReady
```
lobkovilya: Hm, it's a bit strange, for example, `ClientConnection.Xcon` could be changed in the model, but we will overwrite it with `cc.Xcon` which is outdated

controlplane/pkg/model/client_connection.go
```diff
@@ -142,22 +146,24 @@ func (d *clientConnectionDomain) GetAllClientConnections() []*ClientConnection {
...
 }
 
-func (d *clientConnectionDomain) DeleteClientConnection(id string) {
-	d.delete(id)
-}
-
-func (d *clientConnectionDomain) UpdateClientConnection(cc *ClientConnection) {
-	d.store(cc.ConnectionID, cc)
+func (d *clientConnectionDomain) DeleteClientConnection(id string) error {
+	if ok := d.delete(id); ok {
```
denis-tingajkin: [minor] looks like `delete` returns boolean. Can we remove `ok` variable?

controlplane/pkg/model/dataplane.go
```diff
@@ -62,24 +64,29 @@ func newDataplaneDomain() dataplaneDomain {
...
 }
 
-func (d *dataplaneDomain) DeleteDataplane(name string) {
-	d.delete(name)
-}
-
-func (d *dataplaneDomain) UpdateDataplane(dp *Dataplane) {
-	d.store(dp.RegisteredName, dp)
+func (d *dataplaneDomain) DeleteDataplane(name string) error {
+	if ok := d.delete(name); ok {
```
denis-tingajkin: [minor] looks like delete returns boolean. Can we remove ok variable?

controlplane/pkg/model/endpoint.go
```diff
@@ -73,12 +83,11 @@ func (d *endpointDomain) GetEndpointsByNetworkService(nsName string) []*Endpoint
...
 }
 
-func (d *endpointDomain) DeleteEndpoint(name string) {
-	d.delete(name)
-}
-
-func (d *endpointDomain) UpdateEndpoint(endpoint *Endpoint) {
-	d.store(endpoint.EndpointName(), endpoint)
+func (d *endpointDomain) DeleteEndpoint(name string) error {
+	if ok := d.delete(name); ok {
```
denis-tingajkin: same

controlplane/pkg/model/base.go
Bolodya1997: No, `update = true` means `AddOrUpdate`.

controlplane/pkg/nsm/nsm_heal.go
Bolodya1997: This has been completely reworked with `ClientConnectionEditor`.

controlplane/pkg/model/dataplane.go
Bolodya1997: Done : a7226f2

controlplane/pkg/model/client_connection.go
Bolodya1997: The approach was ultimately reworked with `model.ClientConnectionEditor`.

nickolaev:
@haiodo shall we consider this one obsolete already?

haiodo:
Closing, since we will do a complex NSM refactoring for SDK.

Issue #1191, opened 2019-06-03
Some tests in test-packet-usecase-ipv6 fail


nickolaev:
## Expected Behavior

Tests should pass

## Current Behavior

Some IPv6 integration tests fail

## Failure Information (for bugs)

Tests `TestNSCAndICMPLocalVeth` and `TestNSCAndICMPRemoteVeth` do fail.

## Steps to Reproduce

Run CI.

## Context

CircleCI

## Failure Logs

https://circleci.com/gh/networkservicemesh/networkservicemesh/62375

PR #1194, opened 2019-06-03
Fix dadfailure when configured with IPv6

rdimitrov:
Both ends of the VETH pair were assigned with the same IP address

## Description

Some of the IPv6 tests were failing sporadically due to a strange race condition caused by the fact that we assign the same IP address on both ends of the VETH pair. 
This was causing the IPv6 interface to stay in a `tentative` and `dadfailed` state.

The solution is to assign an address only to the VETH end that will be used in the container.

## Motivation and Context

Related to https://github.com/networkservicemesh/networkservicemesh/issues/1191

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


edwarnicke:
Weird... could this have been causing issues for IPv4 as well?

haiodo:
Nice.

PR #1220, opened 2019-06-10
Extract healing processing to different goroutine

Bolodya1997:
## Description
Creates a special goroutine for the healing processing.

## Motivation and Context
At the moment, all healing processing starts from the model / monitoring event listeners. It freezes processing subsequent events and leads to huge non-readable stacktrace in case of error.

Now there is a special goroutine to perform all healing processing. Listeners only send events to start healing.

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


Bolodya1997:
Depends on #1189.

controlplane/pkg/nsm/nsm_heal_processor.go
```diff
@@ -16,47 +16,141 @@ import (
...
+		cc:        cc,
+		healState: healState,
+	}
+}
+
+func (p *healProcessor) serve() {
+	for {
+		e := <-p.eventCh
+		if e.cc == nil {
+			return
```
lobkovilya: Should we stop the healing goroutine? Maybe `continue` is more appropriate in that case?

Bolodya1997: Changed to:
```
e, ok := <-p.eventCh
if !ok {
    return
}
```

PR #1229, opened 2019-06-11
remove node selection going to master

hwinkel:
hardcoded nodeselection to master might fail in all kubernetes services

## Description
removed the node selection spec.

## Motivation and Context
Deployment of a helm chart on managed k8s platforms will fail.


nickolaev:
Superseded by #1236

PR #1239, opened 2019-06-13
Refactor controlplane/pkg/apis/nsm

Bolodya1997:
## Description
Extend unified interfaces for local/remote Connection, NetworkServiceRequest. Replace implementation with interface where it is possible.

## Motivation and Context
Local/remote Connection, NetworkServiceRequest are almost identical and there are a lot of places with duplicate code for these classes.

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


.golangci.yml
```diff
@@ -111,5 +111,8 @@ issues:
     - path: controlplane/pkg/monitor/
       linters:
         - dupl
+    - path: controlplane/pkg/nsm.nsm.go
+      linters:
+        - gocyclo # TODO: reduce nsm.request() complexity
```
denis-tingajkin: Can you create an issue for this problem?

.golangci.yml
```diff
@@ -111,5 +111,8 @@ issues:
     - path: controlplane/pkg/monitor/
       linters:
         - dupl
```
denis-tingajkin: As I understood this PR should fixes code duplication. Can you remove this linter right?

controlplane/pkg/apis/crossconnect/helpers.go
```diff
@@ -1,6 +1,64 @@
...
+
+	return nil
+}
+
+// SetDestinationConnection sets cross connect destination connection
+func (c *CrossConnect) SetDestinationConnection(dst connection.Connection) {
+	if dst.IsRemote() {
+		c.Destination = &CrossConnect_RemoteDestination{
+			RemoteDestination: dst.(*remote.Connection),
+		}
```
denis-tingajkin: you can do return to avoid extra else block

controlplane/pkg/apis/nsm/connection/connection.go
```diff
@@ -0,0 +1,44 @@
...
+const (
+	// StateUp means that connection is up
+	StateUp State = 0
+	// StateDown means that connection is down
+	StateDown State = 1
+)
+
+// Connection is an unified interface for local/remote connections
+type Connection interface {
+	IsRemote() bool
```
denis-tingajkin: consider about extracting interfaces from this interface

controlplane/pkg/apis/remote/connection/connection_helpers.go
```diff
@@ -0,0 +1,145 @@
+// Copyright 2018 VMware, Inc.
```
denis-tingajkin: if the file changed the header should be changed to 2018-2019. @nickolaev correct me if I wrong

controlplane/pkg/nsm/nsm_mechanims.go
```diff
@@ -4,88 +4,71 @@ import (
...
-}
 
-func findRemoteMechanism(mechanismPreferences []*remote_connection.Mechanism, mechanismType remote_connection.MechanismType) *remote_connection.Mechanism {
-	for _, m := range mechanismPreferences {
-		if m.Type == mechanismType {
-			return m
-		}
-	}
-	return nil
+	return nil, fmt.Errorf("failed to select mechanism, no matched mechanisms found")
```
denis-tingajkin: `NSM:(5.1-%v) Failed to select mechanism. No matched mechanisms found...`

.golangci.yml
Bolodya1997: Here it is: #1241.

controlplane/pkg/apis/crossconnect/helpers.go
Bolodya1997: For me the left one looks better :)
```
if dst.IsRemote() {                                   | if dst.IsRemote() {
    c.Destination = &CrossConnect_RemoteDestination{  |     c.Destination = &CrossConnect_RemoteDestination{
        RemoteDestination: dst.(*remote.Destination), |         RemoteDestination: dst.(*remote.Connection),
    }                                                 |     }
} else {                                              |     return
    c.Destination = &CrossConnect_LocalDestination{   | }
        LocalDestination: dst.(*local.Connection),    | c.Destination = &CrossConnect_LocalDestination{
    }                                                 |     LocalDestination: dst.(*local.Connection),
}                                                     | }
```

controlplane/pkg/apis/nsm/connection/connection.go
Bolodya1997: I am not actually sure if there is a case when we can use such interface:
```
type ApiObject interface {
    IsRemote() bool
    Equals(other ApiObject)
    Clone() ApiObject
}
```
Can you please share your ideas about this?

denis-tingajkin: As you can see methods `IsRemote()`, `IsValid() error`, `IsComplete() error` are used in couple of places. The main idea considers creating share interfaces for these interfaces.
Something like:
```
type ShareInterace interface{
    IsRemote() bool
    IsValid() error
}
```
```
type Connection interface {
    ShareInterface
    ...
}
```
```
type Mechanism interface{
    ShareInterface
    ...
}
```

controlplane/pkg/apis/remote/connection/connection_helpers.go
Bolodya1997: Fixed : 3db2edb

controlplane/pkg/nsm/nsm_mechanims.go
Bolodya1997: Fixed on the upper level with:
```
return nil, fmt.Errorf("NSM:(5.1-%v) %v", requestID, err)
```

controlplane/pkg/apis/nsm/connection/connection.go
Bolodya1997: If we use such methods like `IsRemote()`, `IsValid()` we mostly want to distinguish the type of the passed value if it is `Connection`, or `Request`, or something other.

On the other hand, there are only few places where we use those mentioned methods. For the rest of the project we need to use only getter and setter methods.
So if there is a point to narrow interfaces, I would rather prefer to do something like:
```
type Shared interface {
    IsRemote() bool
    IsValid() error
}

type Connection interface {
    Equals(connection Connection) bool
    Clone() Connection

    // getter and setter methods
}

type ExtendedConnection interface {
    Shared
    Connection
}
```

But for me it looks like overcomplicating the code instead of making it better.

controlplane/pkg/apis/local/networkservice/helpers.go
```diff
@@ -5,41 +5,82 @@ import (
...
+}
+
+// GetRequestMechanismPreferences returns request mechanism preferences
+func (ns *NetworkServiceRequest) GetRequestMechanismPreferences() []connection.Mechanism {
+	preferences := make([]connection.Mechanism, 0, len(ns.MechanismPreferences))
+	for _, m := range ns.MechanismPreferences {
+		preferences = append(preferences, m)
+	}
+
+	return preferences
```
denis-tingajkin: return append([]connection.Mechanism{}, ns.MechanismPreferences)

controlplane/pkg/apis/local/networkservice/helpers.go
```diff
@@ -5,41 +5,82 @@ import (
...
+}
+
+// SetRequestMechanismPreferences sets request mechanism preferences
+func (ns *NetworkServiceRequest) SetRequestMechanismPreferences(mechanismPreferences []connection.Mechanism) {
+	preferences := make([]*local.Mechanism, 0, len(mechanismPreferences))
+	for _, m := range mechanismPreferences {
+		preferences = append(preferences, m.(*local.Mechanism))
```
denis-tingajkin: is missing check on type?

controlplane/pkg/apis/local/networkservice/helpers.go
Bolodya1997: In Go a **slice of type** is not a **slice of interface**, so we need to manually create a new slice and copy all the elements one by one.

controlplane/pkg/apis/local/networkservice/helpers.go
Bolodya1997: `m.(*local.Mechanism)` will panic if `m` is not a `*local.Mechanism`. So it is actually the check.

Issue #1215, opened 2019-06-07
NSMD fails to restart if endpoint networkservice has been deleted from k8s registry


Bolodya1997:
## Expected Behavior

NSMD should restart and restore existing endpoints and networkservices even if networkservice has been deleted from k8s registry.

## Current Behavior

NSMD fails to restart in such case.

## Steps to Reproduce

1. `make k8s-infra-deploy`
2. `make k8s-icmp-deploy`
3. `make k8s-nsmgr-delete`
4. `kubectl -n nsm-system delete networkservice icmp-responder`
5. `make k8s-nsmgr-deployonly` <-- NSDM fails to start

## Failure Logs

```
time="2019-06-06T04:43:30Z" level=info msg="Using Jaeger host/port: 10.105.203.227:6831"
2019/06/06 04:43:30 Initializing logging reporter
time="2019-06-06T04:43:30Z" level=info msg="Starting NSM gRPC server listening on socket: /var/lib/networkservicemesh/nsm.io.sock"
time="2019-06-06T04:43:30Z" level=info msg="Requesting NsmRegistryClient..."
time="2019-06-06T04:43:30Z" level=info msg="Waiting for liveness probe: tcp:127.0.0.1:5000"
time="2019-06-06T04:43:30Z" level=info msg="Registry port now available, attempting to connect..."
time="2019-06-06T04:43:30Z" level=info msg="Successfully connected to 127.0.0.1:5000"
time="2019-06-06T04:43:30Z" level=info msg="Requesting ClusterInfoClient..."
time="2019-06-06T04:43:30Z" level=info msg="Waiting for liveness probe: tcp:127.0.0.1:5000"
time="2019-06-06T04:43:30Z" level=info msg="Registry port now available, attempting to connect..."
time="2019-06-06T04:43:30Z" level=info msg="Successfully connected to 127.0.0.1:5000"
2019/06/06 04:43:30 Reporting span 6f543696b220b478:6f543696b220b478:0:1
2019/06/06 04:43:30 Reporting span 48106cbe5b5a1f6d:48106cbe5b5a1f6d:0:1
2019/06/06 04:43:30 Reporting span 68f1a2a272211419:68f1a2a272211419:0:1
time="2019-06-06T04:43:30Z" level=error msg="failed to set local NSM Failed to get list of own Endpoints: rpc error: code = Unknown desc = no NetworkService with name: icmp-responder"
[...]

PR #1248, opened 2019-06-17
Refactor NSE restore on NSM restart

Bolodya1997:
## Description
Refactor NSE restore on NSM restart.

## Motivation and Context
Issue #1215.

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/controlplane/pkg/nsmd/nsmd.go
+++ b/controlplane/pkg/nsmd/nsmd.go
@@ -170,120 +170,180 @@ func (nsm *nsmServer) EnumConnection(context context.Context, request *nsmdapi.E
 	return &nsmdapi.EnumConnectionReply{Workspace: workspaces}, nil
 }
 
-func (nsm *nsmServer) restoreClients(registeredEndpoints *registry.NetworkServiceEndpointList) {
-
+func (nsm *nsmServer) restore(registeredEndpointsList *registry.NetworkServiceEndpointList) {
 	if os.Getenv(NsmdDeleteLocalRegistry) == "true" {
 		logrus.Errorf("Delete of local nse/client registry... by ENV VAR: %s", NsmdDeleteLocalRegistry)
 		nsm.localRegistry.Delete()
 	}
+
 	clients, nses, err := nsm.localRegistry.LoadRegistry()
 	if err != nil {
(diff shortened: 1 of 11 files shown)


k8s/pkg/apis/networkservice/v1alpha1/types.go
```diff
@@ -61,6 +61,7 @@ type NetworkServiceEndpoint struct {
 
 type NetworkServiceEndpointSpec struct {
 	NetworkServiceName string `json:"networkservicename"`
+	Payload            string `json:"payload"`
```
lobkovilya: Why do we need `Payload` here? As far as I know we already have one in `NetworkService.Spec`

controlplane/pkg/nsmd/nsmd.go
```diff
@@ -170,120 +170,180 @@ func (nsm *nsmServer) EnumConnection(context context.Context, request *nsmdapi.E
...
+	discoveryClient registry.NetworkServiceDiscoveryClient,
+	registryClient registry.NetworkServiceRegistryClient,
+	name string,
+	nse nseregistry.NSEEntry,
+	ws *Workspace,
+	registeredNSEs map[string]string,
+	networkServices map[string]bool) (string, nseregistry.NSEEntry, error) {
+
+	if networkService, ok := registeredNSEs[name]; ok {
+		if networkServices[networkService] {
```
lobkovilya: If we have 2 NSE that implement the same NS, and the second NSE is absent in registry. Does it mean that we will not register the second NSE because of this if condition?

k8s/pkg/apis/networkservice/v1alpha1/types.go
Bolodya1997: There can be an issue when the `NSE` resource is in registry, but the corresponding `networkservice` resource has been deleted.

controlplane/pkg/nsmd/nsmd.go
Bolodya1997: `registeredNSEs` are `NSE` resources from registry. We only need to reregister such NSE if there is no corresponding `networkservice` resource in registry.

PR #1249, opened 2019-06-17
Move the static analysis class to large

nickolaev:
This shoudl make the target pass

## Motivation and Context
Static analysis is failing.

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,7 +43,7 @@ jobs:
 
 # code analysis
   test-static-code-analysis:
-    resource_class: medium+
+    resource_class: large
     docker:
       - image: circleci/golang
     steps:


denis-tingajkin:
In my opinion, It's bad that the linter takes many resources. Can we use a more stable version 1.16.0?
Another possible solution: we can use the default preset of linters to reduce resource usage.

nickolaev:
@denis-tingajkin I see what you say, yet I personally prefer to have "more" checks even at the cost of higher resources. Are you saying that 1.7.1 shall be considered unstable?  IMO it is just enabling more and more things to check, and that should be good for the project.

PR #1217, opened 2019-06-08
WIP: Add ARM64 build

rdimitrov:
Signed-off-by: Radoslav Dimitrov <dimitrovr@vmware.com>

## Description

Related to #809 
## Motivation and Context

A draft effort of having ARM64 image builds within the CI. Note that it is not tested.

## How Has This Been Tested?

- [ ] Covered by existing integration testing
- [ ] Added integration testing to cover
- [ ] Tested locally
- [x] Have not tested

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

Issue #809, opened 2019-03-06
Building and run NSM on ARM64


stanislav-chlebec:
We're tracking here the efforts to run NSM on an ARM64 board. A couple of issues have been observed so far, mainly with the way we use `vpp/vppagent`. The target is to have a minimal NSMgr + Dataplane deployment, which can accommodate NetworkServices Endpoints and Clients.

#828 treats the basic image build (no cross compilation).

-------------------------------------------------------------------------------------
Original text:
Hello 
I am Stanislav Chlebec from Pantheon Tech company.
I have built the set of networkservicemesh docker images for AMR64 platform.
Could you please allow me to push these images to your repo?
Thanks

edwarnicke:
@stanislav-chlebec Is it possible for us to build them as part of our CI and push them ourselves?

stanislav-chlebec:
@edwarnicke  Good idea. I will prepare and upload scripts. But currently I build on ARM64 platform. I have no scripts for crosscompilation....

nickolaev:
@stanislav-chlebec should be pretty easy to do this form our CI. We need QEMU arm static to do the translation, and the build will be pretty slow. Check this link:
https://ownyourbits.com/2018/06/27/running-and-building-arm-docker-containers-in-x86/

I suggest we put this in a separate job and call it only for `master` similar to our `docker-push-latest` job https://github.com/networkservicemesh/networkservicemesh/blob/master/.circleci/config.yml#L364


I would be interested in your environment, what machines do you use and what you were able to test/try? Did you manage to figure out the download of a working VPP image or are you building it yourself?

Our CI runs in Packet where we can potentially get ARM hosts too, although probably we'll postpone this after the release in May.

rdimitrov:
Hey, @stanislav-chlebec
Nice effort 👍 I also built the NSM images for ARM64 before and it went without any issues. I faced some problems with the VPP related images though - there were no consistent images pushed, so had to build one from source. It would be interesting to share how you manage to overcome that and run NSM at the end? 

As far as building on non-ARM64 hardware, if everything is pure Go perhaps we should be able to use the native cross-compilation of `go build` for ARM64, for ex. 
```
GOOS=linux GOARCH=arm64 go build
```

stanislav-chlebec:
@nickolaev environment: ThunderX ARM64
I only begin with this project - I was able to build docker images via make k8s-save command.
[...]

PR #1264, opened 2019-06-19
[WIP] Arm64 cross build

stanislav-chlebec:
## Description
I took PR https://github.com/networkservicemesh/networkservicemesh/pull/1217
and continued to work.
It was edited to incorporate previous changes in docker images content (some docker images became obsolete)
It was fixed errors caused by using ARM64 images in x86_64 environment by incorporating qemu-aarch-static library
It allows to build arm64 docker images on x86_64 platform using make

## Motivation and Context
Why is this change required? What problem does it solve? 
Building of ARM64 images inside of NSM CI environment on x86_64 platform

If it fixes an open issue, please link to the issue here.
https://github.com/networkservicemesh/networkservicemesh/pull/1217
https://github.com/networkservicemesh/networkservicemesh/issues/809

## How Has This Been Tested?
- [x] Tested locally
- [ ] Have not tested

   The images were copied manually on two bare metal ARM64 servers. 
   Then integration tests were executed
   http://147.75.83.101:8080/view/NSM/job/N13%20make%20k8s-integration-tests/46/consoleText

## Types of changes

## Checklist:

Where I should document it?


.circleci/config.yml
```diff
@@ -545,6 +545,8 @@ jobs:
             export TAG="${COMMIT}"
             make docker-<< parameters.container >>-build
             make docker-<< parameters.container >>-push
+            BUILD_ARCH=arm64 make docker-<< parameters.container >>-build-arm64
+            BUILD_ARCH=arm64 make docker-<< parameters.container >>-push-arm64
```
nickolaev: Please consider parametrising the buil-container target with "arch" instead.

.mk/docker_arm64.mk
```diff
@@ -0,0 +1,69 @@
+# Copyright (c) 2018 Cisco and/or its affiliates.
```
nickolaev: 2019

.mk/vpp_agent.mk
```diff
@@ -1,16 +1,26 @@
...
 endif
-ifeq (${ARCH}, aarch64)
+ifeq (${BUILD_ARCH}, arm64)
   export VPP_AGENT=ligato/vpp-agent-arm64:v2.1.1
   export VPP_AGENT_DEV=ligato/dev-vpp-agent-arm64:v2.1.1
 endif
 
 .PHONY: docker-vppagent-dataplane-dev-build
-docker-vppagent-dataplane-dev-build: docker-vppagent-dataplane-build
+docker-vppagent-dataplane-dev-build: docker-vppagent-dataplane-build install-qemu
```
nickolaev: install qemu? why?

scripts/install-qemu-user-static.sh
```diff
@@ -0,0 +1,8 @@
+#/bin/bash
+dpkg -l | grep qemu-user-static
```
nickolaev: Please don't do that. I don't have `dpkg` or `qemu-aarch64` on my mac.

side-cars/build/Dockerfile.nsm-init-arm64
```diff
@@ -0,0 +1,17 @@
...
+ENV GO111MODULE=on
+
+RUN mkdir /root/networkservicemesh
+ADD ["go.mod","/root/networkservicemesh"]
+ADD ["./scripts/go-mod-download.sh","/root/networkservicemesh"]
+WORKDIR /root/networkservicemesh/
+RUN ./go-mod-download.sh
+
+ADD [".","/root/networkservicemesh"]
+RUN GOARCH=arm64 CGO_ENABLED=0 GOOS=linux go build -ldflags '-extldflags "-static"' -o /go/bin/nsm-init ./side-cars/cmd/nsm-init/init.go
```
nickolaev: Please consider reusing the existing DOcker files and parametrise the GOARCH. See how we do it for VPP version.

.mk/vpp_agent.mk
stanislav-chlebec: This is a check done on Radoslav Dimitrov's PR 
https://circleci.com/gh/networkservicemesh/networkservicemesh/66971?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link
see error while building NSM docker images which are based on ligato/vpp-agent-arm64...

Digest: sha256:77f8c05335843a07e276846e54b9f994b532b3b1a80855af48281e91892851cb
Status: Downloaded newer image for ligato/vpp-agent-arm64:v2.1.1
 ---> 291f13a4df3f
Step 14/17 : COPY --from=build /go/bin/vppagent-firewall-nse /bin/vppagent-firewall-nse
 ---> 732e1a2939f9
Step 15/17 : RUN rm /opt/vpp-agent/dev/etcd.conf; echo 'Endpoint: "0.0.0.0:9112"' > /opt/vpp-agent/dev/grpc.conf; echo "disabled: true" > /opt/vpp-agent/dev/linux-plugin.conf
 ---> Running in 4d6ce52b2731

It would mean that you want  to run rm command inside ARM64 container but it would end with an error because rm is built for aarch64 architecture....

That is the reason why I added everywhere where have to be some processing inside ARM64 docker image that qemu-aarch64-static binary.

scripts/install-qemu-user-static.sh
stanislav-chlebec: This should be run only in case that the architecture aarch64 is detected.

stanislav-chlebec: Is it possible to run mac on arm64?

side-cars/build/Dockerfile.nsm-init-arm64
stanislav-chlebec: Ok, I will try.

stanislav-chlebec:
Is it possible to run current PR in circleci?
I would like to see if it builds ARM64 docker images.

rdimitrov:
Hey, @stanislav-chlebec 
I'm starting to think that the cleanest approach would be to have a separate job in our CI where we can `build/push` the ARM64 images on **actual ARM64 hosts** provided by Packet. 

This way everything around the building should be fairly the same to x86 with the difference when doing the actual docker push, for ex. if we want to have something like `networkservicemesh/nsmd:latest-arm64`.

I'll let the others share as well if it's possible to get ARM64 hosts for our CI or we should proceed with the cross-building approach...

scripts/install-qemu-user-static.sh
stanislav-chlebec: You are right and I am wrong with that argument that it will be executed only on case that aarch64 is detected. 
But I am not sure how this kind of thing is solved on mac...

side-cars/build/Dockerfile.nsm-init-arm64
stanislav-chlebec: This is partly granted in commit 2c28e76 . I will finish the rest of Dockerfiles in the same style...

edwarnicke:
@rdimitrov Is the fake qemu not working?  Or is it just looking at being less intrusive?

rdimitrov:
@edwarnicke My idea is if there's a way to do it less complicated. Adding cross-compilation support using QEMU is an option, but I'm really wondering if we need it that much?

1. In case someone's trying something on ARM, the chances are he'll probably be either getting the Docker images from Dockerhub or building them on the ARM host, right? So not sure if we need to have such support. _Also, I know it is already available build-in through Docker Desktop(at least on macOS)_

2. What I'm saying is that it might be better if we have a CI job for building the ARM images natively using, for example, a "c1.large.arm" instance from Packet.
**Pros:** No need to add a lot of changes other than tagging the images for ARM.
**Cons:** A bit more expensive CI job + lack of support if you want to develop ARM stuff remotely on your x86 machine

3. Also, having an ARM-based CI job can be used later for running a basic set of integration tests on it.

So what are your thoughts on having a CI job using an ARM instance from Packet?

edwarnicke:
@rdimitrov Seems wise :)

stanislav-chlebec:
> 
> 
> @rdimitrov Is the fake qemu not working? Or is it just looking at being less intrusive?

Just to explain that "FAKE qemu"
If we have the same Dockerfile for both platform
and only one need some file which we will provide via docker COPY command
also other target have to COPY the file with the same name.
I simply did not want to instal needlesly qemu on x86_64 plaform so I provided som preprepared bash file

stanislav-chlebec:
QEMU elimination:
It is possible only in case when you are executing the command in ARM64 docker containers while preparing the target ARM64 docker container.
You can see that e.g. when preparing docker/Dockerfile.admission-webhook  there is no need for qemu but in case test/applications/build/Dockerfile.vpp-test-common  you need it because typically you are trying to run some command inside of ARM64 docker container.
If you manage to remove RUN clauses from Dockerfiles which are inside of supposedly ARM64 containers then there is no need to copy there qemu-aarch64-static...

stanislav-chlebec:
>     2. What I'm saying is that it might be better if we have a CI job for building the ARM images natively using, for example, a "c1.large.arm" instance from Packet.
>        **Pros:** No need to add a lot of changes other than tagging the images for ARM.
>        **Cons:** A bit more expensive CI job + lack of support if you want to develop ARM stuff remotely on your x86 machine

**Pros**
I tried to run integration tests with ARM64 natively built docker images and with ARM64 docker images bilt on x86_64 platform. The results were different - in case of docker images built on x86_64 it ended with failure earlier (with some o error) and less tests were properly ended. So I woukd prefer nativelly build ARM64 docker images

rdimitrov:
@stanislav-chlebec It's really nice that you compared the two image builds :thumbsup: 
Indeed, building everything natively would make things way easier. I'm interested to see how it'll come up at the end.

nickolaev:
I think this is a no go as it is implemented today. I will propose to put this on hold and rethink it again once we have dedicated hardware for testing.

nickolaev:
Closing this one as we can not use it. AMR64 is worth considering, but we have more pressing topics today.

PR #1265, opened 2019-06-19
Remove mentions of ROUTES in configuration

nickolaev:

## Description
There is some stale configuration about configuring routes.

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/sdk/README.md
+++ b/sdk/README.md
@@ -32,7 +32,6 @@ type NSConfiguration struct {
 	TracerEnabled      bool   // TRACER_ENABLED
 	MechanismType      string // MECHANISM_TYPE
 	IPAddress          string // IP_ADDRESS
-	Routes             []string // ROUTES
 }
 ```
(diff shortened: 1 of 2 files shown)


edwarnicke:
@nickolaev How are folks supposed to setup routes?

nickolaev:
So this code is a leftover from some experiments around having Routes in the config. We now have a new `NewCustomFuncEndpoint` composite which can be used to mutate the connection. We do it here:

https://github.com/networkservicemesh/networkservicemesh/blob/master/test/applications/cmd/icmp-responder-nse/main.go#L71

The point is, it is left to the implementator to figure out if they want Routes or Neighbors, it is not a core Endpoint configuration.

edwarnicke:
Got it :)  So basically improved modularity, which is good :)

PR #1267, opened 2019-06-20
Make kernel interface converter have a GW

nickolaev:
Plus make the supplied routes global and not LINK local

## Description
Make sure that when the vpp-agent injects the interfaces it also sets a GW for the supplied routes. Also, the routes are not really Link Local in the broader sense. The patch is inspired by the memif converter approach to routes.

## Motivation and Context
When a Linux (kernel) client get its routes it does not have a GW for them. Make that GW to be the DST IP in the link, effectively the endpoint.

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


edwarnicke:
Good catch!

Issue #1224, opened 2019-06-10
Spec: DNS Integration for NSM


edwarnicke:
## Overview

Network Service Mesh needs to be able to provide a workload with DNS service from Network Services Without breaking K8s DNS

## Link to the Google Docs document

[Spec](https://docs.google.com/document/d/1_1ICFEh2M3_Q8MvAQXNngP9UNQLgYgPgr58KYzib4C0/edit)

nickolaev:
To me, it is yes and no at the same time. Let me explain.

Yes - the problem with resolving service names once we get the link is real and certainly, this will be a huge use-case. Augmenting the pod with an additional local CoreDNS is a powerful idea and as discussed can benefit beyond NSM. For sure the results of our work and experimentation will be something to share across the broader community.

No - I don't like how we are baking more and more IPAM properties in our core API. We already have IP addresses, routes and now we add DNS? I mean, NSM was supposed to be protocol agnostic and I believe this should be a major virtue of it. I would prefer if we refactor `message ConnectionContext` to use `map <string, string>` and keep these abstracted. I know how protobuf optimizes the messages, and my point is more about concepts. Imagine if we have to add all the MPLS specifics or whatever future protocols we want to support.

One more point to note here. I prefer if our admission controller stays flexible. the monitoring client and the CoreDNS injector need be optional, possibly by adding an annotation. Then we'll probably have to figure out a way to maintain our set of side-cars, which is feasible of course.

So overall, great and exciting feature, but let's be careful with the implementation and not embed assumptions in the API.

edwarnicke:
@nickolaev I understand the desire to keep the API simple, and share it.

We originally started out with the map <string,string> for ConnectionContext.  It has two small issues that led to moving to the more descriptive API:

1. Its easy for implementors to get the strings for the map wrong
2. It gets harder to do more structures values as map<string,string> as you wind up putting formating into the string value.

Do you have thoughts on those two issues?

I also would like our admission controller to stay flexible.  On the other side we want to keep in mind:
1. Optimize for the average case
2. Keep dev interaction with the system simple

[...]

PR #1274, opened 2019-06-25
[WIP] Spec: DNS Integration for NSM

denis-tingajkin:
## Description
https://github.com/networkservicemesh/networkservicemesh/issues/1224

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


.circleci/config.yml
```diff
@@ -555,7 +555,17 @@ jobs:
           command: |
             echo "SUCCESS"
 
-# docker push
+  pull-coredns:
+    docker:
+      - image: circleci/golang
```
nickolaev: NIT Interesting, do we need to use the golang image here, just out of curiosity. I guess it just works.

controlplane/pkg/apis/connectioncontext/connectioncontext.proto
```diff
@@ -45,18 +45,36 @@ message ExtraPrefixRequest {
...
+    /* ips of DNS Servers for this DNSConfig.  Any given IP may be IPv4 or IPv6 */
+    repeated string dns_server_ips = 1;
+    /* domains for which this DNSConfig provides resolution.  If empty, all domains. */
+    repeated string resolves_domains = 2;
+    /* true if DNSConfig should be used to resolve listed domains listed in
+     * resolves_domains over other resolvers that did not request priority.  true is
+     * invalid if resolves_domains is empty.
+     */
+    bool prioritize = 3;
+}
```
nickolaev: Can you please move the ConnectionContext at the bottom here, just for clarity. Otherwise loks great IMHO ;)

controlplane/pkg/apis/local/connection/connection.proto
```diff
@@ -47,7 +47,7 @@ message Connection {
     string id = 1;
     string network_service = 2;
     Mechanism mechanism = 3;
-    connectioncontext.ConnectionContext context = 4;
+    connectioncontext.IPContext context = 4;
```
nickolaev: That is strange. Iwoudl very much vote for keeping the "ConnectionCOntext" here, as it name implies it will keep the context associated with this connection. Please consider reverting this and the related changes.

sdk/endpoint/compositeBuilder.go
```diff
@@ -0,0 +1,24 @@
+package endpoint
```
nickolaev: Please add licensing here.

sdk/endpoint/compositeBuilder.go
```diff
@@ -0,0 +1,24 @@
...
+
+type CompositeEndpointBuilder interface {
+	Append(...ChainedEndpoint) CompositeEndpointBuilder
+	Build() *CompositeEndpoint
+}
+
+func NewCompositeEndpointBuilder() CompositeEndpointBuilder {
+	return &compositeEndpointBuilder{
+		endpoints: make([]ChainedEndpoint, 0, 4),
+	}
```
nickolaev: Interesting, it surely makes buidling composites smoother. I am willing to accept this, but shallwe consider adding more methods to the interface: Insert and Remove fore example. Let's make these complete now and don't change it later.

test/applications/cmd/icmp-responder-dns-nse/main.go
```diff
@@ -0,0 +1,69 @@
...
+
+func dnsConfigMutator(c *connection.Connection) error {
+	dnsSidecarIp := strings.Split(c.Context.DstIpAddr, "/")[0] + ":" + fmt.Sprint(myCoreDnsPort)
+	c.Context.DnsConfig = &connectioncontext.DNSConfig{
+		DnsServerIps: []string{dnsSidecarIp},
+	}
+	return nil
+}
+
+//TODO: remove code duplication
```
nickolaev: Indeed, I see these mutators are getting popular. Shall we move them in the sdk/common?

sdk/endpoint/compositeBuilder.go
nickolaev: And well add these in all yuor new files.

test/applications/cmd/icmp-responder-dns-nse/main.go
```diff
@@ -0,0 +1,69 @@
...
+	// Capture signals to cleanup before exiting
+	c := tools.NewOSSignalChannel()
+	ipamEndpoint := endpoint.NewIpamEndpoint(nil)
+	routeAddr := makeRouteMutator([]string{"8.8.8.8/30"})
+	if common.IsIPv6(ipamEndpoint.PrefixPool.GetPrefixes()[0]) {
+		routeAddr = makeRouteMutator([]string{"2001:4860:4860::8888/126"})
+	}
+
+	composite := endpoint.NewCompositeEndpointBuilder().
+		Append(
```
nickolaev: That actually looks good.

controlplane/pkg/apis/local/connection/connection.proto
denis-tingajkin: Yep, I planning to use `ConnectionContext`. It's just for testing.

edwarnicke: +1 on making IPContext a subfield of ConnectionContext :)

controlplane/pkg/apis/connectioncontext/connectioncontext.proto
edwarnicke: +1 for moving ConnectionContext to the bottom :)

controlplane/pkg/apis/connectioncontext/helpers.go
```diff
@@ -30,7 +30,17 @@ func (c *ConnectionContext) IsValid() error {
 	return nil
 }
 
-func (c *ConnectionContext) MeetsRequirements(original *ConnectionContext) error {
+func (c *DNSConfig) Validate() error {
+	if c == nil {
+		return fmt.Errorf("dns config should be nil")
```
edwarnicke: Do you mean dns config should *not* be nil?

sdk/endpoint/compositeBuilder.go
edwarnicke: @nickolaev I wonder if we could lint for license headers?

sdk/endpoint/compositeBuilder.go
edwarnicke: @nickolaev Could you suggest an example of where we'd use Insert or Remove ?

test/applications/cmd/monitoring-dns-nsc/main.go
```diff
@@ -0,0 +1,91 @@
...
+				if conn.Context.DnsConfig != nil {
+					updateCoreFile(corefile, conn.Context.DnsConfig)
+				}
+				switch event.EventType() {
+				case monitor.EventTypeInitialStateTransfer, monitor.EventTypeUpdate:
+					if !proto.Equal(conn, currentConn) {
+						logrus.Infof("Connection updated: %v", conn)
+						currentConn = conn
+					}
+				case monitor.EventTypeDelete:
```
edwarnicke: Are we going to cleanup DNSConfig when a Connection is deleted?

test/applications/cmd/monitoring-dns-nsc/main.go
```diff
@@ -0,0 +1,91 @@
...
+					return
+				}
+			}
+		}
+	}
+}
+
+func updateCoreFile(corefile Corefile, config *connectioncontext.DNSConfig) {
+	forwards := strings.Join(config.DnsServerIps, " ")
+	corefile.Remove(".:53")
```
edwarnicke: Naively reading this, it looks like when we update the CoreFile we remove all existing DNS config and replace it with the DNSConfig from the latest connection?  Is this so?  What about the original DNS config from K8s?  What about the DNSConfig from the previous NSM Connections?  How do we parallelize amongst them?

controlplane/pkg/apis/connectioncontext/helpers.go
denis-tingajkin: Oops! I'll fix in next commit

test/applications/cmd/monitoring-dns-nsc/main.go
denis-tingajkin: > we remove all existing DNS config and replace it with the DNSConfig from the latest connection? 
Yes, you are right. I planning to add merging between configs in next commit.
> What about the original DNS config from K8s?
At this moment we not changing file `resolv.conf`. We just changing the corefile for coredns siedecar
So if our local coreDNS server is not know about the host, then default k8s-dns service will trying to resolve host(or some else from the `resolve.conf`). 
Is this behaviour wrong?
>What about the DNSConfig from the previous NSM Connections? 
We'll store all configs.
>How do we parallelize amongst them?
I think it's part of coredns. I'll investigate and provide more detail answer.

test/applications/cmd/monitoring-dns-nsc/main.go
denis-tingajkin: Yes, I'll fix it in next commit.

sdk/endpoint/compositeBuilder.go
denis-tingajkin: @edwarnicke actually, we have not the linter for license headers. We can create an issue for this problem and I can solve this in separate PR.

controlplane/pkg/apis/local/connection/connection.proto
denis-tingajkin: done

controlplane/pkg/apis/connectioncontext/helpers.go
denis-tingajkin: done

controlplane/pkg/apis/connectioncontext/connectioncontext.proto
denis-tingajkin: done

.circleci/config.yml
denis-tingajkin: done

test/applications/cmd/monitoring-dns-nsc/main.go
denis-tingajkin: done

test/applications/cmd/icmp-responder-dns-nse/main.go
denis-tingajkin: done

.circleci/config.yml
haiodo: I suppose for generic stuff we could use alpine, but I suppose any circleci/image should work fine :-)

controlplane/pkg/apis/connectioncontext/connectioncontext.proto
```diff
@@ -45,18 +45,36 @@ message ExtraPrefixRequest {
...
 
-    repeated ExtraPrefixRequest extra_prefix_request = 8;   /* A request for NSE to provide extra prefixes */
+    repeated ExtraPrefixRequest extra_prefix_request = 8; /* A request for NSE to provide extra prefixes */
     repeated string extra_prefixes = 9; /* A list of extra prefixes requested */
-}
\ No newline at end of file
+}
+
+
+message DNSConfig {
```
haiodo: I suppose we should name it DNSContext? Suggestions? @edwarnicke @nickolaev?

controlplane/pkg/apis/connectioncontext/connectioncontext.proto
```diff
@@ -45,18 +45,36 @@ message ExtraPrefixRequest {
...
+    /* ips of DNS Servers for this DNSConfig.  Any given IP may be IPv4 or IPv6 */
+    repeated string dns_server_ips = 1;
+    /* domains for which this DNSConfig provides resolution.  If empty, all domains. */
+    repeated string resolves_domains = 2;
```
haiodo: Should be be search_domains instead? @edwarnicke  @nickolaev

controlplane/pkg/apis/local/connection/connection_helpers.go
```diff
@@ -55,7 +55,7 @@ func (c *Connection) SetContext(context *connectioncontext.ConnectionContext) {
 
 // UpdateContext checks and tries to set connection context
 func (c *Connection) UpdateContext(context *connectioncontext.ConnectionContext) error {
-	if err := context.MeetsRequirements(c.Context); err != nil {
+	if err := context.IpContext.MeetsRequirements(c.Context.IpContext); err != nil {
```
haiodo: I suppose we need to have MeetsRequirements() for top level context, and delegate to individuals?

scripts/kind.yaml
```diff
@@ -1,6 +1,6 @@
 kind: Cluster
 apiVersion: kind.sigs.k8s.io/v1alpha3
 nodes:
-- role: control-plane
```
haiodo: Do we need this changes here?

dataplane/vppagent/pkg/vppagent/vppagent.go
```diff
@@ -16,7 +16,10 @@ package vppagent
 
 import (
 	"context"
+	"github.com/opentracing/opentracing-go"
```
haiodo: Do we need this changes here?

sdk/endpoint/compositeBuilder.go
haiodo: +1

test/applications/cmd/icmp-responder-dns-nse/main.go
haiodo: Should we better consider to have Build() accept a list of endpoints?
```go
endpoint.NewCompositeEndpoint().Build(endpoint.NewMonitortEndpoint(nil), ..., ipadEndpoint, endpoint.NewConnectionEndpoint(nil))
```
Or even:
```go
endpoint.BuildCompositeEndpoint(endpoint.NewMonitortEndpoint(nil), ..., ipadEndpoint, endpoint.NewConnectionEndpoint(nil))
```

@nickolaev Thoughts?

test/applications/cmd/monitoring-dns-nsc/corefile/corefile.go
```diff
@@ -0,0 +1,196 @@
+package corefile
```
haiodo: @denis-tingajkin did you try to search existing library for this?

denis-tingajkin: @haiodo yes, I tried to search and did not find anything similar.

scripts/kind.yaml
denis-tingajkin: oops, it's a typo. I'll fix in next commit, thanks

k8s/cmd/coredns/plugin/fanout/fanout.go
```diff
@@ -0,0 +1,108 @@
...
+					}
+					return
+				}
+				if err != nil {
+					agent.healthcheck()
+				}
+			}
+		}()
+	}
+	var first connectResult
```
john-a-joyce: This is less a comment on the code and more a question on the spec.   If I am reading correctly the local coredns server in the sidecar always has strict priority over requests since the resolv.conf always points first to it then to the kube-dns server.   So if there was ever a case where a service with the same name was registered in both k8s and in the DNS server the connection points to then the server from the NSM connection would provide the response.   Is this the correct understanding?

k8s/cmd/coredns/plugin/fanout/fanout.go
```diff
@@ -0,0 +1,108 @@
...
+	select {
+	case <-time.After(defaultTimeout):
```
john-a-joyce: related to the above comment (same service name in kube-dns and connection DNS server) is there a possible race condition that the app/pod will timeout on the request to the local coredns server and attempt to use the kube-dns server because the queries to coreDNS server will initially fail while this code runs to populate the coredns records.   I don't see any sign that the app/pod requests are blocked until this code has had the chance to timeout.  It seems that might be needed to prevent the race condition.

k8s/cmd/coredns/plugin/fanout/fanout.go
denis-tingajkin: >If I am reading correctly the local coredns server in the sidecar always has strict priority over requests since the resolv.conf always points first to it then to the kube-dns server. 
Actually, we doing request only to local coredns. It fanout request to all known nameservers inclues default k8s name server. The first positive response is answer to request.

denis-tingajkin:
This will split in three parts.

k8s/cmd/coredns/plugin/fanout/fanout.go
john-a-joyce: Ok looking closer I see that the implemetnation of replaceDNSconfig doesn't match what was described in the doc.   There probably needs to be some deterministic way to decide if the DNS form the connections or DNS from K8s should take priority if both return matches.   The K8s DNS is automatic and not easily switched off IMO it is problematic if there isn't some way to specify priority vs. the first server to respond.   If I am missing some aspect let me know. 

Sine this is closed I guess I need to repeat in another PR when I see it

denis-tingajkin: So, as I understand the problem is next scenario:

1) local coredns know about two DNS Servers
1.1) DNS Server1: default k8s DNS server
1.2) DNS Server2: user's DNS server
2) DNS Server1 and DNS Server2 know about the host with name `www.google.com`
3) The user wants that his DNS Server always answers the question related to `www.google.com`
4) In this case, the user just simple add search domain to your `DnsConfig` (look at the proto message into Spec)
It will generate next corefile
```
. {
	fanout {Default k8s DNS Server IP} {User's DNS Server IP}
}
www.google.com {
	hosts {
		 {User's DNS Server IP} www.google.com
	}
}
```
This config will by priority sents all queries about `www.google.com` to {User's DNS Server IP}

denis-tingajkin: Let me know if I missed something

Issue #1152, opened 2019-05-22
usecase_vpn_test tests are failing inconsistently


Bolodya1997:
## Expected Behavior

All tests from `usecase_vpn_test.go` should always pass.

## Current Behavior

These tests are inconsistently failing on GKE and packet:
* TestVPNLocal
* TestVPNFirewallRemote
* TestVPNNSERemote
* TestVPNNSCRemote

## Failure Information (for bugs)
https://circleci.com/gh/networkservicemesh/networkservicemesh/53948
https://circleci.com/gh/networkservicemesh/networkservicemesh/54094
https://circleci.com/gh/networkservicemesh/networkservicemesh/54051
https://circleci.com/gh/networkservicemesh/networkservicemesh/54211

Bolodya1997:
https://circleci.com/gh/networkservicemesh/networkservicemesh/54512
https://circleci.com/gh/networkservicemesh/networkservicemesh/54352

Bolodya1997:
Still happening sometimes:
https://circleci.com/gh/networkservicemesh/networkservicemesh/64763

lobkovilya:
Failed because sibling packet job has failed and started deleting all packet clusters:
https://circleci.com/gh/networkservicemesh/networkservicemesh/53948
https://circleci.com/gh/networkservicemesh/networkservicemesh/54051
https://circleci.com/gh/networkservicemesh/networkservicemesh/54352

Failed because of 30m timeout exceeded:
https://circleci.com/gh/networkservicemesh/networkservicemesh/54094
https://circleci.com/gh/networkservicemesh/networkservicemesh/54211
https://circleci.com/gh/networkservicemesh/networkservicemesh/54512

Bolodya1997:
Still failing:
https://circleci.com/gh/networkservicemesh/networkservicemesh/69695

PR #1299, opened 2019-07-01
Fix usecase_vpn_test tests are failing inconsistently

denis-tingajkin:
## Motivation and Context
https://github.com/networkservicemesh/networkservicemesh/issues/1152

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/test/integration/usecase_vpn_test.go
+++ b/test/integration/usecase_vpn_test.go
@@ -125,8 +125,8 @@ func testVPN(t *testing.T, ptnum, nodesCount int, affinity map[string]int, verbo
 	}
 
 	pingCommand := "ping"
-	addressPool := "10.60.1.0/24"
-	srcIP, dstIP := "10.60.1.1", "10.60.1.2"
+	addressPool := "172.16.1.0/24"
+	srcIP, dstIP := "172.16.1.1", "172.16.1.2"
 
 	/* Change stuff related to IPv6 */
 	if k8s.UseIPv6() {


nickolaev:
I think that increasing the timeouts shoudl fix it, like this:

https://github.com/networkservicemesh/networkservicemesh/pull/1269/commits/21740140690ee8da4eb75e9eff1d63c1beeaf975

test/integration/usecase_vpn_test.go
```diff
@@ -125,8 +125,8 @@ func testVPN(t *testing.T, ptnum, nodesCount int, affinity map[string]int, verbo
 	}
 
 	pingCommand := "ping"
-	addressPool := "10.60.1.0/24"
-	srcIP, dstIP := "10.60.1.1", "10.60.1.2"
+	addressPool := "172.16.1.0/24"
+	srcIP, dstIP := "172.16.1.1", "172.16.1.2"
```
nickolaev: @denis-tingajkin can you elaborate in the PR body why this is the fix?

haiodo: Also could we check this dynamically for test to choose exclude prefixes for cluster we are running on?

haiodo: Ok discussed, according to @lobkovilya we already have tests with same values, and in future we could think about moving to dynamic obtaining of this values.

Issue #1296, opened 2019-07-01
Default make target fails


fkautz:
## Expected Behavior

$ make
[images built]

## Current Behavior

Some images build, debug devenv image fails

## Failure Information (for bugs)
```sh
$ make
...
Step 15/17 : COPY ["./docker/debug/dev-entry.go","/go/src/"]
COPY failed: stat /var/lib/docker/tmp/docker-builder626786555/docker/debug/dev-entry.go: no such file or directory
make: *** [.mk/docker.mk:79: docker-devenv-build] Error 1
```

## Steps to Reproduce

`make`

## Context

Failed on both ubuntu and OSX

## Failure Logs

```sh
$ make
...
Step 15/17 : COPY ["./docker/debug/dev-entry.go","/go/src/"]
COPY failed: stat /var/lib/docker/tmp/docker-builder626786555/docker/debug/dev-entry.go: no such file or directory
make: *** [.mk/docker.mk:79: docker-devenv-build] Error 1
```

denis-tingajkin:
@fkautz I've tested and `make` works fine after fix. Can you verify this?

PR #1300, opened 2019-07-01
Fix default make target fails 

denis-tingajkin:
## Motivation and Context
https://github.com/networkservicemesh/networkservicemesh/issues/1296

## How Has This Been Tested?

- [X] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


nickolaev:
Are you sure you fix the same issue here?

.circleci/config.yml
```diff
@@ -191,6 +201,12 @@ workflows:
           requires:
             - test-sanity-check
 # build
+      - build-container:
```
haiodo: Why we need devenv here? It only for local development.

denis-tingajkin: >Why we need devenv here? 
I've added this for check that `devenv` builds fine to avoid next regerssions

Are you have any idea how we can check this docker file without building on CI?

haiodo: I suppose we do not need to check it on CI, since this image are only used in local builds. I suppose we could excluded it from make k8s-save builds also.

Issue #1280, opened 2019-06-25
Create VppAgentAcl in SDK


edwarnicke:
## Overview

Create VppAgentAcl to simply apply ACLs to the incoming connection.  It should get the DataChange from the Opaque data if it exists, get the vppagent name of the Endpoint interface being connected from the Opaque data, and write the ACLs to the DataChange in Opaque data.

Issue #1279, opened 2019-06-25
Create VppAgentXCon in SDK


edwarnicke:
## Overview

Create a VppAgentXCon in SDK to cross connect the results of VppAgentClientConnect (#1278) and VppAgentMemifConnect (#1275) (similar to what is done in [the secure-intranet-example](https://github.com/edwarnicke/examples/blob/master/examples/secure-intranet/vppagent-endpoint/cmd/composite.go#L36)

onobrod:
If I got the task right, we want to have few small composites which share one DataChange object and append it with parts of config (MemifConnect and ClientConnect appends source and destination interfaces, XConnect appends XconnectPairs). It brings a question how to properly share DataChange object between composites (currently each composite only share its opaque data, it is kinda different thing), but it is doable, I think. What bothers me more is that we can create composites which are coupled a lot and should be used in a very specific order (ClientConnect should be called just after NewClientEndpoint to get its opaque data, then MemifConnect should be called, then XConnect). Why is it a bad idea to have one XConnect which calls [CrossConnectConverter](https://github.com/networkservicemesh/networkservicemesh/blob/master/dataplane/vppagent/pkg/converter/cross_connect_converter.go) from dataplane? Do we want to have an opportunity to reuse ClientConnect and XConnect separately without other VppAgent* compoistes? Could you please explain more on this?

edwarnicke:
A few points of modularity:

We don't want to couple every NSE to the vpp-agent dataplane code.  It does a specific and particular thing, and we don't want changes made there that serve the vpp-agent dataplane to break the SDK.

Yes, there is a particular order, but by breaking things into composites you allow the replacement of particular pieces of the chain, and insertion of new behaviors between pieces of the chain.  

As to sharing the DataChange... GetOpaque was designed to allow carrying that kind of object along the chain.

Issue #1278, opened 2019-06-25
Create VppAgentClientConnect in SDK


edwarnicke:
## Overview

This is a continuation of #1275 and #1276.  Create a VppAgentClientConnect in SDK suitable for use with endpoint.NewClientEndpoint to connect the Client to the Vppagent as an interface.  It should get the DataChange from Opaque Data (if it exists) and add the Client Connection to that DataChange.

Issue #1276, opened 2019-06-25
Create VppAgentFlush in SDK


edwarnicke:
## Overview

#1275 specifies the first of several VppAgent* CompositeEndpoints to be used for basic VppAgent functionality.  It intentionally does not Flush the DataChange to the vppagent.

VppAgentFlush should be a CompositeEndpoint that looks for the DataChange in the Opaque data and writes it to vpp-agent.

PR #1304, opened 2019-07-02
Add new composites into /sdk/vppagent

onobrod:
## Description
Added vppagent package into sdk. Added MemifConnect, Flush, ClientMemifConnect and XConnect composites.
- **MemifConnect** creates a DataChange with Memif interface and stores it in opaque data.
- **Flush** reads DataChange from opaque and writes it into vpp-agent.
- **ClientMemifConnect** reads created outgoing connection from opaque data, creates a DataChange with Memif interface for this connection and stores it in opaque data.
- **XConnect** reads DataChange from opaque and appends it with cross connect configuration.
- **ACL** applies ACL to the incoming connection.

## Motivation and Context
https://github.com/networkservicemesh/networkservicemesh/issues/1275 https://github.com/networkservicemesh/networkservicemesh/issues/1276 https://github.com/networkservicemesh/networkservicemesh/issues/1278 https://github.com/networkservicemesh/networkservicemesh/issues/1279 https://github.com/networkservicemesh/networkservicemesh/issues/1280

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


onobrod:
This PR depends on #1303. Please, use the following link to review it: https://github.com/networkservicemesh/networkservicemesh/pull/1304/files/7f1ae9a22e7e4bfe6f155b55028e41909d98c794..4367d09f0631a31a3d7801c83152f863177e84a8

sdk/vppagent/client_connect.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	opaque := cc.GetNext().GetOpaque(incomingConnection)
+	if opaque == nil {
+		err := fmt.Errorf("received empty data from Next")
+		logrus.Errorf("Unable to find the outgoing connection: %v", err)
+		return nil, err
+	}
+	outgoingConnection := opaque.(*connection.Connection)
+	incomingConnection.Context = outgoingConnection.GetContext()
```
onobrod: Not sure when this should be done. Only ClientConnect has access to the outgoing connection object, but it looks more like cross connect stuff, doesn't it?

sdk/vppagent/client_connect.go
```diff
@@ -0,0 +1,106 @@
...
+
+	interfaceName := "DST-" + outgoingConnection.GetId()
+
+	dataChange := &configurator.Config{
+		VppConfig: &vpp.ConfigData{
+			Interfaces: []*interfaces.Interface{
+				{
+					Name:    interfaceName,
+					Type:    interfaces.Interface_MEMIF,
```
onobrod: If we use Memif for client connection here, should we rename ClientConnect to MemifClientConnect?

nickolaev:
Good job. Please consider adding substantial documentation, preferably in a separate README under `sdk/`.

sdk/vppagent/client_connect.go
```diff
@@ -0,0 +1,106 @@
...
+	opaque := cc.GetNext().GetOpaque(incomingConnection)
+	if opaque == nil {
+		err := fmt.Errorf("received empty data from Next")
+		logrus.Errorf("Unable to find the outgoing connection: %v", err)
+		return nil, err
+	}
+	outgoingConnection := opaque.(*connection.Connection)
+	incomingConnection.Context = outgoingConnection.GetContext()
+
+	interfaceName := "DST-" + outgoingConnection.GetId()
```
nickolaev: `DST-` is probably meaningless here.

sdk/vppagent/client_connect.go
nickolaev: This is a bit binding to the specific interface. Can we rename this struct to `ClientConnectMemIf` and then have one more which is `ClientConnectKernel`?

sdk/vppagent/cross_connect.go
```diff
@@ -0,0 +1,105 @@
...
+		return nil, err
+	}
+	connectionData := opaque.(*ConnectionData)
+
+	if connectionData.DstName == "" {
+		err := fmt.Errorf("found empty destination name")
+		logrus.Errorf("Invalid connection data: %v", err)
+		return nil, err
+	}
+	if connectionData.SrcName == "" {
```
nickolaev: Src and Dst come from our forwarding plane concepts. Maybe we can reconsider these - In and Out or something esle.

sdk/vppagent/flush.go
```diff
@@ -0,0 +1,134 @@
...
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(context.Background(), 120*time.Second)
+		defer cancel()
+	}
+	if err := tools.WaitForPortAvailable(ctx, "tcp", f.Endpoint, 100*time.Millisecond); err != nil {
+		return nil, err
+	}
+
+	tracer := opentracing.GlobalTracer()
+	conn, err := grpc.Dial(f.Endpoint, grpc.WithInsecure(),
```
nickolaev: I am not a `vppagent` expert, but can we have a single opened grcpConnection and just reuse it here? LIke create it in `NewFlush` and keep it as a struct field, so you can just do `f.conn`

sdk/vppagent/flush.go
```diff
@@ -0,0 +1,134 @@
...
+}
+
+func (f *Flush) reset() error {
+	ctx, cancel := context.WithTimeout(context.Background(), 120*time.Second)
+	defer cancel()
+	if err := tools.WaitForPortAvailable(ctx, "tcp", f.Endpoint, 100*time.Millisecond); err != nil {
+		return err
+	}
+	tracer := opentracing.GlobalTracer()
+	conn, err := grpc.Dial(f.Endpoint, grpc.WithInsecure(),
```
nickolaev: The `f.conn` cna be reused here too.

sdk/vppagent/flush.go
```diff
@@ -0,0 +1,134 @@
...
+	// ensure the env variables are processed
+	if configuration == nil {
+		configuration = &common.NSConfiguration{}
+	}
+	configuration.CompleteNSConfiguration()
+
+	self := &Flush{
+		Endpoint: endpoint,
+	}
+	_ = self.reset()
```
nickolaev: Why not check the error?

sdk/vppagent/flush.go
```diff
@@ -0,0 +1,134 @@
...
+	})
+	if err != nil {
+		logrus.Errorf("Failed to reset VPP Agent: %s", err)
+	}
+	logrus.Infof("Finished resetting VPP Agent")
+	return nil
+}
+
+// Close implements the close handler
+func (f *Flush) Close(ctx context.Context, connection *connection.Connection) (*empty.Empty, error) {
```
nickolaev: Hmm are we sure there is no cleanup to be done on Close?

sdk/vppagent/flush.go
```diff
@@ -0,0 +1,134 @@
...
+	if f.GetNext() != nil {
+		return f.GetNext().Close(ctx, connection)
+	}
+	return &empty.Empty{}, nil
+}
+
+// NewFlush creates a Flush
+func NewFlush(configuration *common.NSConfiguration, endpoint string) *Flush {
```
nickolaev: Naming is hard, can we have a better name here? Maybe Update?

sdk/vppagent/memif_connect.go
```diff
@@ -0,0 +1,130 @@
...
+
+	if err := os.MkdirAll(socketDir, os.ModePerm); err != nil {
+		return nil, err
+	}
+
+	var name string
+	if mc.ConnectionSide == DESTINATION {
+		name = "DST-" + incomingConnection.GetId()
+	} else {
+		name = "SRC-" + incomingConnection.GetId()
```
nickolaev: Src and Dst are just legacy from the old code. Let's replace these.

sdk/vppagent/memif_connect.go
```diff
@@ -0,0 +1,130 @@
...
+	incomingConnection := incoming.(*connection.Connection)
+	if connectionData, ok := mc.Connections[incomingConnection.GetId()]; ok {
+		return connectionData
+	}
+	logrus.Errorf("GetOpaque outgoing not found for %v", incomingConnection)
+	return nil
+}
+
+// NewMemifConnect creates a MemifConnect
+func NewMemifConnect(configuration *common.NSConfiguration, side ConnectionSide) *MemifConnect {
```
nickolaev: So we have ClientCOnnect and MemifConnect. Do we need both?

test/applications/cmd/vppagent-firewall-nse/main.go
```diff
@@ -21,14 +21,15 @@ import (
...
+	"github.com/networkservicemesh/networkservicemesh/sdk/vppagent"
 	"github.com/sirupsen/logrus"
 )
 
 var version string
 
 func main() {
 	logrus.Info("Starting vppagent-firewall-nse...")
-	logrus.Infof("Versiion: %v", version)
+	logrus.Infof("Version: %v", version)
```
denis-tingajkin: nice 👍

sdk/vppagent/memif_connect.go
onobrod: Replace with in/out or eliminate at all?

test/applications/cmd/vppagent-firewall-nse/main.go
onobrod: :)

sdk/vppagent/client_connect.go
onobrod: Renamed to ClientMemifConnect (also thinking about MemifClientConnect name). I think it will be better to implement KernelConnect/ClientKernelConnect in a separate PR. @edwarnicke do we need to add them now? Should I create a task?

sdk/vppagent/client_connect.go
onobrod: Removed the prefix

sdk/vppagent/flush.go
onobrod: I'm also not an expert, but I like your suggestion and have implemented it — the code looks more simpler and clean now, thanks!

sdk/vppagent/flush.go
onobrod: For this func only? Not for the type, right? I thought there is some kind of name convention for these functions. Should I rename other New* functions in vppagent package too?

sdk/vppagent/memif_connect.go
onobrod: Removed

sdk/vppagent/memif_connect.go
onobrod: MemifConnect works with incoming connection while ClientMemifConnect works with outgoing connection and gets it from opaque. So they expect to receive different opaque data from its next composite — MemifConnect expects to receive ConnectionData or nothing, ClientMemifConnect expects to receive outgoing connection.

sdk/vppagent/flush.go
onobrod: Added removing DataChange from VPP Agent and closing connection to VPP Agent, please check

sdk/vppagent/flush.go
onobrod: Added error logging and nil return, please check

sdk/vppagent/cross_connect.go
onobrod: In ConnectionContext type we have SrcIpAddr and DstIpAddr fields. D[stIpAddr is used in MemifConnect](https://github.com/networkservicemesh/networkservicemesh/pull/1303/files#diff-c69e69e1dbbcae3efe6b8d16e38e8bd2R59) to decide whether ip address should be assigned.

Won't it lead to ambiguity if we will work with both src/dst and in/out?

sdk/vppagent/cross_connect.go
```diff
@@ -0,0 +1,119 @@
...
+
+func (xc *XConnect) appendDataChange(rv *configurator.Config, srcName, dstName string) *configurator.Config {
+	if rv == nil {
+		rv = &configurator.Config{}
+	}
+	if rv.VppConfig == nil {
+		rv.VppConfig = &vpp.ConfigData{}
+	}
+
+	rv.VppConfig.XconnectPairs = append(rv.VppConfig.XconnectPairs,
```
onobrod: TODO for all composites: always return a copy of DataChange object, do not mutate opaque data of the next composite

sdk/vppagent/cross_connect.go
nickolaev: OK so in cross connects we do not have IPs, that's the point of the cross connect - transparrent passhtrough. No one in the shain will know that the Corssconnect even exists. Both interfaces in the cross connect do not have IPs.

But that is a separate story. All these small composites should be completely independent. Or at least very loosely coupled. So whatever you name things here shoudl not be directly related to what other compsoites name things.

Src and Dst are not accurate here, since this is a directionless corssconnec. Even SideA and SideB is more direction independent and just says - htere are two sides here and I am connecting them, not really caring where they come from.
Hope that makes sense.

sdk/vppagent/cross_connect.go
nickolaev: Hmm interesting. Return it where? Are you talkoing about a different method?

sdk/vppagent/flush.go
nickolaev: Oh my fault, should have been more clear. `New*` is the standart, that is OK. WHat I meant was `flush` :) So like `NewUpdate` although that sounds weird too. But in any case we probably can figure out a better name than `flush`. Unless of course that is a vppagent terminology?

sdk/vppagent/memif_connect.go
nickolaev: Makes sense. This is a good example where a theorugh documenation will be really useful. Please consider a README.

sdk/vppagent/cross_connect.go
onobrod: Yes, not return, I'm about storing in opaque. I mean we send DataChange object through the chain and most likely we want to have snapshots of this object in each composite. Currently I mutate one DataChange instance in all composites. I think it will be good to copy DataChange object in each composite and mutate only this copy, do not affect other composites opaque.

sdk/vppagent/flush.go
onobrod: Not sure it is VPP Agent terminology, but seems `flush` is usually used to write some buffered data, it is exactly what our composite does.

@edwarnicke maybe you can explain more on this name choice?

sdk/vppagent/acl.go
```diff
@@ -0,0 +1,319 @@
...
+	endpoint.BaseCompositeEndpoint
+	Rules       map[string]string
+	Connections map[string]*ConnectionData
+}
+
+// Request implements the request handler
+func (a *ACL) Request(ctx context.Context, request *networkservice.NetworkServiceRequest) (*connection.Connection, error) {
+
+	if a.GetNext() == nil {
+		logrus.Fatal("The VPP Agent ACL composite requires that there is Next set")
```
haiodo: Should we return an error here?

sdk/vppagent/acl.go
```diff
@@ -0,0 +1,319 @@
...
+		match.Action = action
+		if err != nil {
+			logrus.Errorf("Parsing rule %s failed with %v", rule, err)
+			return nil, err
+		}
+
+		rules = append(rules, match)
+
+		rv.VppConfig.Acls = append(rv.VppConfig.Acls, &acl.ACL{
+			Name:  "IngressACL",
```
haiodo: Should names to be taken from somewhere? from rule's itself? It should not be same I suppose.

sdk/vppagent/acl.go
```diff
@@ -0,0 +1,319 @@
...
+	}
+	action, ok := acl.ACL_Rule_Action_value[strings.ToUpper(actionName)]
+	if !ok {
+		return acl.ACL_Rule_Action(0), fmt.Errorf("rule should have a valid 'action'")
+	}
+	return acl.ACL_Rule_Action(action), nil
+}
+
+func getIP(parsed map[string]string) (*acl.ACL_Rule_IpRule_Ip, error) {
+	dstNet, dstNetOk := parsed["dstnet"]
```
haiodo: Should we put "dstnet", "stcnet" into constants?

sdk/vppagent/acl.go
```diff
@@ -0,0 +1,319 @@
...
+		return &acl.ACL_Rule_IpRule_Ip{
+			DestinationNetwork: dstNet,
+			SourceNetwork:      srcNet,
+		}, nil
+	}
+	return nil, nil
+}
+
+func getICMP(parsed map[string]string) (*acl.ACL_Rule_IpRule_Icmp, error) {
+	icmpType, ok := parsed["icmptype"]
```
haiodo: same for icmptype

sdk/vppagent/client_memif_connect.go
```diff
@@ -0,0 +1,116 @@
...
+	endpoint.BaseCompositeEndpoint
+	Workspace   string
+	Connections map[string]*ConnectionData
+}
+
+// Request implements the request handler
+func (cmc *ClientMemifConnect) Request(ctx context.Context, request *networkservice.NetworkServiceRequest) (*connection.Connection, error) {
+
+	if cmc.GetNext() == nil {
+		logrus.Fatal("The VPP Agent Client Memif Connect composite requires that there is Next set")
```
haiodo: Should we return error here?

sdk/vppagent/cross_connect.go
```diff
@@ -0,0 +1,119 @@
...
+	endpoint.BaseCompositeEndpoint
+	Workspace   string
+	Connections map[string]*ConnectionData
+}
+
+// Request implements the request handler
+func (xc *XConnect) Request(ctx context.Context, request *networkservice.NetworkServiceRequest) (*connection.Connection, error) {
+
+	if xc.GetNext() == nil {
+		logrus.Fatal("The VPP Agent Cross Connect composite requires that there is Next set")
```
haiodo: Should we return error here?

sdk/vppagent/flush.go
haiodo: Yeah, name is not good. What about NewVppUpdateRequest?

sdk/vppagent/flush.go
```diff
@@ -0,0 +1,172 @@
...
+	if err != nil {
+		logrus.Errorf("Failed to reset VPP Agent: %v", err)
+		return nil
+	}
+
+	return self
+}
+
+func createConnection(endpoint string) (*grpc.ClientConn, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), 120*time.Second)
```
haiodo: Should we have timeout in constants or configurable somehow in pipeline?

sdk/vppagent/memif_connect.go
```diff
@@ -0,0 +1,135 @@
...
+	Workspace      string
+	ConnectionSide ConnectionSide
+	Connections    map[string]*ConnectionData
+}
+
+// Request implements the request handler
+func (mc *MemifConnect) Request(ctx context.Context, request *networkservice.NetworkServiceRequest) (*connection.Connection, error) {
+
+	if mc.GetNext() == nil {
+		logrus.Fatal("The VPP Agent Memif Connect composite requires that there is Next set")
```
haiodo: Should we return error here?

sdk/vppagent/acl.go
onobrod: @nickolaev could you please comment on this?

onobrod: Maybe we can append a name of an interface to make it kinda unique

sdk/vppagent/cross_connect.go
onobrod: Thank you a lot for the clarification, much clearer for me now. Will replace with In/Out because ACL needs to be applied only for incoming connection, so I cannot use smth like SideA/SideB or a slice of names.

sdk/vppagent/acl.go
onobrod: Done

sdk/vppagent/acl.go
onobrod: Done

sdk/vppagent/cross_connect.go
onobrod: Done

sdk/vppagent/cross_connect.go
onobrod: Done

sdk/vppagent/acl.go
onobrod: Fixed

sdk/vppagent/client_memif_connect.go
onobrod: Fixed

sdk/vppagent/flush.go
onobrod: Added constants like it is done in other code in the repository

sdk/vppagent/memif_connect.go
onobrod: Fixed

sdk/vppagent/cross_connect.go
onobrod: Fixed

sdk/vppagent/flush.go
onobrod: Reverted — if we store the connection in the field, we don't know when to close it (correct me if I wrong)

sdk/vppagent/acl.go
onobrod: Do not find in the doc anything special about ACL name, seems it should be unique only. Added interface name into it.

onobrod:
Decided not to create a separate README file, appended the existing one instead. Please, comment if the amount of provided documentation is not enough or it should be corrected in some places.

sdk/vppagent/memif_connect.go
onobrod: Done

sdk/vppagent/flush.go
edwarnicke: I'm fine with a better name than 'Flush'... I've been thinking a lot lately about storage, so "flush to disk" is bouncing around in my head.  NewVppUpdate or NewVPPUpdateRequest or NewSendVppUpdate or NewProgramVPP etc would all likely be better than Flush :)
@nickolaev @haiodo @fkautz thoughts on naming?

I do have a question though... would it be OK to fix the naming in a follow up PR?  I'm liking what I'm seeing so far in the PR, and so would love to (if the rest looks good) merge it... :)

sdk/vppagent/client_connect.go
edwarnicke: @onobrod Create an issue :)

nickolaev:
@onobrod We do lack licenses in all the files here. I specifically would like to retain the original license from the ACL code.

nickolaev:
Also we might use some docs on configuring the ACL component.

sdk/vppagent/client_connect.go
onobrod: Done

Issue #1282, opened 2019-06-25
Discussion: Why does SDK traverse down the chain then back up again?


edwarnicke:
I've noticed that the SDK traverses down the chain then back up again.  Meaning that if I have Chained Endpoint:
```go
	composite := endpoint.NewCompositeEndpoint(
		endpoint.NewMonitorEndpoint(configuration),
		newvppAgentACLComposite(configuration, config.getACLRulesConfig()),
		newVppAgentXConnComposite(configuration),
		endpoint.NewClientEndpoint(configuration),
		endpoint.NewConnectionEndpoint(configuration))
```

Each element in the chain first calls its 'Next', and then upon return does its work.  This is a bit confusing, because it means the work of ConnectionEndpoint (last in the list) is done first, then ClientEndpoint etc.  Logically it feels to the user like we are traversing the list last to first.  I presume there was some good reason for doing this, I am curious what it was :)

PR #1324, opened 2019-07-07
Simplification of SDK

edwarnicke:
This PR simplifies the SDK, by using the context.Context to pass
data between Endpoints in the chain.

By doing this, and a few other minor changes, the Endpoints used
in CompositeEndpoint just need to be normal implementers of

networkservice.NetworkServiceServer

Among other things, this means that CompositeEndpoints can be nested
within each other.

Note: Updates to sdk/README.md will be forthcoming.

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


sdk/endpoint/client.go
```diff
@@ -27,50 +27,47 @@ import (
...
+		if err != nil {
+			return nil, err
+		}
+	}
 
-	// TODO: check this. Hack??
-	outgoingConnection.GetMechanism().GetParameters()[connection.Workspace] = ""
-
-	cce.ioConnMap[incomingConnection.GetId()] = outgoingConnection
+	cce.ioConnMap[request.GetConnection().GetId()] = outgoingConnection
```
onobrod: Should we save outgoingConnection into the map just after we created it to be able to close it even if next request will fail?

sdk/vppagent/commit.go
```diff
@@ -23,99 +23,81 @@ const (
...
-		return nil, err
-	}
-
-	incomingConnection, err := f.GetNext().Request(ctx, request)
-	if err != nil {
-		return nil, err
+// Provides/Consumes from ctx context.Context:
+//     VppAgentConfig
+//	   Next
+func (f *Commit) Request(ctx context.Context, request *networkservice.NetworkServiceRequest) (*connection.Connection, error) {
```
onobrod: Please rename the receiver names to `c` also

sdk/vppagent/commit.go
```diff
@@ -23,99 +23,81 @@ const (
...
+	if endpoint.Next(ctx) != nil {
+		return endpoint.Next(ctx).Close(ctx, connection)
 	}
 	return &empty.Empty{}, nil
 }
 
 // Name returns the composite name
-func (f *Flush) Name() string {
+func (f *Commit) Name() string {
 	return "flush"
```
onobrod: Please fix this name also

sdk/vppagent/commit.go
```diff
@@ -23,99 +23,81 @@ const (
...
 // Close implements the close handler
-func (f *Flush) Close(ctx context.Context, connection *connection.Connection) (*empty.Empty, error) {
-	connectionData, err := getConnectionData(f.GetNext(), connection, false)
-	if err != nil {
-		return &empty.Empty{}, err
-	}
+// Provides/Consumes from ctx context.Context:
+//     VppAgentConfig
+//	   Next
+func (f *Commit) Close(ctx context.Context, connection *connection.Connection) (*empty.Empty, error) {
```
onobrod: Seems all /sdk/vppagent composites (except `Commit`) have almost identical Request() and Close() methods. Can we fix this somehow? Maybe can we store flushed data in a map inside `Commit` and just get it from it on Close()?

onobrod:
I like that child endpoints of a composite endpoint are called in direct order now, it looks more clear and seems it is exactly what you asking about in #1282

sdk/vppagent/commit.go
edwarnicke: Good catch :)  Will do :)

sdk/vppagent/commit.go
edwarnicke: Interesting idea :)

edwarnicke:
@onobrod Yes, but what you wrote was congruent with the SDK as it was at the time.  You did well.  It just brought a bunch of other things to my attention which I'm hacking at :)

sdk/endpoint/composite.go
```diff
@@ -19,82 +19,71 @@ import (
...
 	GrpcServer *grpc.Server
 }
 
-// ChainedEndpoint is the basic endpoint composition interface
-type ChainedEndpoint interface {
-	networkservice.NetworkServiceServer
-	Name() string
-	Init(context *InitContext) error
-	GetNext() ChainedEndpoint
-	GetOpaque(interface{}) interface{}
```
onobrod: Need to update sdk documentation here: https://github.com/networkservicemesh/networkservicemesh/blob/master/sdk/README.md#writing-a-chainedendpoint

haiodo:
Work is moved into #1524

PR #1343, opened 2019-07-14
Fix the name of init container by setting the proper env

### LINKED ISSUE #1342: pods depending on networkservicemeshci/nsm-init:latest fail
While attempting to build and deploy the nsm bridge-domain example in the examples repo, came across this problem:

```
  Normal   Pulling    40s (x3 over 80s)  kubelet, nsm-worker2  Pulling image "networkservicemeshci/nsm-init:latest"
  Warning  Failed     39s (x3 over 79s)  kubelet, nsm-worker2  Failed to pull image "networkservicemeshci/nsm-init:latest": rpc error: code = Unknown desc = failed to resolve image "docker.io/networkservicemeshci/nsm-init:latest"
```

In short, nsm-init:latest does not exist in the networkservicemeshci repo. We should either point to networkservicemesh/nsm-init or public a "latest" tag.

nickolaev:
Also, set the ImagePullPolicy to PullIfNotPresent

## Description
Fixes #1342 

## Motivation and Context
After the introduction of the new repo networkservicemeshci and the new methodology where we do not have "latest" the admission webhook is having issues deploying the right images for the init container.

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/.mk/k8s.mk
+++ b/.mk/k8s.mk
@@ -80,6 +80,9 @@ endif
 ifeq ($(CONTAINER_REPO),)
 CONTAINER_REPO=networkservicemesh
 endif
+ifeq ($(CONTAINER_TAG),)
+CONTAINER_TAG=latest
+endif
 
 kubectl = kubectl -n ${NSM_NAMESPACE}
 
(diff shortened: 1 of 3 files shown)


fkautz:
#1342 still fails with this patch. The problem isn't in the tag name, rather it's the repo name being set to "networkservicemeshci" instead of "networkservicemesh."

fkautz:
Looks like the failure is occuring due to a different reason, will need to investigate.

PR #1359, opened 2019-07-18
Admission webhook: propagate jaeger host and port env vars to NSCs init containers

tiswanso:
## Description

Since the mutating webhook admission controller adds the nsm-init container for NSCs, this change optionally adds the env vars for Jaeger tracing to the injected NSC initContainer if they are defined in the admission-webhook.

## Motivation and Context

To be able to configure NSCs to send tracing info to Jaeger for troubleshooting and GRPC sequence information.

## How Has This Been Tested?

- [ x ] Tested locally
- [ ] Have not tested

Testing without the env vars set is covered by the existing CI/integration tests.

## Types of changes

- [ x ] New feature (non-breaking change which adds functionality)

## Checklist:

### DIFF
--- a/k8s/cmd/admission-webhook/const.go
+++ b/k8s/cmd/admission-webhook/const.go
@@ -12,6 +12,8 @@ const (
 	repoEnv                      = "REPO"
 	initContainerEnv             = "INITCONTAINER"
 	tagEnv                       = "TAG"
+	jaegerHostEnv                = "JAEGER_SERVICE_HOST"
+	jaegerPortEnv                = "JAEGER_SERVICE_PORT_JAEGER"
 	repoDefault                  = "networkservicemesh"
 	initContainerDefault         = "nsm-init"
 	tagDefault                   = "latest"
--- a/.mk/docker.mk
+++ b/.mk/docker.mk
@@ -46,7 +46,7 @@ docker-save: $(addsuffix -save,$(addprefix docker-,$(BUILD_CONTAINERS)))
 
 .PHONY: docker-%-save
 docker-%-save: docker-%-build
-	@echo "Saving $*"
+	@echo "Saving $* to scripts/vagrant/images/$*.tar"
 	@mkdir -p scripts/vagrant/images/
 	@docker save -o scripts/vagrant/images/$*.tar ${ORG}/$*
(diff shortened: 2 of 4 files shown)


k8s/cmd/admission-webhook/const.go
```diff
@@ -12,6 +12,8 @@ const (
 	repoEnv                      = "REPO"
 	initContainerEnv             = "INITCONTAINER"
 	tagEnv                       = "TAG"
+	jaegerHostEnv                = "JAEGER_SERVICE_HOST"
+	jaegerPortEnv                = "JAEGER_SERVICE_PORT_JAEGER"
```
fkautz: Is there a reason this is "JAEGER_SERVICE_PORT_JAEGER" instead of "JAEGER_SERVICE_PORT"?

tiswanso: That's the value being used in the InitJaeger() function in pkg/tools/tools.go.  So, it needs to match that.

https://github.com/networkservicemesh/networkservicemesh/blob/master/pkg/tools/tools.go#L180

edwarnicke: Do we want to name JAEGER here, or are we really setting the HOST/PORT for sending OpenTracing information?

edwarnicke: @tiswanso Got it.  Makes sense... principle of least change.  If we want to change that... that's a different effort than this :)

tiswanso: Good question... definitely the current tools impl is opinionated on what is the tracing collector--ie. 
- https://github.com/networkservicemesh/networkservicemesh/blob/master/k8s/cmd/nsmd-k8s/main.go#L29
- https://github.com/networkservicemesh/networkservicemesh/blob/master/side-cars/cmd/nsm-init/main.go#L32
- https://github.com/networkservicemesh/networkservicemesh/blob/master/sdk/endpoint/endpoint.go#L73

That seems to be a common pattern in general (not necessarily specific to NSM) for golang.  Maybe there's a more opentracing generic way to code it complete with config settings but I think that'd be something to investigate and follow-on if desired.

Issue #1360, opened 2019-07-18
Rename environment variable JAEGER_SERVICE_PORT_JAEGER to JAEGER_SERVICE_PORT


fkautz:
Should we rename this? The name feels unnecessarily redundant.

edwarnicke:
Or we could simply use the standard Jaeger env variables:

https://github.com/jaegertracing/jaeger-client-go#environment-variables

And [config.FromEnv()](https://github.com/jaegertracing/jaeger-client-go/blob/master/config/config_env.go#L53)

edwarnicke:
Would need to be fixed [here](https://github.com/networkservicemesh/networkservicemesh/blob/master/pkg/tools/tools.go#L178)

PR #1368, opened 2019-07-22
helm: Add option for enabling jaeger tracing in NSM components

tiswanso:
- include var for setting monitoring k8s service types
- fix missing namespace settings for templates in nsm-monitoring chart

Signed-off-by: Tim Swanson <tiswanso@cisco.com>

## Description

Add the ability to enable Jaegar tracing on components in the NSM helm charts.  Additionally, add a setting that maps to the monitoring services k8s service types--e.g. NodePort. 

| Chart | New setting | Possible Values | Description |
|---|---|---|--|
| deployments/helm/nsm |  global.JaegerTracing | `true`, `false` |  Enables the env vars for Jaeger tracing in each NSM component |
|  deployments/helm/nsm-monitoring | monSvcType | `ClusterIP`, `NodePort`, `LoadBalancer` | The service type for the monitoring UIs--currently Jaeger and Skydive |

Example usage of the settings:
```
helm template deployments/helm/nsm --namespace nsm-system --set global.JaegerTracing=true > ~/tmp/nsm_k8s_mfst
helm template deployments/helm/nsm-monitoring --namespace nsm-system --set monSvcType=NodePort > ~/tmp/nsm_k8s_nsm_monitoring.yaml
```

## Motivation and Context

When using NSM with monitoring, it's very useful to enable Jaeger and have it be available via a service type that allows external access to the webUI.

## How Has This Been Tested?

- [ ] Covered by existing integration testing
- [ ] Added integration testing to cover
- [x] Tested locally
- [ ] Have not tested

I'm not sure if anything in the test suites renders installation manifests from helm.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1379, opened 2019-07-24
Use standard Jaeger envs

onobrod:
Signed-off-by: Viktoria Dlugopolskaya <viktoria.dlugopolskaya@xored.com>

## Motivation and Context
https://github.com/networkservicemesh/networkservicemesh/issues/1360

## How Has This Been Tested?

- [ ] Tested locally
- [x] Have not tested

## Types of changes

## Checklist:


pkg/tools/tools.go
```diff
@@ -145,32 +145,23 @@ func ParseKVStringToMap(input, sep, kvsep string) map[string]string {
...
+	cfg, err := config.FromEnv()
+	if err != nil {
+		panic(fmt.Sprintf("ERROR: cannot create Jaeger configuration: %v\n", err))
 	}
 
+	cfg.Sampler.Type = "const"
+	cfg.Sampler.Param = 1
+	cfg.Reporter.LogSpans = true
+
 	hostname, err := os.Hostname()
```
tiswanso: maybe we should only set cfg.ServiceName in here if it hasn't already been set by the config.FromEnv var `JAEGER_SERVICE_NAME`

onobrod: Fixed. What about other config parameters we set? They also may be not empty since they can be set via envs.

k8s/cmd/admission-webhook/const.go
```diff
@@ -12,8 +12,8 @@ const (
 	repoEnv                      = "REPO"
 	initContainerEnv             = "INITCONTAINER"
 	tagEnv                       = "TAG"
-	jaegerHostEnv                = "JAEGER_SERVICE_HOST"
-	jaegerPortEnv                = "JAEGER_SERVICE_PORT_JAEGER"
+	jaegerHostEnv                = "JAEGER_AGENT_HOST"
+	jaegerPortEnv                = "JAEGER_AGENT_PORT"
```
edwarnicke: Do we still need these constants?  Are they being used anywhere?

edwarnicke:
Don't we also need to fix the places (like the web admissions controller, various Pod specs etc) that are passing these variables through?  Does tracing still work given this change?

pkg/tools/tools.go
tiswanso: Good point.  I agree that it's better to detect whether they're set the from env and only default to these values if unset.

k8s/cmd/admission-webhook/const.go
tiswanso: We need to make the constant env name match the env name used by pkg/tools/tools.go::InitJaeger() so the values put in the initContainer patch are consumed properly.

tiswanso:
> Don't we also need to fix the places (like the web admissions controller, various Pod specs etc) that are passing these variables through? Does tracing still work given this change?

@edwarnicke -- the diffs in this change should cover all that since the InitJaeger() function is slurping in the env var settings directly.  The admission webhook code for the propagating the settings to the NSC initContainer is updated properly with the change made to the env var consts.

I'm pulling the PR down to try it out.

tiswanso:
fyi, there's a race between this and my PR ( #1368) for the helm chart changes to set the Jaeger vars.  The var names would need to be updated either in #1368 if this merges first or in this one if #1368 merges first.

pkg/tools/tools.go
onobrod: Done

PR #1298, opened 2019-07-01
Interdomain NSM

artem-belov:
## Description
Client in one domain consume a Network Service provided by an Endpoint in another domain.

## Motivation and Context
#714

## How Has This Been Tested?

- [ ] Covered by existing integration testing
- [x] Added integration testing to cover
- [x] Tested locally
- [ ] Have not tested

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

Issue #714, opened 2019-01-31
Spec: Interdomain NSM


edwarnicke:
[Draft Spec](https://docs.google.com/document/d/1QjdJbYvzzsn6ARFq2iQcdBG9RN9m-Lx9kOq53sasnH8/edit#)

matrohon:
I'm not able  to comment the draft.
This spec covers the use case where the client knows about domains he can ask for network services, but doesn't cover the use case where client just ask for a net service without knowing about the domain topology, am I wrong?

edwarnicke:
@artem-belov Updated the spec for pNSRs etc.  Please have a look and let me know what you think.

@nickolaev @haiodo @fkautz Please also have a look :)

PR #1417, opened 2019-07-31
Allow inbound traffic for aws and gke clouds

artem-belov:

## Description
Allow inbound traffic for aws and gke clouds for interdomain connection purposes

## Motivation and Context
Required for Interdomain NSM connection (#1298) #714

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


nickolaev:
Any security implications? Do we want to restrict the inbound from specific IPs? I can imagine it will be hard to sync accross all clouds the proper set of IPs but still opening all ports to each and everyone is maybe a bit tooo much ;)

artem-belov:
@nickolaev I think it is the best way to avoid firewall issues when testing remote mechanisms. 
I suggest to close 22nd port and icmp traffic but to leave ip white list as is.
What do you think?

nickolaev:
@artem-belov but you probably need to pass only the VxLAN traffic (for now). Can you permit only "udp 4789" or whatever standard port there is?
I understand that these arre short-lived, and probably *safe*. My concern is to not blindly leave everything open just like this. But if we find this is the only sane way, I will accept it. Let me know.

artem-belov:
@nickolaev Yep we can open ports for inderdomain nsmgr requests and for vxlan mechanism.  Then just open ports for each new remote mechanism.
@haiodo @edwarnicke What do you think?

haiodo:
@artem-belov let's open only required ports at this moment.

PR #1425, opened 2019-08-01
Update Helm and add it to CI

nickolaev:
## Description
Update the helm charts to the latest develomeents and add the ORG parameter.
Add helm.mk with make helm-install... and make helm-delete... targets.
Use these new targets in the CI with the examples.

## Motivation and Context
We need CI for the Helm charts and they needed update.

## How Has This Been Tested?

- [ ] Tested locally
- [x] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/.cloudtest/aws.yaml
+++ b/.cloudtest/aws.yaml
@@ -20,7 +20,12 @@ providers:
       - NSM_AWS_ACCESS_KEY_ID
       - NSM_AWS_SECRET_ACCESS_KEY
     scripts:
-      start: make aws-start
+      start: |
+        make aws-start
+        helm init
+        kubectl create serviceaccount --namespace kube-system tiller
+        kubectl create clusterrolebinding tiller-cluster-rule --clusterrole=cluster-admin --serviceaccount=kube-system:tiller
+        kubectl patch deploy --namespace kube-system tiller-deploy -p '{"spec":{"template":{"spec":{"serviceAccount":"tiller"}}}}'
       stop: |
         make aws-destroy
       prepare: make k8s-config
--- a/.cloudtest/aws.yaml
+++ b/.cloudtest/aws.yaml
@@ -30,4 +30,4 @@ providers:
         helm init
         kubectl create serviceaccount --namespace kube-system tiller
         kubectl create clusterrolebinding tiller-cluster-rule --clusterrole=cluster-admin --serviceaccount=kube-system:tiller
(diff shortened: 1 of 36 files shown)


.cloudtest/aws.yaml
```diff
@@ -10,7 +10,7 @@ providers:
     env:
       - CLUSTER_RULES_PREFIX=aws
       - AWS_CLUSTER_NAME=$(cluster-name)-$(date)-${CIRCLE_BUILD_NUM}-$(rands10)
-      - CONTAINER_REPO=networkservicemesh
+      - CONTAINER_REPO=networkservicemeshci
```
haiodo: I think better to move CONTAINER_REPO into env-check, to just ensure it is passed.

.cloudtest/azure.yaml
```diff
@@ -10,7 +10,7 @@ providers:
     env:
       - CLUSTER_RULES_PREFIX=azure
       - AZURE_CLUSTER_NAME=$(cluster-name)-$(date)-${CIRCLE_BUILD_NUM}-$(rands10)
-      - CONTAINER_REPO=networkservicemesh
+      - CONTAINER_REPO=networkservicemeshci
```
haiodo: Same here.

.cloudtest/gke.yaml
```diff
@@ -11,7 +11,7 @@ providers:
       - GKE_PROJECT_ID=ci-management
       - CLUSTER_RULES_PREFIX=gke
       - GKE_CLUSTER_NAME=$(cluster-name)-$(date)-${CIRCLE_BUILD_NUM}-$(rands10)
-      - CONTAINER_REPO=networkservicemesh
+      - CONTAINER_REPO=networkservicemeshci
```
haiodo: And here.

.cloudtest/vagrant.yaml
```diff
@@ -10,7 +10,7 @@ providers:
     env:
       - CLUSTER_RULES_PREFIX=vagrant
       - CONTAINER_BUILD_PREFIX=docker
-      - CONTAINER_REPO=networkservicemesh
+      - CONTAINER_REPO=networkservicemeshci
```
haiodo: and here.

.cloudtest/aws.yaml
nickolaev: OK will try and see how this works.

tiswanso:
The default tag of `master` while the build system defaults to `latest` screwed me up pretty good.

Issue #1397, opened 2019-07-26
Example icmp/vpn are failing intermittently


edwarnicke:
## Expected Behavior

CI tests to run reliably, and not fail intermittently.

Example icmp/vpn fail intermittently:
https://circleci.com/gh/networkservicemesh/networkservicemesh/82822

denis-tingajkin:
https://circleci.com/gh/networkservicemesh/networkservicemesh/83978?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link

denis-tingajkin:
https://circleci.com/gh/networkservicemesh/networkservicemesh/84180?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link

lobkovilya:
https://circleci.com/gh/networkservicemesh/networkservicemesh/87132

denis-tingajkin:
https://circleci.com/gh/networkservicemesh/networkservicemesh/91897

denis-tingajkin:
examples were moved to helm. Example icmp is stable for now. For example vpn we have another issue https://github.com/networkservicemesh/networkservicemesh/issues/1623

haiodo:
Closing since it is duplicate to #1623

PR #1430, opened 2019-08-02
Fix vpn example. Expand logs

denis-tingajkin:

## Motivation and Context
https://github.com/networkservicemesh/networkservicemesh/issues/1397

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/k8s/conf/vpn-gateway-nse.yaml
+++ b/k8s/conf/vpn-gateway-nse.yaml
@@ -36,7 +36,7 @@ spec:
             - name: TRACER_ENABLED
               value: "true"
             - name: IP_ADDRESS
-              value: "10.60.1.0/24"
+              value: "172.16.1.0/24"
           resources:
             limits:
               networkservicemesh.io/socket: 1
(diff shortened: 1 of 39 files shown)


k8s/conf/vpn-gateway-nse.yaml
```diff
@@ -36,7 +36,7 @@ spec:
             - name: TRACER_ENABLED
               value: "true"
             - name: IP_ADDRESS
-              value: "10.60.1.0/24"
+              value: "172.16.1.0/24"
```
nickolaev: Why this change?

denis-tingajkin: To exclude case with IP intersections on clouds. Because by logs which I saw looks like it is a root cause. Also, I've expanded logs for bash tests to get more information about the problem if it is still reproducible.

denis-tingajkin: Now bash tests logs save on CI in artifacts directory.

PR #1469, opened 2019-08-13
Go mod vendor

lobkovilya:
## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


edwarnicke:
Why are we looking at returning to vendoring? :)

lobkovilya:
First of all it is not "returning" :) Because now it is kind of hybrid mode of `go mod` and `vendor`. Flow is almost the same with `go mod` - just use `go get ...`, dependency stored to `go.mod`, but right before commit you have to exec `go mod vendor`. 

Anyway I just experiment with that and obviously that decision requires discussion. But there are couple good points:
- we are not depend on some servers that hosts dependencies (there were situation when servers weren't available)
- build time on CI reduces from ~4,5 min to ~2 min

nickolaev:
I have doubts here too. Let's see how this goes and we can decide.

edwarnicke:
@lobkovilya Are we not hitting the docker layer cache for building in CI?

Issue #1339, opened 2019-07-11
Spec: NSM Plugin Architecture


onobrod:
## Overview

The spec describes how plugin concept can be applied to Network Service Mesh and why it can be useful

## Link to the Google Docs document

[Spec](https://docs.google.com/document/d/1ETSCcSqklOZAlMMfiWp3x1O4J1o8OKnSprKZu68BTWg/edit?usp=sharing)

edwarnicke:
This seems to me to be a good idea.  Would it make sense to use a registration mechanism that looks similar in the dataplane as well?

onobrod:
@edwarnicke they are already similar: both dataplane and plugin send their socket locations on registration. And in both scenarios NSM is a registrar. Dataplane registration is more complicated: it has also a liveness stream and unregister service (maybe we will need unregister service too).

PR #1473, opened 2019-08-14
Plugins #2.1: Add name to registration request, unify plugin manager API and plugin API

onobrod:
## Motivation and Context
https://github.com/networkservicemesh/networkservicemesh/issues/1339

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


controlplane/pkg/apis/plugins/connection_plugin.proto
```diff
@@ -2,7 +2,15 @@ syntax = "proto3";
 
 package plugins;
 
-import "github.com/networkservicemesh/networkservicemesh/controlplane/pkg/apis/connectioncontext/connectioncontext.proto";
```
haiodo: May be we name it just Connection?

onobrod: I also don't like ConnectionInfo naming, but it we name it Connection, we will get methods like `NewConnection(connection)`, `connection.getConnection()` and `connection.SetConnection(connection)`. It confuses me more. What about ConnectionWrapper?

onobrod: Renamed to ConnectionWrapper

haiodo: Protobuf generate GetLocalConnection(), GetRemoteConnection(), so I think it is ok to have it named as just a connection.

haiodo: We could reused it in NSM since we have interface to handle both of them.

onobrod: Created https://github.com/networkservicemesh/networkservicemesh/issues/1497

Issue #1474, opened 2019-08-14
Implement discovery and registry plugins


onobrod:
## Overview

In this issue we plan to migrate our registry server to discovery and registry plugins
Main plugin issue: https://github.com/networkservicemesh/networkservicemesh/issues/1339

PR #1501, opened 2019-08-19
Plugins #3: Add discovery and registry plugin models

onobrod:
## Description
1) added discovery plugin model and discovery plugin manager
2) added registry plugin model and registry plugin manager
3) added a link to plugin registry into nsmd components
4) added validation to plugin registry (throw an error if not all required plugins are defined)
5) updated the doc

## Motivation and Context
https://github.com/networkservicemesh/networkservicemesh/issues/1474

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


onobrod:
@nickolaev @haiodo please review

controlplane/pkg/apis/plugins/discovery_plugin.proto
```diff
@@ -0,0 +1,21 @@
...
+    bool found = 1;
+    string payload = 2;
+    registry.NetworkService network_service = 3;
+    map<string, registry.NetworkServiceManager> network_service_managers = 4;
+    repeated registry.NetworkServiceEndpoint network_service_endpoints = 5;
+}
+
+service DiscoveryPlugin {
+    rpc FindNetworkService (FindNetworkServiceRequest) returns (FindNetworkServiceResponse);
+}
```
edwarnicke: Why do we need a new API beyond the existing Discovery API?

controlplane/pkg/apis/plugins/registry_plugin.proto
```diff
@@ -0,0 +1,17 @@
...
+message NSEList {
+    repeated registry.NetworkServiceEndpoint network_service_endpoints = 1;
+}
+
+service RegistryPlugin {
+    rpc RegisterNSM (registry.NetworkServiceManager) returns (registry.NetworkServiceManager);
+    rpc RegisterNSE (registry.NSERegistration) returns (registry.NSERegistration);
+    rpc RemoveNSE (registry.RemoveNSERequest) returns (google.protobuf.Empty);
+    rpc GetNSEs (google.protobuf.Empty) returns (NSEList);
+}
```
edwarnicke: Why are we creating a duplicate Registry api in addition to the existing one?

edwarnicke:
I am having trouble understanding why we are treating Registry and Discovery as plugins here, and why we are creating a new API that mirrors the existing Registry and Discovery APIs.  Could you help me understand?

haiodo:
@edwarnicke I think intentionally it is same registry API, and it should be in one place.

Plugin in compare to registry is a question how we connect to cluster specific code.

image

controlplane/pkg/apis/plugins/discovery_plugin.proto
onobrod: We want to replace the existing Discovery API with the new one — Discovery Plugin API

controlplane/pkg/apis/plugins/registry_plugin.proto
onobrod: Same for Registry API. In this PR I only add the new model. On the next PR I will remove the old model. So we will have only one.

controlplane/cmd/nsmd/main.go
```diff
@@ -59,7 +59,7 @@ func main() {
 	var server nsmd.NSMServer
 	var err error
 	// Start NSMD server first, load local NSE/client registry and only then start dataplane/wait for it and recover active connections.
-	if server, err = nsmd.StartNSMServer(model, manager, serviceRegistry, apiRegistry); err != nil {
+	if server, err = nsmd.StartNSMServer(model, manager, serviceRegistry, pluginRegistry, apiRegistry); err != nil {
```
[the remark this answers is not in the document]
onobrod: Not sure I clearly understand how we should handle it in StartNSMServer()... These three registries are different data structures and we do different things with them. One we want to store in a nsm server instance (like service registry). Another we want to just call some methods on (like api registry)

onobrod:
@edwarnicke we are creating Plugins API that will help us to change the direction of the nsmd <-> nsmd-k8s dependency. It should replace the existing API

Currently nsmd depends on nsmd-k8s and rely that nsmd-k8s is running on 5000 port. We want to make nsmd-k8s depending on nsmd, so it goes to nsmd and says on what unix socket it is running on (plugin registration process). After registration, nsmd will be able to call nsmd-k8s when it will needed it

controlplane/pkg/apis/plugins/registry_plugin.proto
onobrod: It is planned to leave only NetworkServiceRegistry service since it is used by NSEs. NetworkServiceDiscovery and NsmRegistry services will be replaced with the plugins

onobrod:
Closing, since I will no longer maintain this PR

PR #1522, opened 2019-08-22
Clear old mechanism configuration before update

artem-belov:

## Description
Clear old mechanism configuration before updating connection
Taps should not be touched

## Motivation and Context
SRv6 mechanism will add new segments list to policy instead if update:
```
[0].-    BSID: fd25::1:2
    Behavior: Encapsulation
    Type: Default
    FIB table: 0
    Segment Lists:
      [0].- < fe80::a00:27ff:fec6:ca82, fd25::1:3 > weight: 0S
      [1].- < fe80::a00:27ff:fec6:ca82, fd25::1:5 > weight: 0
```
Also old mechanism configuration may not be cleared if changing remote mechanism

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/dataplane/vppagent/pkg/converter/cross_connect_converter.go
+++ b/dataplane/vppagent/pkg/converter/cross_connect_converter.go
@@ -105,17 +105,18 @@ func (c *CrossConnectConverter) MechanismsToDataRequest(rv *configurator.Config,
 
 	srcName := srcPrefix + c.GetId()
 
+	var err error
 	if c.GetRemoteSource() != nil {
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -50,7 +50,7 @@ jobs:
 
 # code analysis
   test-static-code-analysis:
-    resource_class: medium+
+    resource_class: xlarge
```
denis-tingajkin: Is resource class `large`  solves the problem?

artem-belov: Ok

PR #1528, opened 2019-08-22
Add request and connection plugin to proxy-nsmd-k8s

onobrod:
## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


controlplane/pkg/apis/remote/connection/constants.go
```diff
@@ -2,7 +2,10 @@ package connection
...
 // Remote connection constants
 const (
-	VXLANSrcIP = "src_ip"
-	VXLANDstIP = "dst_ip"
-	VXLANVNI   = "vni"
+	VXLANSrcIP    = "src_ip"
+	VXLANSrcExtIP = "src_ext_ip"
+	VXLANDstIP    = "dst_ip"
+	VXLANDstExtIP = "dst_ext_ip"
+	VXLANUseExtIP = "use_ext_ip"
```
edwarnicke: This is to specific to the particularities of the pNSMgr for interdomain in here.  There's a standing question of how to solve the VNI issue for interdomain, it should be handled separately from plugin questions, and discussed seperately.  I am nervous about the notion of bringing minutia of the full path (internal/external IP) this deeply into the API.  It may still be the right way, but it needs its own issue, a discussion, and its own PR.

edwarnicke:
Could you say a bit more about the benefits of splitting out low level details such as this int a plugin?

PR #1541, opened 2019-08-28
Fix not allow same application selected as destination.

haiodo:

## Description

Fix for firewall, to not choose firewall to prevent infinite loop and stop on timeout.

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/controlplane/pkg/selector/match_selector.go
+++ b/controlplane/pkg/selector/match_selector.go
@@ -63,7 +63,8 @@ func (m *matchSelector) matchEndpoint(nsLabels map[string]string, ns *registry.N
 		for _, destination := range match.GetRoutes() {
 			// Each NSE should be matched against that destination
 			for _, nse := range networkServiceEndpoints {
-				if isSubset(nse.GetLabels(), destination.GetDestinationSelector()) {
+				if isSubset(nse.GetLabels(), destination.GetDestinationSelector()) &&
+					!isSubset(nsLabels, destination.GetDestinationSelector()) {
 					nseCandidates = append(nseCandidates, nse)
 				}
 			}
--- a/controlplane/pkg/selector/match_selector_test.go
+++ b/controlplane/pkg/selector/match_selector_test.go
@@ -20,6 +20,8 @@ import (
 	"strconv"
 	"testing"
 
+	"github.com/onsi/gomega"
+
 	"github.com/networkservicemesh/networkservicemesh/controlplane/pkg/apis/local/connection"
 	"github.com/networkservicemesh/networkservicemesh/controlplane/pkg/apis/registry"
 )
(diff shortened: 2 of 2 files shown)


controlplane/pkg/selector/match_selector.go
```diff
@@ -63,7 +63,8 @@ func (m *matchSelector) matchEndpoint(nsLabels map[string]string, ns *registry.N
 		for _, destination := range match.GetRoutes() {
 			// Each NSE should be matched against that destination
 			for _, nse := range networkServiceEndpoints {
-				if isSubset(nse.GetLabels(), destination.GetDestinationSelector()) {
+				if isSubset(nse.GetLabels(), destination.GetDestinationSelector()) &&
+					!isSubset(nsLabels, destination.GetDestinationSelector()) {
```
nickolaev: I can't see why the requested Labels should be matched against the destination selector. This might work for the particular case but it looks wrong to me.

haiodo: Question they should not be matched,
Please check test example.

Another idea is to have a list of all already used endpoint ids smoother in context, and do not allow to choose same NSE few times.

controlplane/pkg/selector/match_selector_test.go
```diff
@@ -477,3 +479,101 @@ func Test_matchSelector_SelectEndpoint(t *testing.T) {
...
+	})
+	g.Expect(endpoint1).ToNot(gomega.BeNil())
+	endpoint2 := selector.SelectEndpoint(c, ns, []*registry.NetworkServiceEndpoint{
+		firewallEp,
+		gatewayEp,
+	})
+	g.Expect(endpoint2).To(gomega.BeNil())
+}
+
+func createMatches(nsName string) *registry.NetworkService {
```
nickolaev: `createFirewallMatchers` maybe?

controlplane/pkg/selector/match_selector.go
nickolaev: So we want to avoid any kind of "loops" in the graph we build. The problem is that there is no "centralized" graph building but instead each NSMgr does its own selection when the particular Client requests a service. So actually the central "context" is the Endpoint registry in Kubernetes. The only feasible way to avoid loops seems to be to mark the Endpoints in the registry with a "NS Instance ID", but again that will be very hard to coordinate between all the NSMgrs.

nickolaev: But if we want to fix the immediate problem, we can at least try to ensure that a Client never gets an Endpoint declared in the same pod.

haiodo: Hmps, At this point we already know we are not first endpoint in chain, so potentially we could remove all initial selections? So firewall will be filtered, Thoughts?

haiodo:
Handled by @lobkovilya already.

PR #1546, opened 2019-08-29
Fix CloudTest to work with timeouts and make Packet Destroy more smart.

haiodo:
## Description

1. Fix for use of global timeout in CloudTest
2. Fix Packet.Destroy to use timeout and wait for devices to be really destroyed.
3. Fix CloudTest to stop operation on cluster if both start/destroy caused errors.

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,8 +1,13 @@
 module github.com/networkservicemesh/networkservicemesh
 
 // Use ./scripts/switch_k8s_version.sh to change k8s version
+replace gonum.org/v1/gonum => github.com/gonum/gonum v0.0.0-20190331200053-3d26580ed485
+
+replace gonum.org/v1/netlib => github.com/gonum/netlib v0.0.0-20190331212654-76723241ea4e
 
 require (
+	github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751 // indirect
+	github.com/alecthomas/units v0.0.0-20190717042225-c3de453c63f4 // indirect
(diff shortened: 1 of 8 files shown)


go.mod
```diff
@@ -1,8 +1,13 @@
 module github.com/networkservicemesh/networkservicemesh
 
 // Use ./scripts/switch_k8s_version.sh to change k8s version
+replace gonum.org/v1/gonum => github.com/gonum/gonum v0.0.0-20190331200053-3d26580ed485
```
haiodo: @nickolaev this two guys are not accessible from Russia anymore, not sure why. Checked from few places.

test/cloudtest/pkg/commands/main.go
```diff
@@ -208,8 +208,15 @@ func PerformTesting(config *config.CloudTestConfig, factory k8s.ValidationFactor
 	// Fill tasks to be executed..
 	ctx.createTasks()
 
-	ctx.performExecution()
-	return ctx.generateJUnitReportFile()
+	err := ctx.performExecution()
+	reportfile, err2 := ctx.generateJUnitReportFile()
+	if err2 != nil && err != nil {
+		logrus.Errorf("Error during generation of report: %v", err2)
```
artem-belov: Check if err2 not nil should be enough here

haiodo: Thanks.

PR #1548, opened 2019-08-29
Allow local vendoring

haiodo:
Passing VENDORING="-mod vendor" will enable local vendoring

## Description

To speedup local development, local Vendoring cloud be enable by passing environment variable
```
export VENDORING="-mod vendor"
```

It will call `go mod vendor` to store all deps into vendor folder and will trigger all docker images to use vendored sources.

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/k8s/build/Dockerfile.nsmdp
+++ b/k8s/build/Dockerfile.nsmdp
@@ -1,5 +1,6 @@
 FROM golang:alpine as build
 ARG VERSION=unspecified
+ARG VENDORING
 RUN apk --no-cache add git
 ENV PACKAGEPATH=github.com/networkservicemesh/networkservicemesh/
 ENV GO111MODULE=on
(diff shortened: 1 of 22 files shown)


k8s/build/Dockerfile.nsmdp
```diff
@@ -8,10 +9,10 @@ RUN mkdir /root/networkservicemesh
 ADD ["go.mod","/root/networkservicemesh"]
 ADD ["./scripts/go-mod-download.sh","/root/networkservicemesh"]
 WORKDIR /root/networkservicemesh/
-RUN ./go-mod-download.sh
+RUN VENDORING=${VENDORING} ./go-mod-download.sh
 
 ADD [".","/root/networkservicemesh"]
-RUN CGO_ENABLED=0 GOOS=linux go build -ldflags "-extldflags "-static" -X  main.version=${VERSION}" -o /go/bin/nsmdp ./k8s/cmd/nsmdp
+RUN CGO_ENABLED=0 GOOS=linux go build ${VENDORING} -ldflags "-extldflags '-static' -X  main.version=${VERSION}" -o /go/bin/nsmdp ./k8s/cmd/nsmdp
```
denis-tingajkin: Hm.. I tried to use apostrophes but it is not worked for me.

scripts/go-mod-download.sh
```diff
@@ -1,5 +1,13 @@
 #!/usr/bin/env sh
 
+if [ -z "$VENDORING" ]
+then
+      echo "No vendoring"
+else
+      echo "Vendoring is enabled, not need to download stuff."
```
nickolaev: ` no need to download`

scripts/vagrant/Vagrantfile
```diff
@@ -17,7 +17,7 @@ Vagrant.configure("2") do |config|
   end
   config.vm.provider "virtualbox" do |v, override|
     override.vm.box = "bento/ubuntu-18.04"
-    v.memory = 4096
+    v.memory = 2048
```
nickolaev: This change should be part of a separate PR.

scripts/go-mod-download.sh
haiodo: fixed

haiodo:
> I have two small remarks, please address them. Other than that, looks OK.
> 
> On a side note - we do have lots of repetitions in the Dockerfiles, maybe we can have a common Dockerfile.build image with all the neededprerequisites and then just extend it for each binary we build.

@nickolaev I'm actually thinking about applying template approach for our docker build infrastructure. So planed to do it on some of next PRs.

It will allow to make templates smaller and give ability to have individual docker builds and use go caching to have just one container to build on build system, since using of go cache will improve a build speed a lot.

PR #1553, opened 2019-08-31
Support for metrics in Kernel-forwarder

rdimitrov:
## Description

Introduce initial support for metrics in Kernel-forwarder.
It keeps track of each interface being present, collect metrics for it and updates the metric collector.

For the time being it uses the format used by VPP, for ex. `{"SRC-<ID>":<metrics>}` or `{"DST-<ID>":<metrics>}`

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


ivanayov:
From the history [DataplaneMetricsEnabledDefault](https://github.com/networkservicemesh/networkservicemesh/blob/master/dataplane/pkg/common/dataplane.go#L51) was `true` by default, but set to `false` by mistake with the kernel-forwarder. Can you please set back to `true`?

Issue #1139, opened 2019-05-20
Packet-deploy fails sometimes


lobkovilya:
## Expected Behavior
Successful deploy of instances to packet

## Current Behavior
https://circleci.com/gh/networkservicemesh/networkservicemesh/52267

## Possible solution 
Change location if deploy has been failed

haiodo:
It seems we need to randomly check different zone.

lobkovilya:
After increasing log level to *DEBUG* received such output:
[build_60836_step_110_container_0.txt](https://github.com/networkservicemesh/networkservicemesh/files/3239638/build_60836_step_110_container_0.txt)

lobkovilya:
https://github.com/terraform-providers/terraform-provider-packet/issues/153

PR #1559, opened 2019-09-02
Scale back packet retries due to high API volume

edwarnicke:
Packet has complained that we are creating issues with high API volume:

https://support.packet.com/tickets/BQHG-3192-IDKS

This may be because of our 5 retries times 2 servers == 10 requests when there is an issue with packet.  While we sort out what to do about it, lets reduce the retries to 1 for the moment.

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/.cloudtest/packet.yaml
+++ b/.cloudtest/packet.yaml
@@ -3,7 +3,7 @@ providers:
   - name: "packet"
     kind: "packet"
     instances: 2
-    retry: 5
+    retry: 1
     node-count: 2
     enabled: true
     timeout: 600 # 10 minutes to start cluster


vielmetti:
Previously, https://github.com/networkservicemesh/networkservicemesh/issues/1139

haiodo:
It seems we maybe need to wait for shutdown, for devices here.

https://github.com/networkservicemesh/networkservicemesh/blob/744d4adaf8831a63256cc1c691c81bb68e9a5053/test/cloudtest/pkg/providers/packet/packet_provider.go#L369

@vielmetti all our logic related to Create/Delete of clusters are now here: https://github.com/networkservicemesh/networkservicemesh/blob/master/test/cloudtest/pkg/providers/packet/packet_provider.go#L369

So I suppose in addition I will add  add some warmups, between passing creation/deletion requests to packet today.

haiodo:
Also it seems I've found issue on our side:

We have timeout for 5 mins to create packet cluster, and if it fails we try to destroy all stuff.

But packet response with:
```
packngo.Rate{RequestLimit:0, RequestsRemaining:0, Reset:packngo.Timestamp{0001-01-01 00:00:00 +0000 UTC}}
DELETE https://api.packet.net/devices/fc1666e9-450d-4364-b0cd-e394c0350d7b: 422 Cannot delete a device while it is provisioning 
```
So I definitely need to wait for status and device dis-appearance during destroy. And add support for timeout properlty.

haiodo:
@edwarnicke I've added more complex fix to this issue at https://github.com/networkservicemesh/networkservicemesh/pull/1546

fkautz:
@haiodo @edwarnicke should this be closed?

vielmetti:
The associated Packet ticket has been closed, I'm OK to close this. (If it reappears, we will know where to look first.)

Issue #1551, opened 2019-08-31
k8s-deploy seems broken


rdimitrov:
## Expected Behavior

`make k8s-deploy` should result in a working deployment.
## Current Behavior

`make k8s-deploy` results in a broken deployment.

## Steps to Reproduce

1. Checkout master
2. `make vagrant-start`
3. `make k8s-save`
4. `make k8s-deploy`

## Context

* Kubernetes Version:
* etc.

## Failure Logs

```bash
Every 2.0s: kubectl get pods --all-namespaces -o wide                                                                                   dimitrovr-a01.vmware.com: Sat Aug 31 07:31:39 2019

NAMESPACE     NAME                                          READY   STATUS             RESTARTS   AGE     IP               NODE           NOMINATED NODE   READINESS GATES
kube-system   coredns-5c98db65d4-4lsmf                      1/1     Running            0          9m1s    10.32.0.3        kube-master    <none>           <none>
kube-system   coredns-5c98db65d4-lb2t4                      1/1     Running            0          9m1s    10.32.0.2        kube-master    <none>           <none>
kube-system   etcd-kube-master                              1/1     Running            0          9m15s   172.28.128.178   kube-master    <none>           <none>
kube-system   kube-apiserver-kube-master                    1/1     Running            0          9m15s   172.28.128.178   kube-master    <none>           <none>
kube-system   kube-controller-manager-kube-master           1/1     Running            0          9m15s   172.28.128.178   kube-master    <none>           <none>
kube-system   kube-proxy-8jlfl                              1/1     Running            0          6m34s   172.28.128.179   kube-worker1   <none>           <none>
kube-system   kube-proxy-bn6fv                              1/1     Running            0          9m1s    172.28.128.178   kube-master    <none>           <none>
kube-system   kube-scheduler-kube-master                    1/1     Running            0          9m15s   172.28.128.178   kube-master    <none>           <none>
kube-system   weave-net-6g8ln                               2/2     Running            0          6m34s   172.28.128.179   kube-worker1   <none>           <none>
kube-system   weave-net-tjp92                               2/2     Running            0          9m1s    172.28.128.178   kube-master    <none>           <none>
nsm-system    alpine-nsc-5979b9ccc9-hgvdt                   1/3     ErrImagePull       0          3m42s   10.44.0.7        kube-worker1   <none>           <none>
[...]

PR #1502, opened 2019-08-19
Eliminating "k8s/conf" dir #2: Move kernel-forwarder and proxy-nsmgr to Helm

lobkovilya:
## How Has This Been Tested?

- [ ] Covered by existing integration testing
- [ ] Added integration testing to cover
- [ ] Tested locally
- [ ] Have not tested

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1560, opened 2019-09-03
Delete deprecated API and fix docs

denis-tingajkin:

## Description
By this branches:
1) https://github.com/networkservicemesh/networkservicemesh/pull/1502
2) https://github.com/networkservicemesh/networkservicemesh/pull/1492

We eliminated "k8s/conf to use helm, but we did not delete deprecated API

## Motivation and Context
https://github.com/networkservicemesh/networkservicemesh/issues/1551
https://github.com/networkservicemesh/networkservicemesh/issues/1591

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

- [X] I have updated the documentation accordingly.


docs/guide-build.md
```diff
@@ -77,13 +79,13 @@ make k8s-check
 
 which will try pinging from NSCs to NSEs.
 
-You can remove the effects of k8s-deploy with:
+You can remove the effects of helm-%-install with:
```
lobkovilya: It is not correct command, we have format like `helm-install-%`

rdimitrov:
@denis-tingajkin I believe this doesn't solve the issue you're referring to.

`nsm-install-icmp-responder` will still fail as `k8s-save` won't built the dependent images and thus 'k8s-load-images` won't load them. 

Also we should add `make helm-init` to the docs otherwise it will fail.

denis-tingajkin:
@rdimitrov as I know we already have helm tests wihch working fine on CI.  Am I missed something? can you please describe the failture case with steps to reproduce?

rdimitrov:
Sure, it's actually just following the ones in the docs - 
1. Clone repository
2. `make vagrant-start`
3. `make k8s-save`
4. `make k8s-load-images`
5. `make helm-install-nsm`
6. `make helm-install-icmp-responder`

It's between 4 and 5 where I suggest to have `make helm-init`, assuming it's a fresh install.

denis-tingajkin:
@rdimitrov  thanks, I got it. Yes, you are right we need also edit DEPLOYS list

docs/guide-build.md
denis-tingajkin: thans, fixed

.mk/k8s.mk
```diff
@@ -36,7 +36,7 @@ endif
 DEPLOY_NSM = nsmgr $(DEPLOY_FORWARDING_PLANE)
 DEPLOY_PROXY_NSM = proxy-nsmgr
 DEPLOY_INFRA = $(DEPLOY_TRACING) $(DEPLOY_WEBHOOK) $(DEPLOY_NSM) $(DEPLOY_PROXY_NSM) $(DEPLOY_MONITOR)
-DEPLOYS = $(DEPLOY_INFRA) $(DEPLOY_ICMP) $(DEPLOY_VPN)
+DEPLOYS = $(DEPLOY_INFRA) $(DEPLOY_ICMP) $(DEPLOY_VPN) nsm-coredns nsm-monitor
```
rdimitrov: Do you think it'd be better if it is part of DEPLOY_NSM or a separate DEPLOY variable? I'm saying that because if someone deploys only the infra part, then these two will not be deployed/loaded.

denis-tingajkin: actually `DEPLOY_INFRA` used only to collect final list of DEPLOYS. But we can re-use these variables in the future. Fixed.

.mk/k8s.mk
```diff
@@ -16,7 +16,7 @@ K8S_CONF_DIR = k8s/conf
 
 # Deployments - common
 DEPLOY_TRACING = jaeger
-DEPLOY_WEBHOOK = admission-webhook
+DEPLOY_WEBHOOK = admission-webhook nsm-coredns nsm-monitor
```
edwarnicke: I don't quite understand why we are adding these here.  Could you explain?

edwarnicke: OK... realized I need to read more further down to understand... will try that :)

denis-tingajkin:
@edwarnicke I think we can save `k8s-%-deploy` staff and rewrite it to use helm inside. Please share your thoughts about this.

rdimitrov:
@denis-tingajkin I'm up for that 👍 

Personally, I find the `k8s-` targets really intuitive, so I was thinking if we migrate and/or unify all `helm-`prefixed targets to `k8s-` ones (save, load, deploy) Although they'd use helm underneath, they are still k8s related. Plus it will keep that good consitent view of the make targets we have so far. Do you think that makes sense?

denis-tingajkin:
@edwarnicke here one open question:

Should we save `k8s-deploy` staff with using helm?

edwarnicke:
@denis-tingajkin That would simplify the transition 😀

docs/guide-build.md
```diff
@@ -77,13 +79,13 @@ make k8s-check
...
-You can remove the effects of k8s-deploy with:
+You can remove the effects of helm-install-% with:
 
 ```bash
-make k8s-delete
+make helm-delete-%
 ```
 
-As in the case with `save` and `build`, you can always do this for a particular component, like ```make k8s-nsc-deploy``` or ```make k8s-nsc-delete```.
+As in the case with `save` and `build`, you can always do this for a particular component, like ```make helm-install-nsc``` or ```make helm-delete-nsc```.
```
nickolaev: There is no `helm-install-nsc`

nickolaev: Essentially we are loosing this target and this should be completely fine.

denis-tingajkin: thanks for review. Fixed

PR #1578, opened 2019-09-08
Fix segmentation violation when IP prefix is invalid

fkautz:


nickolaev:
@fkautz does this one still make sense?

PR #1585, opened 2019-09-09
WIP: Use cross compile to build containers

haiodo:
1. Use crosscompile flags to build all applications with local go cache.
2. Remove outdated build artifacts.

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


.mk/k8s.mk
```diff
@@ -160,7 +161,7 @@ k8s-proxy-nsmgr-save:  $(addsuffix -save,$(addprefix ${CONTAINER_BUILD_PREFIX}-,
 .PHONY: k8s-proxy-nsmgr-load-images
 k8s-proxy-nsmgr-load-images:  $(addsuffix -load-images,$(addprefix ${CLUSTER_RULES_PREFIX}-,$(PROXY_NSMGR_CONTAINERS)))
 
-VPPAGENT_DATAPLANE_CONTAINERS = vppagent-dataplane vppagent-dataplane-dev
+VPPAGENT_DATAPLANE_CONTAINERS = vppagent-dataplane
```
nickolaev: I would bet that the lack of the dev slims down the buidl time.

edwarnicke:
I get an error when I try to use this locally:

```
go build vendor/golang.org/x/crypto/internal/subtle: mkdir /usr/local/go/pkg/linux_amd64: permission denied
```

.circleci/config.yml
```diff
@@ -32,7 +34,7 @@ jobs:
       - run: sudo ./scripts/install-protoc.sh
       - run: ./scripts/for-each-module.sh "go generate ./..."
       - run: make vet check
-      - run: ./scripts/for-each-module.sh "go build ./..."
+      - run: sudo sudo chmod -R guo+w /usr/local/go/pkg
```
edwarnicke: This seems related to my error:
```
go build vendor/golang.org/x/crypto/internal/subtle: mkdir /usr/local/go/pkg/linux_amd64: permission denied
```

Is there a way to avoid needing to munge system level permissions?  Why are we having to do so?

test/kubetest/pods/vppagent_dataplane.go
```diff
@@ -58,7 +58,7 @@ func createVPPDataplanePod(name string, node *v1.Node, liveness, readiness *v1.P
 			Containers: []v1.Container{
 				containerMod(&v1.Container{
 					Name:            "vppagent-dataplane",
-					Image:           "networkservicemesh/vppagent-dataplane-dev:latest",
+					Image:           "networkservicemesh/vppagent-dataplane:latest",
```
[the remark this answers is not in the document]
haiodo: dev image was with debug symbols, but I've not seen any vpp crashes for a very long time, so removed building of it intentionally.

build/functions.sh
```diff
@@ -0,0 +1,86 @@
...
+export GOARCH=amd64
+
+export GREEN='\033[0;32m'
+export NC='\033[0m' # No Color
+
+mkdir -p "${BIN_DIR}"
+
+function appToSrc {
+    case "$1" in
+        admission-webhook)      echo "./cmd/admission-webhook" ;;
```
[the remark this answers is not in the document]
edwarnicke: @nickolaev I have the same base reaction you do here... its another place to maintain things.  That said, we have simply relocated this information from Dockerfiles to this script.  I'd like to see if we can think of a better way though.

build/build.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+APP=$1
+ORG=$2
+
+BIN_DIR="$(pwd)/build/dist/${APP}"
+
+# Include function to map apps
+# shellcheck disable=SC1091
```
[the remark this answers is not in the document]
haiodo: https://github.com/koalaman/shellcheck/wiki/SC1091 It shows error with source, I've not find a way to fix it properly.

.mk/k8s.mk
haiodo: not much, but a bit yes.

.circleci/config.yml
haiodo: Yes, put GOROOT not into system location fixes this. And I suppose this is only relevant for macos

build/functions.sh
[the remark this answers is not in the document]
haiodo: make file approach is also not good in my view, since it still require a lot of stuff to maintain.
This Pr is more like check if cross compile is working fine for us.

So probable we could go with generic make file approach for any of components, or we could choose some from this list: http://go-lang.cat-v.org/dev-utils

.circleci/config.yml
edwarnicke: I still don't understand why we have this issue to begin with... why are we munging with GOROOT?

edwarnicke: OK... figured it out.  If you don't use -i you don't have this problem.  In fact its [recommended not to use -i anymore at all](https://github.com/golang/go/issues/24674#issuecomment-379143645)

.circleci/config.yml
```diff
@@ -125,8 +135,9 @@ jobs:
             if [ "${CIRCLE_BRANCH}" == "master" ]; then
               export STORE_LOGS_IN_ANY_CASES=true
             fi
+            sudo sudo chmod -R guo+w /usr/local/go/pkg
             pushd test
-            CGO_ENABLED=0 GOOS=linux go build -o ../cloudtest ./cloudtest/cmd/cloudtest
+            CGO_ENABLED=0 GOOS=linux go build -i -o ../cloudtest ./cloudtest/cmd/cloudtest
```
edwarnicke: If you remove the -i here, you don't need to chmod /usr/local/go/pkg

build/functions.sh
```diff
@@ -0,0 +1,86 @@
...
+        exit 1
+    esac
+}
+
+function compileApp {
+    src=$(appToSrc "$1")
+    echo "Compile ${src}"
+    rootDir=$(srcWdir "$1")
+    pushd "${rootDir}" 1> /dev/null || exit 1
+    if ! go build -i -ldflags "-extldflags '-static' -X  main.version=${VERSION}" -o "${BIN_DIR}/$1" "${src}"
```
edwarnicke: If you remove the -i here, you don't need to chmod /usr/local/go/pkg

haiodo: Thanks.

Issue #1563, opened 2019-09-04
EthernetContext


edwarnicke:
## Overview

As noted in #1225, we can have situations where autohealing results in temporary ping failures because the ethernet address for the IP address changes, and thus existing arp table entries are wrong.

What this reveals is a lack of proper ```EthernetContext``` in ```ConnectionContext``` to convey optionally the dst/src ethernet addresses.

The basics are fairly simple.  Add to [ConnectionContext](https://github.com/networkservicemesh/networkservicemesh/blob/master/controlplane/pkg/apis/connectioncontext/connectioncontext.proto#L77) a new submessage 'EthernetContext':

```proto
message ConnectionContext {
    IPContext ip_context = 1; /* IP related context */
    DNSContext dns_context = 2; /* DNS related context */
    EthernetContext = 3; /* Ethernet related context */
}
```

where EthernetContext is defined as:

```proto
message EthernetContext {
   string src_mac_address = 1; /* Mac address of source side of connection */
   string dst_mac_address = 2; /* Mac address of destination side of connection */
}
```

# Implementation Changes

## Forwarders
The [vppagent-forwarder](https://github.com/networkservicemesh/networkservicemesh/tree/master/dataplane/vppagent) and [kernel-forwarder](https://github.com/networkservicemesh/networkservicemesh/tree/master/dataplane/kernel-forwarder) will need to be updated such that if and only if there is an EthernetContext with a properly set src_mac_address is set, and the mechanism is 'kernel interface' we set the mac address.  Likewise for NSEs with mechanism kernel interface and dst_mac_address set.

## NSE Composite Endpoint

Once #1524 has been merged (soon), write a Composite Endpoint analogous to the IPAMEndpoint that adds an EthernetContext with a [locally administered](http://www.noah.org/wiki/MAC_address#locally_administered_address) dst_mac_address for the Endpoint.  Modify the [vppagent.MemifConnect](https://github.com/networkservicemesh/networkservicemesh/pull/1524/files#diff-c69e69e1dbbcae3efe6b8d16e38e8bd2) to apply the dst_mac_address if provided.

Modify the [vppagent.ClientMemifConnect](https://github.com/networkservicemesh/networkservicemesh/pull/1524/files#diff-6e5547a57fb01ef859eb3f928545c54c) to apply the src_mac_address if provided.

## References
#1225

denis-tingajkin:
In progress

denis-tingajkin:
[...]

Issue #1225, opened 2019-06-11
NSC needs few attempts for ping to be passed after Dataplane Heal


artem-belov:
## Expected Behavior

Ping should be passed from first attempt after connection recovery done

## Current Behavior

First few ping attempts are failing after connection recovery done

## Failure Information (for bugs)

TestDataplaneHealMultiNodesRemote:
Nsc keeps old arp entries with wrong NSE hardware address and it makes few failed ping attempts before receiving correct arp entry.

## Steps to Reproduce

Change `fixture.CheckNsc ` to `kubetest.CheckNSC` in TestDataplaneHealMultiNodesRemote and run test

edwarnicke:
Interesting corner case.  How long does it take to fully restore connectivity?

denis-tingajkin:
Looks like simple typo in `Steps to reproduce`. Should be `Change kubetest.HealCheckerNsc to kubetest.CheckNSC...`

PR #1569, opened 2019-09-05
Refactor vppagent.Dataplane to in new SDK fashion #2

denis-tingajkin:
Signed-off-by: denis-tingajkin <denis.tingajkin@xored.com>

## Motivation and Context
https://github.com/networkservicemesh/networkservicemesh/issues/1544

## How Has This Been Tested?

- [X] Covered by existing integration testing
- [ ] Added integration testing to cover
- [ ] Tested locally
- [ ] Have not tested

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [X] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1606, opened 2019-09-12
WIP: Add EthernetContext to solve Issues#1225

denis-tingajkin:
## Motivation and Context
https://github.com/networkservicemesh/networkservicemesh/issues/1563
https://github.com/networkservicemesh/networkservicemesh/issues/1225

## How Has This Been Tested?

- [X] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


sdk/endpoint/ethernet.go
```diff
@@ -0,0 +1,75 @@
...
+	and := []byte{254, 255, 255, 255, 255, 255}
+	min := len(result)
+	if min > len(input) {
+		min = len(input)
+	}
+	for i := 0; i < min; i++ {
+		result[i] = or[i] | input[i]&and[i]
+	}
+	return result
+}
```
edwarnicke: This is very good if I want to specify the Ethernet Address.  But often, I don't.  We just want to capture the Ethernet Address generated for the interface by VPP and set it as the dstMac in the Connection returned from the Request.

denis-tingajkin: @edwarnicke This is clear for me but in the proposal  https://github.com/networkservicemesh/networkservicemesh/issues/1563 we should apply mac addresses from an endpoint only for memif connections. (Please, correct me if this is wrong)

What we should do if dataplane received a request with ethernet context with not empty mac addresses and the connection has kernel interface mechanism?

My options are:
1) Ignore these addresses and rewrite fields uses vpp agent generated mac address
2) Throw error
3) Your option

denis-tingajkin:
depends on https://github.com/networkservicemesh/networkservicemesh/pull/1569

PR #1610, opened 2019-09-13
Stabilization of linter

denis-tingajkin:

## Motivation and Context
Excludes case: https://circleci.com/gh/networkservicemesh/networkservicemesh/99232

## How Has This Been Tested?

- [X] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,7 +60,7 @@ jobs:
           name: lint
           command: |
             ./scripts/lint-download.sh
-            make lint-check-diff
+            ./scripts/lint-check-diff.sh
     environment:
       GO111MODULE: "on"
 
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -60,7 +60,7 @@ jobs:
           name: lint
           command: |
             ./scripts/lint-download.sh
-            make lint-check-diff
+            ./scripts/lint-check-diff.sh
```
artem-belov: Could you please keep `make lint-check-diff` here

denis-tingajkin: I think we should not keep `make lint-check-diff` there, because the 
1) the script runs golangci lint with memory limitation. We can't decide for user what the memory config he should to use. 
2) the bash script has attempts. Attempts is not need for the user.

@artem-belov let me know if I missed something.

edwarnicke:
This is fixing a linter issue, and the failed tests are unrelated to it.  Going to merge it against failed CI because its helping us get back to more stable CI.

PR #1638, opened 2019-09-18
Updating go version to 1.13

lobkovilya:

## Motivation and Context

`GOPROXY` that we use doesn't have `spiffe/spire` dependency. Go 1.13 has extremely useful feature that allows set `GOPROXY` as comma separated list:
```
GOPROXY=https://proxy.golang.org,direct
``` 
In that case rare dependencies will be downloaded directly.
## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


controlplane/build/Dockerfile.nsmd
```diff
@@ -1,8 +1,9 @@
-FROM golang:alpine as build
-
 ARG VERSION=unspecified
 ARG VENDORING
 ARG GOPROXY
+ARG GO_VERSION
+
+FROM golang:${GO_VERSION}-alpine as build
```
alex-yust: How about extracting the whole 'golang:${GO_VERSION}-alpine' thing as an ARG somethinglike BASE_IMAGE?

.circleci/config.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     resource_class: xlarge
     working_directory: /go/src/github.com/networkservicemesh/networkservicemesh
     docker:
-      - image: circleci/golang:1.12-stretch
```
haiodo: Maybe we could use some env variable to put version inside?

haiodo: On other hand, this values here show us a version used, and if will be some env variable in CI we will not know it value.

controlplane/build/Dockerfile.nsmd
haiodo: I'm fine with just go version to be passed.

PR #1640, opened 2019-09-18
Fix the ./scripts/switch_k8s_version.sh to work in a multi-mod world.

edwarnicke:
Also restore missing comment in ./go.mod, k8s/go.mod, test/go.mod, and controllers/sriov-controller/go.mod

### DIFF
--- a/controllers/sriov-controller/go.mod
+++ b/controllers/sriov-controller/go.mod
@@ -13,6 +13,7 @@ require (
 )
 
 replace (
+	// ./scripts/switch_k8s_version.sh to change k8s version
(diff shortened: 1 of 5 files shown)


edwarnicke:
@nickolaev I'm happy to roll the 1.16.0 upgrade in here if you'd like, or in a separate PR.  Just let me know :)

edwarnicke:
@nickolaev So I think we've discovered that it will be more work to transition to 0.16.0 ... want to merge this so when the time comes this part of the mechanics is readily at hand?

Issue #1590, opened 2019-09-10
NSMD-k8s and plugins should not crash.


haiodo:
## Expected Behavior

NSMD-k8s and plugins should not crash.

## Current Behavior

I have a crash of nsmd-k8s, after running nsmgr for  along period (~1h)
after crashing and restarting, nsmd-k8s tells: level=fatal msg="Failed to start K8s Plugin rpc error: code = Unknown desc = already have a plugin with the same name"

## Steps to Reproduce

1. Step 1
2. Step 2
3. You get the idea...

## Context

* Kubernetes Version:
* etc.

matrohon:
here is the failure reported by nsmd-k8s: 

```
panic: close of closed channel

goroutine 113 [running]:
github.com/networkservicemesh/networkservicemesh/k8s/pkg/prefixcollector.watchSubnet.func1(0xc0003f71a0, 0x15ac5a0, 0xc000077b80, 0xc0002645a0, 0xc000189390, 0xc000189380, 0xc0004b6b70, 0xc00000c7c8)
	/root/networkservicemesh/k8s/pkg/prefixcollector/collector_server.go:181 +0x64a
created by github.com/networkservicemesh/networkservicemesh/k8s/pkg/prefixcollector.watchSubnet
	/root/networkservicemesh/k8s/pkg/prefixcollector/collector_server.go:173 +0xf6
```

it then automatically restarts by fails with the following error:

```
time="2019-09-10T13:09:58Z" level=info msg="Receive resource: name kube-master, subnet 10.42.0.0/24"
2019/09/10 13:09:58 Reporting span 683338a03285cfb2:683338a03285cfb2:0:1
time="2019-09-10T13:09:58Z" level=fatal msg="Failed to start K8s Plugin rpc error: code = Unknown desc = already have a plugin with the same name"
```

matrohon:
To me, there are two issues:
- the panic
- the incapacity of registering a plugin that has failed and restarted

matrohon:
I rebased and let nsmgr run several hours, and this kind of panic is happening:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x503056]

goroutine 90 [running]:
net.networkNumberAndMask(0x0, 0xc000248e08, 0xc000248eb8, 0xc000023f50, 0xc0000bde88, 0x1, 0x1)
[...]

PR #1641, opened 2019-09-19
WIP: Plugin re-registration after nsmd recovery

alex-yust:
## Motivation and Context

#1590 

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


controlplane/pkg/plugins/utils.go
```diff
@@ -38,6 +39,32 @@ func StartPlugin(name, registry string, services map[plugins.PluginCapability]in
 		return err
 	}
 
+	go func() {
+		registered := true
+		for ; ; time.Sleep(5 * time.Second) {
```
lobkovilya: Please move `5 * time.Second` to constant

controlplane/pkg/plugins/utils.go
```diff
@@ -38,6 +39,32 @@ func StartPlugin(name, registry string, services map[plugins.PluginCapability]in
...
+			resp, err := http.Get("http://localhost:5555/liveness")
```
lobkovilya: Let's create something like `probesClient` in package `probes` to avoid dependencies on specific port and URL path

controlplane/pkg/plugins/utils.go
```diff
@@ -38,6 +39,32 @@ func StartPlugin(name, registry string, services map[plugins.PluginCapability]in
...
+			if err != nil {
+				logrus.Errorf("failed to get NSMD liveness: %v", err)
+			}
+
+			nsmdAlive := err == nil && resp.StatusCode >= 200 && resp.StatusCode < 400
```
lobkovilya: We consider success only in case of `StatusCode` equals to 200

controlplane/pkg/plugins/utils.go
```diff
@@ -38,6 +39,32 @@ func StartPlugin(name, registry string, services map[plugins.PluginCapability]in
...
+				} else {
+					logrus.Infof("Successfully re-registered plugin '%v' at '%v'", name, endpoint)
+					registered = true
+				}
+			}
+			if !nsmdAlive && registered {
+				logrus.Error("NSMD failure detected, waiting for it to restart...")
+				registered = false
+			}
+			// nsmdAlive && registered: all is well, do nothing
```
lobkovilya: I think that comments quite obvious, variables names are comprehensive enough

controlplane/pkg/plugins/utils.go
```diff
@@ -38,6 +39,32 @@ func StartPlugin(name, registry string, services map[plugins.PluginCapability]in
 		return err
 	}
 
+	go func() {
```
lobkovilya: I'd move the whole goroute func into separate func with name `monitorNSMD` or something like that

controlplane/pkg/plugins/utils.go
denis-tingajkin: can we also use time.After()?

controlplane/pkg/plugins/utils.go
alex-yust: - [ ] Only allow 200 return code for liveness probe

controlplane/pkg/plugins/utils.go
alex-yust: - [ ] remove obvious comments

controlplane/pkg/plugins/utils.go
denis-tingajkin: also, I guess we can create env variable for plugins to add possible edit the check period.

test/integration/basic_nsmd_deploy_restart_test.go
```diff
@@ -69,6 +69,10 @@ func TestNSMgrRestartDeploy(t *testing.T) {
 	g.Expect(restarts).To(Equal(int32(1)))
 	g.Expect(<-errChan).To(BeNil())
 	g.Expect(<-prevLogsChan).To(ContainSubstring("SIGABRT: abort"))
+
+	logrus.Info("Waiting for k8s plugin to re-register in restarted NSMD...")
+	err = k8s.WaitLogsContainsRegex(nodesConf[0].Nsmd, "nsmd-k8s", "Successfully re-registered plugin", fastTimeout)
```
denis-tingajkin: please consider not use fast timeout here. Log api is slow in kubernetes and it could be a reason of test failing.

controlplane/pkg/plugins/utils.go
alex-yust: - [ ] move the whole goroute func into separate func with name monitorNSMD

controlplane/pkg/plugins/utils.go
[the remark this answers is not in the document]
alex-yust: In this case, there would be nested 'if registered {}' to output "NSMD failure detected.." message. which complicates readability imho

test/integration/basic_nsmd_deploy_restart_test.go
alex-yust: - [ ] use defaultTimeout

controlplane/pkg/plugins/utils.go
alex-yust: - [ ] move 5 * time.Second to constant

alex-yust: > also, I guess we can create env variable for plugins to add possible edit the check period.

I (and Ilya) have doubts that the option would ever be used

denis-tingajkin: @alex-yust , @lobkovilya can we use this env variable for testing? We could set specific time for other tests and set for this test short period. What do you think?

controlplane/pkg/plugins/utils.go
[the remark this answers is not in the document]
denis-tingajkin: can we simply use log.Error("....", status Code, err, registered)... and not use nested if blocks?

controlplane/pkg/plugins/utils.go
alex-yust: @denis-tingajkin do you mean env variable for 'timeout' (like fastTimeout) or for 'check period' (5 second constant here in the code)?

denis-tingajkin: @alex-yust oops, sorry, I meant period.

controlplane/pkg/plugins/utils.go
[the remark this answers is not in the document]
alex-yust: Then It will report to log every 'check period' which might be a lot of text, is it Ok?
Currently, 'NSMD failure detectedNSMD failure detected' is output only once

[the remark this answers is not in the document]
denis-tingajkin: It will not report every 'check period', It will report only if we have error or response status is not 200

controlplane/pkg/plugins/utils.go
alex-yust: - [ ]  use time.After()

controlplane/pkg/plugins/utils.go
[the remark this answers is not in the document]
alex-yust: Sorry, I meant it will report every 'check period' during nsmd failure, like 5 minutes, for example.
is it Ok?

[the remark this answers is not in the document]
denis-tingajkin: maybe we could use `logrus.Fatal` after timeout in 5 minutes? What do you think?

controlplane/pkg/plugins/utils.go
```diff
@@ -97,3 +101,27 @@ func registerPlugin(name, endpoint, registry string, capabilities []plugins.Plug
 
 	return err
 }
+
+func monitorNSMD(name string, endpoint string, registry string, capabilities []plugins.PluginCapability) {
+	registered := true
+	for ; ; time.After(livenessCheckPeriod) {
```
denis-tingajkin: please use `<-time.After(...)`

alex-yust: - [ ] use <-time.After(...)

controlplane/pkg/plugins/utils.go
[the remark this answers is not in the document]
alex-yust: Fatal means we should terminate monitoring gorutine after 5 minutes of inaccessible nsmd?
Are there some benefits for implementing such a complicated behavior?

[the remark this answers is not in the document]
denis-tingajkin: My point is that we could keep it simple. 
I mean the checks at lines 114 and 122.
This is minor comment and can be ignored.

controlplane/pkg/plugins/utils.go
```diff
@@ -97,3 +101,27 @@ func registerPlugin(name, endpoint, registry string, capabilities []plugins.Plug
...
+		resp, err := http.Get("http://localhost:5555/liveness")
```
denis-tingajkin: can we get rid of using global client?

alex-yust: > can we get rid of using global client?

@denis-tingajkin could you please elaborate on this?

controlplane/pkg/plugins/utils.go
alex-yust: CheckNsmdLiveness() function is defined in probes package to encapsulate the code for checking NSMD liveness.

controlplane/pkg/plugins/utils.go
denis-tingajkin: @alex-yust http.Get() uses global client. We could 
initialize and use instance of htpp client before cycle.

alex-yust: - [ ] use local http.Client

controlplane/pkg/plugins/utils.go
```diff
@@ -97,3 +101,26 @@ func registerPlugin(name, endpoint, registry string, capabilities []plugins.Plug
 
 	return err
 }
+
+func monitorNSMD(name string, endpoint string, registry string, capabilities []plugins.PluginCapability) {
```
denis-tingajkin: please, consider to abstact the monitoring staff and to put it into probes? I think it could be reuseful. For example
probes.StartMonitoring(..., reconnect func() error)

pkg/probes/probes.go
```diff
@@ -82,3 +83,46 @@ func (probes *probesImpl) BeginHealthCheck() {
 		}
 	}()
 }
+
+// ProbeChecker convenient checker for standard k8s probes liveness/readiness
+type ProbeChecker interface {
+	Alive() (bool, error)
```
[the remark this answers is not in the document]
alex-yust: Why not, error can be easily ignored in client code, and it's a common thing in Go to return an error alongside the main result.

controlplane/pkg/plugins/utils.go
```diff
@@ -97,3 +101,26 @@ func registerPlugin(name, endpoint, registry string, capabilities []plugins.Plug
...
+	registered := true
+	probeChecker := probes.NewNsmdProbeChecker()
+	for ; ; <-time.After(livenessCheckPeriod) {
+		nsmdAlive, err := probeChecker.Alive()
```
denis-tingajkin: I think we should use probeChecker.Ready() here

pkg/probes/probes.go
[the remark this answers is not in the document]
denis-tingajkin: yes, you are right, but in your code you use the error only for log. We actually need only to know is probe server is alive. I think we can also log the error insede Alive(), Ready() methods.

pkg/probes/probes.go
```diff
@@ -26,6 +26,7 @@ import (
 
 const (
 	healthcheckProbesPort = "0.0.0.0:5555"
+	nsmdProbeURL          = "http://localhost:5555"
```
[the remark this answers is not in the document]
alex-yust: proxy nsmd, nsmd, admission webhook - do they always have (and will have) same url?
If that is the case, should we call the constant something like nsmPodProbeURL?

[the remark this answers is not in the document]
denis-tingajkin: they using same url at  this moment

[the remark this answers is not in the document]
alex-yust: And it's unlikely they will have different url in the foreseeable future, because they are in one pod, right?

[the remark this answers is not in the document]
denis-tingajkin: 1) they are in different pods
2) probe server startes on localhost
3) In future we can change the port only but it is not related to current task, I guess

controlplane/pkg/plugins/utils.go
alex-yust: - [ ] check probeChecker.Ready() before registration

controlplane/pkg/plugins/utils.go
denis-tingajkin: in addition, it will be possible to cover unit tests

pkg/probes/probes.go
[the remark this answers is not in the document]
alex-yust: The current code allows the client code to decide whether to report an error. Otherwise there will be either no diagnostic or too much of it (since probes are checked on regular basis).

pkg/probes/probes.go
```diff
@@ -82,3 +83,46 @@ func (probes *probesImpl) BeginHealthCheck() {
...
+	return &probeCheckerImpl{
+		probeRootURL: hostURL,
+		httpClient: http.Client{
+			Timeout: 10,
+		},
+	}
+}
+
+// NewNsmdProbeChecker creates a NSMD probe checker
+func NewNsmdProbeChecker() ProbeChecker {
```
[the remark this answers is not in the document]
alex-yust: @denis-tingajkin please explain: I don't understand why a probe checker to track NSMD liveness should be created with function named NewDefaultProbeChecker(), the name seems not descriptive.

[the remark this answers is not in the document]
denis-tingajkin: @alex-yust sure, the probes package should not know about `nsmd`. It is a tool and can be used in different places.

pkg/probes/probes.go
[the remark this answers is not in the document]
denis-tingajkin: > Otherwise there will be either no diagnostic or too much of it (since probes are checked on regular basis).
But you already log it always, take a look at https://github.com/networkservicemesh/networkservicemesh/pull/1641/files#diff-b305f1bf4a29da22b05096c2f948b079R111
Or I missed someting?

controlplane/pkg/plugins/utils.go
lobkovilya: I don't think that adding new abstraction is good idea while we have only one usage of it. We can do it in further refactoring in case of facing logical duplications

pkg/probes/probes.go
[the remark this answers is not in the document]
alex-yust: - [x] do not return err from ProbeChecker methods

pkg/probes/probes.go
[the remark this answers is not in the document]
alex-yust: NewNsmdProbeChecker() is removed, NewProbeChecker() is used

pkg/probes/probes.go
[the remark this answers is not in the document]
alex-yust: nsmdProbeURL constant is removed

pkg/probes/probes.go
```diff
@@ -19,13 +19,14 @@ import (
...
-
 	"github.com/sirupsen/logrus"
+
+	"github.com/networkservicemesh/networkservicemesh/pkg/probes/health"
 )
 
 const (
-	healthcheckProbesPort = "0.0.0.0:5555"
+	probePort             = 5555
+	healthcheckProbesPort = "0.0.0.0:" + string(probePort)
```
denis-tingajkin: string(int) will convert int to unicode symbol...

denis-tingajkin: please make probePort as string

alex-yust: - [ ] use fmt.Sprintf() to build healthcheckProbesPort

alex-yust: - [ ] assign probePort as string

pkg/probes/probes.go
```diff
@@ -19,13 +19,14 @@ import (
...
-
 	"github.com/sirupsen/logrus"
+
+	"github.com/networkservicemesh/networkservicemesh/pkg/probes/health"
 )
 
 const (
-	healthcheckProbesPort = "0.0.0.0:5555"
+	probePort             = "5555"
+	healthcheckProbesPort = "0.0.0.0:%d" + probePort
```
denis-tingajkin: please remove `%d`. healthcheckProbesPort value is `0.0.0.0:%d5555`

alex-yust:
Alas, turns out nsmd container restart in about 1 second:
nsmd-k8s:
```
time="2019-09-20T19:35:58+07:00" level=info msg="Executing: [kill -6 1] in pod nsmgr-kube-master:nsmd"
time="2019-09-20T19:35:59+07:00" level=info msg="Kill status   <nil>"
```
nsmd:
```.............
time="2019-09-20T12:36:00Z" level=info msg="All connections are recovered..."
time="2019-09-20T12:36:00Z" level=info msg="Dataplane is available, waiting for initial state received and processed..."
time="2019-09-20T12:36:00Z" level=info msg="Dataplane server is ready"
time="2019-09-20T12:36:00Z" level=info msg="Public listener is ready"
time="2019-09-20T12:36:00Z" level=info msg="NSM gRPC API Server: [::]:5001 is operational"
time="2019-09-20T12:36:00Z" level=info msg="Serve api is ready"```

... which makes liveness probes useless in detection of nsmd restart unless polling delay is less than 1s

alex-yust:
Another PR will be opened since the fix is re-written via new approach

PR #1653, opened 2019-09-20
Fix guide-build.md

Tutkovics:
helm-install-nsm --> make helm-install-nsm
Need to add _make_ to beginning of line. (x2)


edwarnicke:
@Tutkovics Welcome!  Thank you for the doc fix.  Could you add a signed off by line to pass the DCO?

git commit -s --amend
git push -f

should do it :)

PR #1665, opened 2019-09-25
Plugin re-registration after nsmd restarts take 2 (grpc version)

alex-yust:
## Motivation and Context

#1590 

## How Has This Been Tested?

- [ ] Covered by existing integration testing
- [x] Added integration testing to cover
- [ ] Tested locally
- [ ] Have not tested

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] #1670

Issue #1592, opened 2019-09-10
Floating Interdomain


edwarnicke:
## Overview

The full promise of interdomain requires being able to have a 'floating' NSM domain that is free from any specific runtime domain (K8s cluster), but in which NSEs/NSCs from any domain may participate.

![Floating Interdomain Supporting Diagrams](https://user-images.githubusercontent.com/1455208/64615974-2b3b1b00-d3a1-11e9-8331-e0b1d554ca71.png)

This requires a Network Service Registry for the 'floating' domain that is *not* the kube-api-server that is backing a particular K8s cluster.

## Implementation

kube-api-server is capable of running independently, without its associated controllers, as a simple workload.  In this form, it acts as a simple CRD store.  The suggestion here would be to run a kube-api-server, with the nsmd-k8s instances to expose the GRPC api pointing to *that* kube-api-server rather than the default one for the cluster.  

![Floating Interdomain Supporting Diagrams (1)](https://user-images.githubusercontent.com/1455208/64617148-9128a200-d3a3-11e9-9e94-3304ff5c3e42.png)

## HA

HAs can be achieved by running multiple kube-api-servers as an HA cluster, and fronting them with a Service.  Then have replicas of nsmd-k8s running that address that Service, and themselves expose the GRPC API via a Service.

![Floating Interdomain Supporting Diagrams (2)](https://user-images.githubusercontent.com/1455208/64617178-9f76be00-d3a3-11e9-9130-8f41327f0234.png)

PR #1670, opened 2019-09-26
Liveness monitor server

artem-belov:

## Description
Liveness monitor throws error in Error channel on losing connection 

## Motivation and Context
Required for Floating Intedomain (#1592) and plugins re-registration (#1665)

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


edwarnicke:
Could you explain the purpose in adding this?

artem-belov:
@edwarnicke 
For Floating interdomain we need to monitor nsmd from remote registry server to be sure that all registered NSEs are available for connection.
For #1665 we need to monitor plugin to start re-register process on plugins's restart

edwarnicke:
@artem-belov OK... so you are trying to figure out if an NSE that has registered has gone away.  Liveliness probing across domains is not a scalable way to do that.  It also hits all kinds of problems when you have semi-isolated connectivity domains.

Instead of doing liveliness probes... we should think about properly expiring NSE registrations.

One way of doing so would be to add an 'ExpiresAt' field to [NetworkServiceEndpoint](https://github.com/networkservicemesh/networkservicemesh/blob/master/controlplane/api/registry/registry.proto#L31) in the registry.  We could then 

1.  Not select expired NSEs
2. Have something that reaps expired NSEs from the registry.

alex-yust:
@edwarnicke the livemonitor is still a nice solution to monitor service liveness at least inside domain, since it takes only 1 line of code to get a liveness server on a grpc connection (please look at #1665) 
and few lines of code to monitor it.

PR #1676, opened 2019-09-28
Fix: Cloudtest sometimes does not save shell script's stdout/stderr

denis-tingajkin:
## Description
In the investigation of failing https://circleci.com/gh/networkservicemesh/networkservicemesh/105477 I found that sometimes cloudtest shell script result can be not saved.  
Root cause: we using async read from stdout/stderr and in the case when the command exits we can lose the output.

## Motivation and Context
https://circleci.com/gh/networkservicemesh/networkservicemesh/105477

## How Has This Been Tested?

- [X] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/test/cloudtest/pkg/utils/shell_utils_test.go
+++ b/test/cloudtest/pkg/utils/shell_utils_test.go
@@ -1,0 +1,21 @@
+package utils
+
+import (
+	"bufio"
+	"context"
+	"fmt"
+	"strings"
+	"testing"
+
+	"github.com/onsi/gomega"
+)
+
+func TestProcessOutputShouldNotLostOutput(t *testing.T) {
+	assert := gomega.NewWithT(t)
+	const expected = "output..."
+	for i := 0; i < 1000; i++ {
+		output, err := RunCommand(context.Background(), fmt.Sprintf("echo \"%v\"", expected), "", func(s string) {}, bufio.NewWriter(&strings.Builder{}), nil, nil, true)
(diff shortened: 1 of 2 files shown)


test/cloudtest/pkg/utils/shell_utils_test.go
```diff
@@ -0,0 +1,21 @@
...
+	"strings"
+	"testing"
+
+	"github.com/onsi/gomega"
+)
+
+func TestProcessOutputShouldNotLostOutput(t *testing.T) {
+	assert := gomega.NewWithT(t)
+	const expected = "output..."
+	for i := 0; i < 1000; i++ {
```
haiodo: How long it take to execute this kind of test?

denis-tingajkin: ~3s
`<testcase classname="github.com/networkservicemesh/networkservicemesh/test/cloudtest/pkg/utils" name="TestProcessOutputShouldNotLostOutput" time="2.630000"/>` from JunitResult

haiodo: I think it is not required to eat almost 3 seconds for such not important case.

denis-tingajkin: @haiodo What the limit time should be per unit test?

denis-tingajkin: finally reduced time to 1s

Issue #1659, opened 2019-09-23
Inconsistent behavior when multiple DNS servers provided to fanout


john-a-joyce:
## Expected Behavior

Expecting the coredns/fanout can deliver a consistent DNS response when multiple servers are configured.  

## Current Behavior

Querying for the same host will yield inconsistent results.  Sometimes they return a successful response.   Sometimes a failure and sometimes it seems partially successful.   

## Steps to Reproduce

1. Create services in two separate DNS repositories.  In my case 1 was the local kube-dns and the other was kube-dns in a KIND cluster. 
2.  Configure the fanout with multiple DNS servers.   In my case it was configured like this:
''`kubectl logs helloworld-v1-99cc49865-k9qfx -c nsm-coredns
Starting nsm-coredns
Version: 75502fff
. {
        log
        fanout 10.96.0.10 10.87.49.210:30853
}
'''
3.  Kubectl exec into an pod with the nsm-coredns container active
4.  Use nslookup to check for entries.  "busybox" is in kube-dns "busyboy" is in KIND.

Some Results:
  
root@helloworld-v1-99cc49865-k9qfx:/# nslookup busyboy  
Server:         127.0.0.1
Address:        127.0.0.1#53

** server can't find busyboy: NXDOMAIN

root@helloworld-v1-99cc49865-k9qfx:/# nslookup busyboy
Server:         127.0.0.1
Address:        127.0.0.1#53

Name:   busyboy.default.svc.cluster.local
Address: 10.97.146.183

root@helloworld-v1-99cc49865-k9qfx:/# nslookup busybox
Server:         127.0.0.1
Address:        127.0.0.1#53

** server can't find busybox: NXDOMAIN

root@helloworld-v1-99cc49865-k9qfx:/# nslookup busybox
Server:         127.0.0.1
Address:        127.0.0.1#53

Name:   busybox.default.svc.cluster.local
Address: 10.105.73.86

## Context

As noted the most important thing is have two independent DNS servers configured and provide to the nsm-coredns container.   
  
Kubernetes Version:  1.13
NSM version is from master but has some vl3 related changes that have not been committed.  

## Failure Logs

I can provide more complete logs from the nsm-coredns container if that is useful.

denis-tingajkin:
@john-a-joyce, thanks for the report. Interesting why sometimes you have a negative result... 

Some thoughts about negative result reason: 
[...]

PR #1690, opened 2019-09-30
Fix: Inconsistent behavior when multiple DNS servers provided to fanout

denis-tingajkin:

## Motivation and Context
https://github.com/networkservicemesh/networkservicemesh/issues/1659

## How Has This Been Tested?

- [X] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


john-a-joyce:
@denis-tingajkin - do you have a pointer to an nsm-coredns image that I can try directly without rebuilding locally?

denis-tingajkin:
@john-a-joyce no, I don't have the image.
If you want to test then you need to rebuild the image locally based on these changes.

john-a-joyce:
@denis-tingajkin - I tested this change in my environment by CHERRY-PICKING these two commits.  I want to emphasize the cherry pick as I was able to test an image build from your branch without corrupting my environment.   The behavior is better but I noticed a slightly different issue.   I will append the details to the issue

denis-tingajkin:
@john-a-joyce thanks for testing. It can be fixed with the last changes. Please check if you have a chance.

k8s/cmd/nsm-coredns/plugin/fanout/fanout.go
```diff
@@ -67,40 +67,79 @@ func (f *Fanout) ServeDNS(ctx context.Context, w dns.ResponseWriter, m *dns.Msg)
...
+			}
+			if count == 0 {
+				return result
+			}
+			if r.err != nil {
+				break
+			}
+			if r.response.Rcode != dns.RcodeSuccess {
+				break
+			}
```
lobkovilya: Why do we need this checks? I think we can add `else` branch to 
```go
if isBetter(result, &r) {
 	result = &r
} else {
	break
}
```

denis-tingajkin: >Why do we need this checks? 

Because `isbetter` compares two connection results. After that, we need to verify that the potential result is positive to return it instantly.

lobkovilya: I mean the case when `isBetter` returns false. In that case we won't update `result`, so checking of `r` is useless

denis-tingajkin: if `isBetter` is false we will break on line 109-114. The problem of your proposal is if `isBetter` is true and it is not positive answer. For this case, we need to check on 109-114 liens.

lobkovilya: If `isBetter` is true and it is "not positive answer" then why do we put it into `result` before checks?

denis-tingajkin: because we should return a negative answer, for example,  if before the result was a connection error.

denis-tingajkin: it is important for dns

lobkovilya: Then I don't understand what "not positive answer" you are talking about. Seems like if `isBetter` returns `true` then we can update `result` and  go to the next `for` iteration

denis-tingajkin: positive answer is DNS message witch has RCODE=0

PR #1626, opened 2019-09-16
NSMgr SDK like refactoring

haiodo:
## Description

Changes contains a huge refactoring of NSMgr connection establish/close functionality. 
Now it uses SDK approach and separated into individual chain elements, every element implements `NetworkService` protocol and do own small part of work.

## How Has This Been Tested?

- [x] Covered by existing integration testing
- [ ] Added integration testing to cover
- [x] Tested locally
- [ ] Have not tested

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #1733, opened 2019-10-08
Security #4: JWT 

lobkovilya:
## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


haiodo:
I think it should go after #1626 since it changes nsm.go, mostly refactored by me and moved into chains :-)

controlplane/pkg/nsm/nsm.go
```diff
@@ -295,6 +298,16 @@ func (srv *networkServiceManager) request(ctx context.Context, request networkse
 	// 11. We are done with configuration here.
 	logrus.Infof("NSM:(11-%v) Request done...", requestID)
 
+	if tools.GetConfig().SecurityProvider == nil {
+		logrus.Warnf("insecure: return connection without signature: %v", err)
+		return cc.GetConnectionSource(), nil
+	}
+
+	if err := security.SignConnection(
```
alex-yust: The continuation of condition line 'merges' with body, 
please consider to extract the assignment out of 'if' statement.

controlplane/api/local/connection/connection.proto
```diff
@@ -50,6 +50,7 @@ message Connection {
   connectioncontext.ConnectionContext context = 4;
   map<string, string> labels = 5;
   State state = 6;
+  string ResponseJWT = 7;
```
edwarnicke: Could we call this ResponseToken so was to not make it so JWT specific?

pkg/go.mod
```diff
@@ -22,3 +22,11 @@ require (
 )
 
 replace github.com/census-instrumentation/opencensus-proto v0.1.0-0.20181214143942-ba49f56771b8 => github.com/census-instrumentation/opencensus-proto v0.0.3-0.20181214143942-ba49f56771b8
+
```
edwarnicke: Please don't pull controlplane/api as a dependency into pkg

controlplane/api/local/connection/connection_helpers.go
```diff
@@ -137,3 +137,11 @@ func (c *Connection) IsComplete() error {
 
 	return nil
 }
+
+func (c *Connection) GetSignature() string {
```
denis-tingajkin: As I know JWT has 3 parts:
Header
Payload
Signature

Is this method should return part of JWT or can we rename it?

controlplane/pkg/common/context.go
```diff
@@ -270,3 +272,20 @@ func WorkspaceName(ctx context.Context) string {
...
+	if parent == nil {
+		parent = context.Background()
+	}
+	return context.WithValue(parent, securityContextKey, sc)
+}
+
+func SecurityContext(ctx context.Context) security.Context {
+	value := ctx.Value(securityContextKey)
+	if value == nil {
+		logrus.Info("IT IS NIL")
```
denis-tingajkin: :)

controlplane/pkg/remote/security_service.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+}
+
+func (s *securityService) Request(ctx context.Context, request *networkservice.NetworkServiceRequest) (*connection.Connection, error) {
+	span := spanhelper.GetSpanHelper(ctx)
+	sc := security.NewContext()
+
+	nctx := common.WithSecurityContext(ctx, sc)
+	nsc := common.SecurityContext(nctx)
+	logrus.Infof("remote.New SC = %p", nsc)
```
denis-tingajkin: should we use a logger from context?

controlplane/pkg/remote/security_service.go
```diff
@@ -0,0 +1,78 @@
...
+		provider: provider,
+	}
+}
+
+func (s *securityService) Request(ctx context.Context, request *networkservice.NetworkServiceRequest) (*connection.Connection, error) {
+	span := spanhelper.GetSpanHelper(ctx)
+	sc := security.NewContext()
+
+	nctx := common.WithSecurityContext(ctx, sc)
+	nsc := common.SecurityContext(nctx)
```
denis-tingajkin: can we use `sc` instead of `nsc` here?

pkg/security/context.go
```diff
@@ -0,0 +1,22 @@
...
+type Context interface {
+	GetResponseOboToken() string
+	SetResponseOboToken(token string)
+}
+
+type contextImpl struct {
+	responseOboToken string
+}
+
+func NewContext() Context {
```
denis-tingajkin: probably we can get rid of this structure. Just simply put two strings into context via context.WithValue()

controlplane/api/local/connection/connection_helpers.go
```diff
@@ -137,3 +137,11 @@ func (c *Connection) IsComplete() error {
...
+	return c.GetResponseJWT()
```
denis-tingajkin: As I know JWT has 3 parts:
Header
Payload
Signature

Is this method should return part of JWT or can we rename it?

lobkovilya: The name of that method doesn't imply dependency on some specific technology that being used to implement that signature.

controlplane/pkg/common/context.go
lobkovilya: I know that I invited everyone to review it regardless PR is draft, but now it is in draft for real, because I adapt it to new unified api and SDK like refactoring. I just didn't expect that someone starts review it now :)

controlplane/pkg/remote/security_service.go
lobkovilya: yes, we should

controlplane/pkg/remote/security_service.go
lobkovilya: just for logging

pkg/security/context.go
lobkovilya: We need structure to allow `SecurityService` see `responseOboToken` that had been set by `EndpointService`

controlplane/pkg/common/context.go
denis-tingajkin: got it. Let me know when it's ready for review.

pkg/security/test/jwt_test.go
```diff
@@ -83,3 +86,45 @@ func TestChain(t *testing.T) {
 
 	Expect(security.VerifySignature(signature3, sc3.GetCABundle(), SpiffeID3)).To(BeNil())
 }
+
+func TestJWTPerformance(t *testing.T) {
```
denis-tingajkin: Probably for this kind tests we can use `testing.B` and run these tests in a separate job on CI. @edwarnicke , @lobkovilya , @haiodo What do you think?

lobkovilya: The name of test might be confusing, but in real it works about 200ms :)

denis-tingajkin: > The name of test might be confusing, but in real it works about 200ms :)

Yes, I've checked before putting a comment :) 
The main idea is to write a more pure test to track performance vie testing.B.

pkg/security/utils.go
```diff
@@ -15,12 +15,173 @@
...
+		if err != nil {
+			return nil, status.Errorf(codes.Unauthenticated, err.Error())
+		}
+
+		md, ok := metadata.FromIncomingContext(ctx)
+		if !ok {
+			return nil, status.Errorf(codes.InvalidArgument, "missing metadata")
+		}
+
+		if len(md["authorization"]) == 0 {
```
haiodo: make constant?

PR #1740, opened 2019-10-09
Reduce to a single (hopefully fast) build-containers

edwarnicke:
## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -180,7 +180,7 @@ jobs:
             export TAG="${COMMIT}"
             export CONTAINER_REPO="networkservicemeshci"
             export GOPROXY=https://proxy.golang.org,direct
-            make -j 8 docker-push
+            make docker-push
 
 # docker push
   docker-push-tag:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -165,10 +165,8 @@ jobs:
       NSM_NAMESPACE: "nsm-system"
 
 # build
-  build-container:
-    parameters:
-      container:
-        type: string
+  build-containers:
+    resource_class: xlarge
     docker:
       - image: circleci/golang:1.13-stretch
     steps:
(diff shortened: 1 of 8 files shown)


.circleci/config.yml
```diff
@@ -182,17 +180,7 @@ jobs:
...
-
-# build ready - empty job
-  build-container-success:
-    docker:
-      - image: circleci/golang:1.13-stretch
-    steps:
-      - run:
-          command: |
-            echo "SUCCESS"
+            make -j 8 docker-push
```
alex-yust: We can use just '-j' option (no concrete number) complemented by load restriction option '-l'.
Also, log aggregation option '--output-sync=line' might be a useful addition
https://ftp.gnu.org/old-gnu/Manuals/make-3.79.1/html_chapter/make_toc.html#TOC47

.mk/k8s.mk
```diff
@@ -56,7 +56,7 @@ SPIRE_ENABLED=false
 endif
 
 kubectl = kubectl -n ${NSM_NAMESPACE}
-images_tar = $(subst .tar,,$(filter %.tar, $(shell ls ./scripts/vagrant/images)))
+images_tar = $(subst .tar,,$(filter %.tar, $(shell mkdir -p ./scripts/vagrant/images;ls ./scripts/vagrant/images)))
```
haiodo: Actually I think we need to move from using of ./scripts/**vagrant** folder as a place for docker images

alex-yust: > we need to move from using of ./scripts/vagrant folder as a place for docker images

+1
We might also want a variable in build.mk for this, like IMAGE_DIR

.circleci/config.yml
edwarnicke: The physical machine underlying has 8 cores.  I briefly experimented with -j 16, but it didn't help much.  If you have more advice on how to tune, I'd love to take it :)

As to --output-sync=line ... this is a very good idea... why line and not target though?

.mk/k8s.mk
edwarnicke: Yeah... we need a better way here... I took the path of the minimum fix that actually fixed the breakage in order to get to consolodated builds.  I'm very open to a better approach to that.

What I *do* like about the images_tar line is that it attempts to dynamically figure out the images_tars ... that's a very good trend in our Makefiles :)

PR #1768, opened 2019-10-15
Add only-run to cloudtest

edwarnicke:
If you add only-run to .cloudtest configuration it will only run the listed tests.

Super useful for running a small subset of tests that are failing in order to debug quickly.

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/test/cloudtest/pkg/commands/main.go
+++ b/test/cloudtest/pkg/commands/main.go
@@ -1056,7 +1056,17 @@ func (ctx *executionContext) findGoTest(executionConfig *config.ExecutionConfig)
 	for _, t := range execTests {
 		t.Kind = model.TestEntryKindGoTest
 		t.ExecutionConfig = executionConfig
-		result = append(result, t)
+		match := true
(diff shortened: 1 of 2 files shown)


test/cloudtest/pkg/commands/main.go
```diff
@@ -1056,7 +1056,17 @@ func (ctx *executionContext) findGoTest(executionConfig *config.ExecutionConfig)
 	for _, t := range execTests {
 		t.Kind = model.TestEntryKindGoTest
 		t.ExecutionConfig = executionConfig
-		result = append(result, t)
+		match := true
```
alex-yust: match := len(executionConfig.OnlyRun) == 0
?

edwarnicke: Overall a good idea... but in the event len(executionConfig.OnlyRun) == 0... I want us to run *all* the tests.  So if we have an empty only-run list... just run everything :)

PR #1770, opened 2019-10-15
fixes fuzzing ParseGatewayIP

### LINKED ISSUE #1615: Fuzzing: panic in parseGatewayIP
## Current Behavior

panic: runtime error: slice bounds out of range

goroutine 1 [running]:
github.com/networkservicemesh/networkservicemesh/dataplane/pkg/common.parseGatewayIP(0xfb8210, 0x1, 0x1, 0x1, 0xfb8210)
	/home/user/go/src/github.com/networkservicemesh/networkservicemesh/dataplane/pkg/common/egress_interface.go:96 +0x1c6
github.com/networkservicemesh/networkservicemesh/dataplane/pkg/common.FuzzParseGatewayIP(0x794aed72f000, 0x1, 0x1, 0x3)
	/home/user/go/src/github.com/networkservicemesh/networkservicemesh/dataplane/pkg/common/egress_interface_fuzz.go:25 +0xb1
go-fuzz-dep.Main(0xc0000ddf78, 0x2, 0x2)
	/tmp/go-fuzz-build528173389/goroot/src/go-fuzz-dep/main.go:36 +0x1b6
main.main()
	/tmp/go-fuzz-build528173389/gopath/src/github.com/networkservicemesh/networkservicemesh/dataplane/pkg/common/go.fuzz.main/main.go:17 +0x5d

goroutine 19 [syscall]:
os/signal.signal_recv(0x0)
	/tmp/go-fuzz-build528173389/goroot/src/runtime/sigqueue.go:139 +0x9c
os/signal.loop()
	/usr/share/go/src/os/signal/signal_unix.go:23 +0x52
created by os/signal.init.0
	/usr/share/go/src/os/signal/signal_unix.go:29 +0x41
exit status 2

## References

#1518 
[crashers.GatewayIP.zip](https://github.com/networkservicemesh/networkservicemesh/files/3612441/crashers.GatewayIP.zip)

pringlewood:
Signed-off-by: Wayne <wayne@flickswitch.co.za>

## Description
fixes #1615

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


dataplane/pkg/common/egress_interface.go
```diff
@@ -92,17 +96,20 @@ func parseProcFile(reader *bufio.Reader) (string, net.IP, error) {
 	return "", nil, errors.New("Failed to locate default route...")
 }
 
-func parseGatewayIP(defaultGateway string) net.IP {
+func parseGatewayIP(defaultGateway string) (net.IP, error) {
```
lobkovilya: @haiodo I'm not sure, but seems like `parseGatewayIP` could be replaced with `net.ParseIP`. Please correct me if I'm wrong

dataplane/pkg/common/egress_interface.go
```diff
@@ -92,17 +96,20 @@ func parseProcFile(reader *bufio.Reader) (string, net.IP, error) {
...
-	iv1, _ := strconv.ParseInt(defaultGateway[2:4], 16, 32)
-	iv2, _ := strconv.ParseInt(defaultGateway[4:6], 16, 32)
-	iv3, _ := strconv.ParseInt(defaultGateway[6:], 16, 32)
-	ip[0] = byte(iv3)
-	ip[1] = byte(iv2)
-	ip[2] = byte(iv1)
-	ip[3] = byte(iv0)
-	return ip
+	if len(defaultGateway) == 8 {
+		iv0, _ := strconv.ParseInt(defaultGateway[0:2], 16, 32)
```
ivanayov: I think we should get the errors from ParseInt here and check them. Otherwise we cannot guarantee if accurate value will be returned for ip. Some of the `iv` values may be `0` because of an error.

dataplane/pkg/common/egress_interface_test.go
```diff
@@ -9,10 +9,18 @@ import (
...
+
+	gw, err := parseGatewayIP("0")
+	g.Expect(err.Error()).To(Equal("Failed to parse IP from string"))
+	g.Expect(gw.IsUnspecified()).To(BeTrue())
+}
 func TestParseDefaultGateway(t *testing.T) {
 	g := NewWithT(t)
 
-	gw := parseGatewayIP("010011AC")
+	gw, err := parseGatewayIP("010011AC")
```
ivanayov: It's good to add a test case with non-accurate value, as `"010011AS"` or random string

dataplane/pkg/common/egress_interface_test.go
```diff
@@ -9,10 +9,18 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
+func TestParseGatewayZero(t *testing.T) {
+	g := NewWithT(t)
```
ivanayov: For some reason `gomega` encourages using `.`, but in general it's a bad practice. It may cause compile errors, and it's also bad for the readability, as one will hardly guess where `NewWithT` comes from.
I would advise removing the `.` and using `gomega.NewWithT`.

dataplane/pkg/common/egress_interface_test.go
pringlewood: 👋 @ivanayov thanks for the review, when parsing ```"010011AS"``` it actually parses fine to  ```0.17.0.2``` This is probably not a valid ip, but I am not sure how to deal with this as there is no validity check on ip I can raise a error with any ideas ?

I added a test for string length > 8 and fixed the assert  as it was not asserting the gateway ip string was actually the value expected.

ivanayov: @pringlewood yep, actually the error checking is enough

dataplane/pkg/common/egress_interface_test.go
pringlewood: I tried that, the tests break around the asserts.  i.e

```g.Expect(eth0).To(Equal(""))```

It complains about Equal, do you have any other pointers for me? (first time writing go :-))

Comment (ivanayov) [reply dataplane/pkg/common/egress_interface_test.go]: @pringlewood great you have started, it's pretty good one :) 
The `.` import means all the package exported identifiers will be declared in the importing source file's file block and are accessed without a qualifier, i.e. you can use all exported functions, without referencing the package.

So if you change to `gomega.Equal()`, should be fine.

File: dataplane/pkg/common/egress_interface.go (hunk 1 above)
Comment (pringlewood) [reply dataplane/pkg/common/egress_interface.go]: @lobkovilya not sure ParseIP will work here from the doc string:

```The string s can be in dotted decimal ("192.0.2.1")```

File: dataplane/pkg/common/egress_interface.go [hunk 5]
```diff
@@ -78,25 +78,36 @@ func parseProcFile(scanner *bufio.Scanner) (string, net.IP, error) {
 		if strings.TrimSpace(parts[1]) == "00000000" {
 			outgoingInterface := strings.TrimSpace(parts[0])
 			defaultGateway := strings.TrimSpace(parts[2])
-			ip := parseGatewayIP(defaultGateway)
+			ip, err := parseGatewayIP(defaultGateway)
+			if err != nil {
+				logrus.Printf("Error Processing gateway IP %v for outgoing interface: %v", defaultGateway, outgoingInterface)
```
ivanayov: Not something crucial, but there's a convention in golang to write log messages in lower case.

dataplane/pkg/common/egress_interface.go
denis-tingajkin: @pringlewood Do you mean this case https://play.golang.org/p/ZuCQ-EXbVsI?

dataplane/pkg/common/egress_interface.go
denis-tingajkin: it is related to errors

dataplane/pkg/common/egress_interface.go
```diff
@@ -78,25 +78,36 @@ func parseProcFile(scanner *bufio.Scanner) (string, net.IP, error) {
 		if strings.TrimSpace(parts[1]) == "00000000" {
 			outgoingInterface := strings.TrimSpace(parts[0])
 			defaultGateway := strings.TrimSpace(parts[2])
-			ip := parseGatewayIP(defaultGateway)
+			ip, err := parseGatewayIP(defaultGateway)
+			if err != nil {
```
denis-tingajkin: please use `logrus.Erorrf` for errors

dataplane/pkg/common/egress_interface.go
```diff
@@ -78,25 +78,36 @@ func parseProcFile(scanner *bufio.Scanner) (string, net.IP, error) {
...
 			return outgoingInterface, ip, nil
 		}
 	}
 	return "", nil, errors.New("Failed to locate default route...")
 }
 
-func parseGatewayIP(defaultGateway string) net.IP {
+func parseGatewayIP(defaultGateway string) (net.IP, error) {
+	if len(defaultGateway) != 8 {
+		return nil, errors.New("Failed to parse IP from string")
```
denis-tingajkin: please, use lower case error message

edwarnicke: @denis-tingajkin I thought we had CI enforcing that style check?  Did I misremember?

edwarnicke: @denis-tingajkin Got it :)

lobkovilya: @denis-tingajkin no, this is because `errors` now from `github.com/pkg/errors`, I guess linter just can't track such kind of issues

denis-tingajkin: @lobkovilya oops! I forgot that we using `github.com/pkg/errors` instead of `errors` for now. Yes, you can be right.

dataplane/pkg/common/egress_interface.go
pringlewood: @denis-tingajkin yup, I updated the example to try and parse a string as per the file: 

https://play.golang.org/p/j5IQ1tj_vlm

dataplane/pkg/common/egress_interface_test.go
pringlewood: Thanks updated the tests

dataplane/pkg/common/egress_interface.go
denis-tingajkin: @pringlewood got it

forwarder/pkg/common/egress_interface.go
```diff
@@ -78,25 +78,36 @@ func parseProcFile(scanner *bufio.Scanner) (string, net.IP, error) {
 		if strings.TrimSpace(parts[1]) == "00000000" {
 			outgoingInterface := strings.TrimSpace(parts[0])
 			defaultGateway := strings.TrimSpace(parts[2])
-			ip := parseGatewayIP(defaultGateway)
+			ip, err := parseGatewayIP(defaultGateway)
+			if err != nil {
+				logrus.Errorf("error processing gateway IP %v for outgoing interface: %v", defaultGateway, outgoingInterface)
+				break
```
denis-tingajkin: I see some problems in method `parseProcFile`
1) `len` of parts is not checked at line 77. We naively expect that has len >= 3
2) we can use `return` instead of a` break` at lines 70, 84 and return an actual error. Also, we can wrap it. For example `errors.Wrap(err, "failed to locate default route...")`

forwarder/pkg/common/egress_interface.go
```diff
@@ -78,25 +78,36 @@ func parseProcFile(scanner *bufio.Scanner) (string, net.IP, error) {
...
-func parseGatewayIP(defaultGateway string) net.IP {
+func parseGatewayIP(defaultGateway string) (net.IP, error) {
+	if len(defaultGateway) != 8 {
+		return nil, errors.New("failed to parse IP from string")
+	}
+	iv0, err := strconv.ParseInt(defaultGateway[0:2], 16, 32)
+	iv1, err := strconv.ParseInt(defaultGateway[2:4], 16, 32)
+	iv2, err := strconv.ParseInt(defaultGateway[4:6], 16, 32)
+	iv3, err := strconv.ParseInt(defaultGateway[6:], 16, 32)
+	if err != nil {
```
denis-tingajkin: it is a not good check because for example
Imagine the next case:
1) `iv0` has not parsed as int, `err` is not nil
2) `iv1` has  parsed as int, `err` is `nil` 

Finally, the method can return `err` as `nil`, but actually we faced a problem.

ivanayov: @denis-tingajkin I'm not sure that there can be such use-case. `iv1` can be parsed as int = `0` and then `err` will be of type `ErrSyntax`, but not `nil`. `iv1` can also be the max int value when it cannot be represented by signed int, and then the error will be `ErrRange`.
It's always `int` value, but anytime parsing failed, there should be an error.
Can you share an example?

denis-tingajkin: @ivanayov please take a look https://play.golang.org/p/HwHeMvQw291

edwarnicke:
@pringlewood Try rebasing... you appear to be hitting an issue that has been fixed on master since you started your work :)

forwarder/pkg/common/egress_interface.go
pringlewood: I added error checking so that it caters for the above example. But I don't know how to solve this in a dry way other then checking for errors after each parse of the sliced string.

forwarder/pkg/common/egress_interface.go
pringlewood: Thanks for the pointers @denis-tingajkin updated.

forwarder/pkg/common/egress_interface.go
ivanayov: @denis-tingajkin , @pringlewood The problem is that there's only one error check at the end after `iv3` and `err` is overwritten to `nill` by the last accurate value.
With a check for every value, it's working well:
https://play.golang.org/p/UsI5rrJFLA2

pringlewood: @ivanayov thanks, I implemented this.

denis-tingajkin: @ivanayov exactly, that what I meant on the first comment :)

forwarder/pkg/common/egress_interface.go
```diff
@@ -67,36 +67,59 @@ func findDefaultGateway4() (string, net.IP, error) {
...
-func parseGatewayIP(defaultGateway string) net.IP {
+func parseGatewayIP(defaultGateway string) (net.IP, error) {
+	if len(defaultGateway) != 8 {
+		return nil, errors.New("failed to parse IP from string")
+	}
+	iv0, err := strconv.ParseInt(defaultGateway[0:2], 16, 32)
+	if err != nil {
+		return nil, errors.New("string does not represent a valid IP address")
+	}
+	iv1, err := strconv.ParseInt(defaultGateway[2:4], 16, 32)
```
denis-tingajkin: @pringlewood can we use loop instead of these similar checks?
For example
```
	if len(defaultGateway) != 8 {
		return nil, errors.New("failed to parse IP from string")
	}
        ip := net.IP{}
        for i := 0; i < len(deaultGateway) /2; i++ {
                iv, err := strconv.ParseInt(defaultGateway[i * 2, i * 2 + 2], 16, 32)
                if err != nil {
                        return nil, errors.New("string does not represent a valid IP address") //TODO wrap actual errror
	        }
                ip[i] = byte(iv)
        }
        return ip
```

pringlewood: @denis-tingajkin thanks for the direction, I implemented this

PR #1776, opened 2019-10-17
Fix Jaeger to work again in helm.

edwarnicke:
## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/k8s/cmd/admission-webhook/const.go
+++ b/k8s/cmd/admission-webhook/const.go
@@ -12,6 +12,7 @@ const (
 	repoEnv              = "REPO"
 	initContainerEnv     = "INITCONTAINER"
 	tagEnv               = "TAG"
+	tracerEnabledEnv     = "TRACER_ENABLED"
 	jaegerHostEnv        = "JAEGER_AGENT_HOST"
 	jaegerPortEnv        = "JAEGER_AGENT_PORT"
 	repoDefault          = "networkservicemesh"
--- a/.mk/helm.mk
+++ b/.mk/helm.mk
@@ -33,6 +33,7 @@ $(INSTALL_CHARTS):
 	--set org="${CONTAINER_REPO}",tag="${CONTAINER_TAG}" \
 	--set forwardingPlane="${FORWARDING_PLANE}" \
 	--set insecure="false" \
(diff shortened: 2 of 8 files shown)


k8s/cmd/admission-webhook/const.go
```diff
@@ -12,6 +12,7 @@ const (
 	repoEnv              = "REPO"
 	initContainerEnv     = "INITCONTAINER"
 	tagEnv               = "TAG"
+	tracerEnabledEnv     = "TRACER_ENABLED"
```
nickolaev: Shall we add these to `docs/env.md`?

deployments/helm/nsm/values.yaml
```diff
@@ -25,4 +25,4 @@ spire:
 
 global:
   # set to true to enable Jaeger tracing for NSM components
-  JaegerTracing: false
+  JaegerTracing: true
```
nickolaev: Hmm I would suggest this to be false by default and for development purposes to enable it in the `make helm-install-nsm` target.

deployments/helm/icmp-responder/templates/nsc.tpl
```diff
@@ -5,7 +5,7 @@ spec:
   selector:
     matchLabels:
       networkservicemesh.io/app: "icmp-responder-nsc"
-  replicas: 4
+  replicas: 1
```
nickolaev: Really depends what we want to show with this simple example. 1 replica might be enough.

k8s/cmd/admission-webhook/const.go
edwarnicke: It is already at the top of docs/env.md ... but that said it defaulted to True.  Fixed that in latest push :)

deployments/helm/nsm/values.yaml
edwarnicke: Good idea. Fixed.

deployments/helm/icmp-responder/templates/nsc.tpl
edwarnicke: This should never have been in the PR, was a tweak I did for local development.

PR #1797, opened 2019-10-20
Enable govet

fkautz:


nickolaev:
@fkautz are we still considering this one?

PR #1798, opened 2019-10-20
Ensure all errors are checked

fkautz:


test/kubetest/rbac/cluster_roles.go
```diff
@@ -224,9 +224,18 @@ func CreateRoleBinding(name, namespace string) Role {
 }
 
 func DeleteAllRoles(clientset kubernetes.Interface) error {
-	(&ClusterRole{}).Delete(clientset, RoleNames["admin"])
-	(&ClusterRole{}).Delete(clientset, RoleNames["view"])
-	(&ClusterRoleBinding{}).Delete(clientset, RoleNames["binding"])
+	if err := (&ClusterRole{}).Delete(clientset, RoleNames["admin"]); err != nil {
+		logrus.Error("Unable to delete kubernetes admin roles, aborting", err)
```
denis-tingajkin: Can we move `logrus.Errorf` from lines 228, 232, 236 to
https://github.com/networkservicemesh/networkservicemesh/blob/master/test/kubetest/k8s.go#L444 ?

test/cloudtest/pkg/commands/main.go
```diff
@@ -720,7 +722,9 @@ func (ctx *executionContext) executeTask(task *testTask, clusterConfigs []string
 				if err != nil {
 					logrus.Errorf("Task failed because cluster is not valid: %v %v %v", task.test.Name, inst.id, err)
 					clusterNotAvailable = true
-					ctx.destroyCluster(inst, true, false)
+					if err := ctx.destroyCluster(inst, true, false); err != nil {
+						logrus.Error("error destroying cluster", err)
```
denis-tingajkin: Missed format. Please use `logrus.Errorf`.

test/kubetest/rbac/cluster_roles.go
fkautz: This is part of the public API, I'd be more comfortable duplicating the error rather than just returning it.

nickolaev:
@fkautz are we still considering this one?

PR #1801, opened 2019-10-22
cloudtest: command line options enhancements

alex-yust:
- global option for cloudtest utility to run only tests that specified, in order of priority:
  -- by name(s) as positional command line parameters
  -- only-run filed of cloudtest config 

- automatic number of clusters instances: the number of instances for each provider is calculated using 'tests-per-cluster-instance' parameter in config and number of tests involving the provider. providers.instances field is now interpreted as maximum allowed number of instances to start.

- 'kind' and 'tags' options to select tests based on provider's kind and test's tags

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

### DIFF
--- a/.cloudtest.yaml
+++ b/.cloudtest.yaml
@@ -24,6 +24,9 @@ retest:  # Allow to do test re-run if some kind of failures are detected, line C
     - "unable to establish connection to VPP (VPP API socket file /run/vpp/api.sock does not exist)"  # a VPP is not started, it will be re-started in general, but will cause test fail.
     # Sometimes (rarely) docker registry is unavailable for a moment
     - "Error response from daemon: Get https://.*docker.io/.*: dial tcp: lookup registry"
+
+tests-per-cluster-instance: 20  # Argument to calculate the number of clusters = <total-tests> / <tests-per-cluster-instance>
+
 reporting:
   junit-report: "results/junit.xml"
 health-check:
--- a/test/cloudtest/pkg/commands/main.go
+++ b/test/cloudtest/pkg/commands/main.go
@@ -155,6 +155,16 @@ func CloudTestRun(cmd *cloudTestCmd) {
 		os.Exit(1)
 	}
 
+	if len(testConfig.OnlyRun) > 0 {
+		logrus.Infof("Imposing top-level 'only-run' tests to all executions: %v", testConfig.OnlyRun)
+		for _, e := range testConfig.Executions {
+			if len(e.OnlyRun) > 0 {
+				logrus.Warningf("Overwriting non-empty 'only-run' on execution '%s'", e.Name)
+			}
+			e.OnlyRun = testConfig.OnlyRun
+		}
+	}
+
 	// Process config imports
 	err = performImport(testConfig)
 	if err != nil {
--- a/test/cloudtest/pkg/commands/main.go
+++ b/test/cloudtest/pkg/commands/main.go
@@ -43,7 +43,6 @@ type Arguments struct {
 	providerConfig  string   // A folder to start scaning for tests inside
 	count           int      // Limit number of tests to be run per every cloud
 	instanceOptions providers.InstanceOptions
-	onlyEnabled     bool // Disable all clusters and enable only enabled in command line.
 }
 
 type clusterState byte
@@ -1191,23 +1190,11 @@ func (ctx *executionContext) createClusters() error {
 	}
 
 	for _, cl := range ctx.cloudTestConfig.Providers {
-		if ctx.arguments.onlyEnabled {
-			logrus.Infof("Disable cluster config:: %v since onlyEnabled is passed...", cl.Name)
-			cl.Enabled = false
-		}
-		for _, cc := range ctx.arguments.clusters {
-			if cl.Name == cc {
-				if !cl.Enabled {
-					logrus.Infof("Enabling config:: %v", cl.Name)
-				}
-				cl.Enabled = true
-			}
-		}
-		if cl.Enabled {
-			logrus.Infof("Initialize provider for config:: %v %v", cl.Name, cl.Kind)
+		if enable, testCount := ctx.shouldEnableCluster(cl); enable {
+			logrus.Infof("Initialize provider for config: %v %v", cl.Name, cl.Kind)
 			provider, ok := clusterProviders[cl.Kind]
 			if !ok {
-				msg := fmt.Sprintf("Cluster provider %s are not found...", cl.Kind)
+				msg := fmt.Sprintf("Cluster provider '%s' not found", cl.Kind)
 				logrus.Errorf(msg)
 				return errors.New(msg)
 			}
@@ -1256,6 +1243,22 @@ func (ctx *executionContext) cleanupClusters(cleanupCtx context.Context) {
 			cl.provider.CleanupClusters(cleanupCtx, cl.config, ctx.manager, ctx.arguments.instanceOptions)
 		}
 	}
+}
+
+func (ctx *executionContext) shouldEnableCluster(cl *config.ClusterProviderConfig) (bool, int) {
+	enabledByCommandLine := utils.Contains(ctx.arguments.clusters, cl.Name)
+	if !cl.Enabled && !enabledByCommandLine {
+		logrus.Infof("Skipping disabled cluster config: %v", cl.Name)
+		return false, 0
+	}
+	cl.Enabled = len(ctx.arguments.clusters) == 0 || enabledByCommandLine
+	if !cl.Enabled {
+		logrus.Infof("Disabling cluster config by cluster filter: %v", cl.Name)
+		return false, 0
+	}
+	testCount := 0
+
+	return cl.Enabled, testCount
 }
 
 func (ctx *executionContext) findTests() error {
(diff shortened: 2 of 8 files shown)


test/cloudtest/pkg/providers/packet/packet_provider.go
```diff
@@ -656,18 +656,18 @@ func (p *packetProvider) ValidateConfig(config *config.ClusterProviderConfig) er
...
 	envValue := os.Getenv("PACKET_AUTH_TOKEN")
 	if envValue == "" {
-		return errors.New("environment variable are not specified PACKET_AUTH_TOKEN")
+		return errors.New("environment variable not set: PACKET_AUTH_TOKEN")
 	}
 
 	envValue = os.Getenv("PACKET_PROJECT_ID")
 	if envValue == "" {
-		return errors.New("environment variable are not specified PACKET_AUTH_TOKEN")
+		return errors.New("environment variable not set: PACKET_AUTH_TOKEN")
```
denis-tingajkin: should be PACKET_PROJECT_ID

alex-yust: Actually, the code may be safely removed, because PACKET_PROJECT_ID and PACKET_AUTH_TOKEN are both listed in env-check: in packet.yaml

denis-tingajkin: *quickly looked on the env_check*.
 Yes, I think we can remove these checks.

denis-tingajkin:
Can we pass a list of `only run tests ` as arguments for cloudtest? Would it be better?
@alex-yust please, share your thoughts

alex-yust:
> Can we pass a list of `only run tests ` as an argument for cloudtest? Would it be better?
> 

Sure it makes sense if we are going to run cloudtest locally(no CI involved) and there is need to do it with varying targets for each run. 
However, since cloudtest is a very slow tool ( >15 minutes for start+stop) I don't see any advantages over 'only-run' field in .cloudtest.yaml.

edwarnicke:
@alex-yust I ofen run cloudtest on kind only... which is quiet a bit quicker (largely because when I've broken something I've broken it independent of environment ;) ).

alex-yust:
@edwarnicke do you mean you need the option to specify test name(s) on command line?
If so, I would suggest the test names to be positional parameters:
cloudtest --cluster vagrant  test1 test2 ...

alex-yust:
- [x] --kind option to filter clusters by kind
- [x] all positional parameters are considered test names to run

alex-yust:
Ok... just discovered there is an execution config with name 'kind' which I confused with field 'kind' in execution config )

alex-yust:
- [x] automatic scaling down number of cluster instances when there are few tests

test/cloudtest/pkg/commands/main.go
```diff
@@ -948,23 +985,42 @@ func (ctx *executionContext) createClusters() error {
 	}
 
 	for _, cl := range ctx.cloudTestConfig.Providers {
-		if ctx.arguments.onlyEnabled {
-			logrus.Infof("Disable cluster config:: %v since onlyEnabled is passed...", cl.Name)
-			cl.Enabled = false
+		enabledByCommandLine := contains(ctx.arguments.clusters, cl.Name)
```
denis-tingajkin: the method looks complexity, can extract private methods?

denis-tingajkin: I wonder why linter doesn't trigger. Probably because the signature of the method is not changed.

test/cloudtest/pkg/config/config.go
```diff
@@ -47,6 +47,7 @@ type ExecutionConfig struct {
 	Env             []string `yaml:"env"`              // Additional environment variables
 	Run             string   `yaml:"run"`              // A script to execute against required cluster
 	OnFail          string   `yaml:"on_fail"`          // A script to execute against required cluster, called if task failed
+	TestsFound      int      `yaml:"-"`                // Number of tests found for the config
```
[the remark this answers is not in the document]
alex-yust: But it's used in another packages:
`Unexported field TestsFound will no longer be accessible from method createClusters  Unexported field TestsFound will no longer be accessible from method findTests`

test/cloudtest/pkg/commands/main.go
```diff
@@ -1064,18 +1126,13 @@ func (ctx *executionContext) findGoTest(executionConfig *config.ExecutionConfig)
...
-		match := true
-		for _, v := range executionConfig.OnlyRun {
-			match = false
-			if v == t.Name {
-				match = true
-				break
-			}
-		}
-		if match {
+		if len(executionConfig.OnlyRun) == 0 || contains(executionConfig.OnlyRun, t.Name) {
```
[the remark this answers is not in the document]
alex-yust: The place is not performance critical, so code simplicity should take priority. 
Also, the length of executionConfig.OnlyRun is expected to be under 10 in most situations (run few failed tests) so using map might not be beneficial at all.

test/cloudtest/pkg/config/config.go
```diff
@@ -57,7 +58,9 @@ type CloudTestConfig struct {
 		JUnitReportFile string `yaml:"junit-report"` // A junit report file location, relative to test root folder.
 	} `yaml:"reporting"` // A reporting options.
 
-	Executions []*ExecutionConfig `yaml:"executions"`
-	Timeout    int64              `yaml:"timeout"` // Global timeout in seconds
-	Imports    []string           `yaml:"import"`  // A set of configurations for import
+	Executions              []*ExecutionConfig `yaml:"executions"`
+	Timeout                 int64              `yaml:"timeout"`                    // Global timeout in seconds
+	Imports                 []string           `yaml:"import"`                     // A set of configurations for import
+	OnlyRun                 []string           `yaml:"only-run"`                   // If non-empty, only run the listed tests
```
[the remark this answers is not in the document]
alex-yust: Current implementation will override 'only-run' field value if tests are specified on command line.

test/cloudtest/pkg/commands/main.go
alex-yust: > method createClusters looks complexity, can extract private methods?

do you mean to extract enable/disable logic?

denis-tingajkin: I meant, we need to refactor this method and expose new methods

test/cloudtest/pkg/config/config.go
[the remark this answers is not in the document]
denis-tingajkin: Do we really need this field when we can pass `OnlyRun` tests from args?

denis-tingajkin:
Also, you can check new functionality by unit testing. We already have examples: https://github.com/networkservicemesh/networkservicemesh/blob/master/test/cloudtest/pkg/tests/all_cluster_fail_test.go#L16

test/cloudtest/pkg/config/config.go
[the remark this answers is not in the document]
alex-yust: * The field is used to pass test names from command line to testing framework
* The field affects all kinds of runs: local, CircleCI and Azure Pipelines, which is good

test/cloudtest/pkg/commands/main.go
alex-yust: I've extracted enable/disable cluster logic to a separate function. Otherwise I'd prefer not to make changes to the code not related to  the PR subject.

test/cloudtest/pkg/providers/packet/packet_provider.go
```diff
@@ -656,19 +656,9 @@ func (p *packetProvider) ValidateConfig(config *config.ClusterProviderConfig) er
 	for _, envVar := range config.EnvCheck {
 		envValue := os.Getenv(envVar)
 		if envValue == "" {
-			return errors.Errorf("environment variable are not specified %s Required variables: %v", envValue, config.EnvCheck)
+			return errors.Errorf("environment variable not set: %s. Required variables: %v", envVar, config.EnvCheck)
 		}
 	}
 
-	envValue := os.Getenv("PACKET_AUTH_TOKEN")
```
haiodo: We need to have this options to be checked, since user could miss to specify them in config.

alex-yust: [x] hard-coded check for "PACKET_AUTH_TOKEN", "PACKET_PROJECT_ID"

.cloudtest.yaml
```diff
@@ -10,6 +10,8 @@ import:
   - .cloudtest/aws.yaml
   - .cloudtest/azure.yaml
 
+tests-per-cluster-instance: 20
```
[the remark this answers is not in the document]
alex-yust: It's a parameter to calculate the number of instances needed, not a limit.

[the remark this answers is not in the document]
haiodo: May be in this case name it "cluster-min-test-count"?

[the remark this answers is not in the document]
alex-yust: > name it "cluster-min-test-count"?

I don't understand how this name is related to calculating number of cluster instances. Currently, the formula is like:

cluster-instance-number = tests-found-for-cluster / tests-per-cluster-instance

What would be more descriptive name for "tests-per-cluster-instance" in this expression?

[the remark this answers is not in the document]
haiodo: I fine with name, @edwarnicke I suppose we could resolve this discussion.

PR #1839, opened 2019-11-06
OPA initial implementation

lobkovilya:
## Motivation and Context

Provides initial support of OPA to SDK

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


sdk/endpoint/auth.go
```diff
@@ -0,0 +1,55 @@
+package endpoint
```
edwarnicke: Rather than calling this 'auth' which is ambigous between Authentication (authn) and authorization (authz) please name it either authorization or authz.

sdk/endpoint/auth.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/networkservicemesh/networkservicemesh/controlplane/api/connection"
+	"github.com/networkservicemesh/networkservicemesh/controlplane/api/networkservice"
+)
+
+type AuthEndpoint struct {
+	policy rego.PreparedEvalQuery
+}
+
+func (auth *AuthEndpoint) Request(ctx context.Context, request *networkservice.NetworkServiceRequest) (*connection.Connection, error) {
+	rs, err := auth.policy.Eval(ctx, rego.EvalInput(request))
```
edwarnicke: I see us inputing the request... do we want to input the authentication token as well from the context?

sdk/endpoint/auth.go
```diff
@@ -0,0 +1,55 @@
...
+
+	for _, r := range rs {
+		for _, e := range r.Expressions {
+			t, ok := e.Value.(bool)
+			if !ok {
+				return nil, status.Error(codes.Internal, "non boolean expression")
+			}
+
+			if !t {
+				return nil, status.Error(codes.PermissionDenied, "no sufficient privileges to call Request")
```
edwarnicke: Is it possible to get any richer data about why permission Denied to help the user?

sdk/endpoint/auth.go
lobkovilya: Yes sure, but we have to have JWT tokens merged in order to do it. And then we will be able to pass the whole chain of SPIFFE IDs

sdk/endpoint/auth.go
lobkovilya: Sure, thank you! Will fix

sdk/endpoint/auth.go
lobkovilya: Yes but I think it depends on provided policy? Now `AuthzEndpoint` is quite generic. But I guess I can add some information about Spiffe Id

lobkovilya:
Implemented in https://github.com/networkservicemesh/sdk/pull/40

PR #1858, opened 2019-11-09
Improve DNS documentation

john-a-joyce:
This commit improves the DNS integration
documentation focusing on how the user can enable
the capability.

Signed-off-by: john-a-joyce <joycej@cisco.com>

## Description
This is a documentation only change.   It provides more detail on all the components in the 
DNS integration path.  It adds more information on the role of the NSE in DNS integration
and includes the key user config options of setting NSE environmental variables to trigger
updates of the DNS configuration in the application pod.  

## Motivation and Context
The key user steps to set environmental variables was missing. 

## How Has This Been Tested?
Documentation only. 
- [ ] Tested locally
- [x] Have not tested

## Types of changes
Documentation. 

## Checklist:

### DIFF
--- a/docs/spec/dns-integration.md
+++ b/docs/spec/dns-integration.md
@@ -11,9 +11,7 @@ Overview
 
 The DNS integration capability is performed by running a DNS server co-resident with the application pod that can direct requests to multiple DNS servers. The DNS servers that should be sent requests is controlled by the NSEs that create connections into that application pod.  If the NSEs populate the DNSContext with a DNS server appropriate for this connection then additional containers in the application pod will ensure those DNS servers are consulted. The additional containers are nsm-coredns and nsm-dns-monitor.  They can be inserted directly or via the admission webhook.  The nsm-coredns container is responsible for handling the DNS resolution mechanics. The nsm-dns-monitor container is responsible for updating the configuration that the nsm-coredns container acts on based on DNS information in the connection contexts provided by the NSEs.
 
-The NSM provided SDK provides a capability to read environmental varibles to determine additional DNS servers and domains that should be added to the DNScontext.  These variables are: DNS_SEARCH_DOMAINS and DNS_SERVER_IPS.
-
-Note:  Some of the additional DNS configuration capabilites K8s offers are overridden adn not availabel when these additional containers are inserted.
+The NSM endpoint SDK provides two functions `NewAddDNSConfigs` and  `NewAddDnsConfigDstIp` to update the connection context so the nsm-dns-monitor can update the coredns configuration.
 
 Implementation details (optional)
 ---------------------------------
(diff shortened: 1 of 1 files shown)


docs/spec/dns-integration.md
```diff
@@ -4,23 +4,32 @@ DNS Integration for NSM
...
+
+Overview
+--------
+
+The DNS integration capability is performed by running a DNS server co-resident with the application pod that can direct requests to multiple DNS servers. The DNS servers that should be sent requests is controlled by the NSEs that create connections into that application pod.  If the NSEs populate the DNSContext with a DNS server appropriate for this connection then additional containers in the application pod will ensure those DNS servers are consulted. The additional containers are nsm-coredns and nsm-dns-monitor.  They can be inserted directly or via the admission webhook.  The nsm-coredns container is responsible for handling the DNS resolution mechanics. The nsm-dns-monitor container is responsible for updating the configuration that the nsm-coredns container acts on based on DNS information in the connection contexts provided by the NSEs.
+
+The NSM provided SDK provides a capability to read environmental varibles to determine additional DNS servers and domains that should be added to the DNScontext.  These variables are: DNS_SEARCH_DOMAINS and DNS_SERVER_IPS.
```
edwarnicke: I don't see these env variables used in the sdk/ .  I do see them used in some test/applications/... but not the SDK.

The parts of the SDK I find with DNS related bits are:

[NewAddDnsConfigs](https://github.com/networkservicemesh/networkservicemesh/blob/master/sdk/endpoint/dns.go#L33)

and

[NewAddDnsConfigDstIP](https://github.com/networkservicemesh/networkservicemesh/blob/master/sdk/endpoint/dns.go#L59)

Did I miss something?

docs/spec/dns-integration.md
```diff
@@ -4,23 +4,32 @@ DNS Integration for NSM
...
+
+Note:  Some of the additional DNS configuration capabilites K8s offers are overridden adn not availabel when these additional containers are inserted.
```
edwarnicke: Which ones?  We tried hard to make sure all normal K8s DNS things worked... what did we miss?

docs/spec/dns-integration.md
```diff
@@ -81,11 +90,19 @@ func main() {
...
+See an example of usage `nsm-dns-monitor` in `test/applications/cmd/monitoring-dns-nsc`
+
+### Using nsm-coredns and nsm-dns-monitor without changing the client's deployment configuration
+To inject the `nsm-coredns` and `nsm-dns-monitor` containers into a client's pod during deployment, you can simply deploy the [admission webhook](https://github.com/networkservicemesh/networkservicemesh/blob/master/docs/spec/admission.md). `Admission webhook` will automatically append the DNS specific containers to your `Network Service Client`.  When using the admission webhook there is no way to disable the insertion of these additional containers.
+
+## NSE Requirements
+In order for the application pod to try multiple DNS servers the NSEs must populate the DNScontext.   The SDK provides a means to populate the DNScontext based on environmental variables provided to the NSE container. These variables are: DNS_SEARCH_DOMAINS and DNS_SERVER_IPS.  The [icmp-responer](test/applications/cmd/icmp-responder-nse/main.go) has an example implementation.
+
```
edwarnicke: The icmp-responder is a great example of using the SDK... but the SDK out of the box isn't using those variables (though maybe we should consider that :) ).

docs/spec/dns-integration.md
john-a-joyce: You didn't miss anything - I explained this more in terms what the test applications does vs. what the SDK does.   I will need to fix that up.

docs/spec/dns-integration.md
john-a-joyce: Yeah - As I responded above I need to be clear on what the SDK does vs. what the icmp-responder example does.

docs/spec/dns-integration.md
john-a-joyce: I was a bit speculative here but some of the options described here:  https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-config almost certainly don't work.   I have not directly tested this so willing to reword this if its an issue,   but I think we should provide some warning to not mix the advanced pod DNS config with the fanout plugin NSM is implementing

edwarnicke: @denis-tingajkin How are we handling these cases @john-a-joyce referenced here?

denis-tingajkin: @edwarnicke At this moment we have implemented `resolv.conf` parser to reading user's nameservers and search domains. But we did not a solution for some not default [options](http://man7.org/linux/man-pages/man5/resolv.conf.5.html).

denis-tingajkin: I've checked. Options also supported. We just keep user's options from POD DNS config.

denis-tingajkin: I think, In general, we can use any K8S DNS configuration capabilities.

john-a-joyce: I would be surprised if all the capabilities that K8s offers work as the user would expect when the fanout is used, but I also have not fully tested or evaluated. So it is best that I just remove this warning entirely at the current time.

john-a-joyce:
I think I address all the review comments.   Let me know if you see any concerns.

nickolaev:
@edwarnicke NSE vse Endpoint and NSC vs Client?

edwarnicke:
> @edwarnicke NSE vse Endpoint and NSC vs Client?

My thinking is roughly this: Is there a possibility of confusing an NSC for a WebClient if we use Client in this documentation?

john-a-joyce:
I am not sure on the conclusion should I change some names or not?

john-a-joyce:
Is there anything more to change here or can this Merge?

PR #1870, opened 2019-11-13
Rename Helm template for forwarding plane

rdimitrov:
Needed as the Helm forwarding plane template was named after the VPP
forwarding plane.

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


edwarnicke:
@rdimitrov Question... how would you forsee this evolving forward when we get to multiple forwarding planes?

nickolaev:
IMO, each forwarding plane has to live in its own repo, providing the helmcharts for its deployment. That will be even more true once we get to "cross-connect-as-a service" (CCaaS?) model. Effectively the NSM deplyment will literally be - NSMgr as a Daemon Set.

edwarnicke:
> IMO, each forwarding plane has to live in its own repo, providing the helmcharts for its deployment. That will be even more true once we get to "cross-connect-as-a service" (CCaaS?) model. Effectively the NSM deplyment will literally be - NSMgr as a Daemon Set.

Understood.  I expect an intermediate stage though where we are migrating to the "cross-connect-as -a-service" model and exercising multiple forwarders out of the networkservicemesh/ repo... so I guess I should recast my question for that medium term.

Also... even with the forwarders in their own repos (a good idea) I expect we will want to publish nsm artifacts in a common helm repo... so will want non-conflicting chart names there.

rdimitrov:
So far we're using this general helm template for both VPP and the kernel-forwarder and I guess it was left behind at some point during the refactorings, so that's the reason for renaming it. 

About the question, until there's no difference between each forwarding plane, I guess we can loop over the list of forwarders we want to deploy using again the same helm template. 

But in case there're differences in their helm charts, I haven't thought of anything neat so far.

edwarnicke:
@rdimitrov Ah... I'd missed [that it was handling both forwarding planes with a variable to switch on](https://github.com/networkservicemesh/networkservicemesh/blob/master/deployments/helm/nsm/templates/vppagent-forwarder.tpl#L1).  Given that this makes a lot of sense to me :)

Lets see if we can get it past CI.  Also, could we get it on release-0.2. as well?

rdimitrov:
Great 👍 I'll prepare the same PR for the release-0.2 branch. 

ps. Do you want me to force push or we can re-run the CI?

PR #1904, opened 2019-11-18
Cleanup cloudtest messages

alex-yust:
* Remove multiple chaotic calls to printStistics() during clusters startup
* Remove statistics from 'test complete' message

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/test/cloudtest/pkg/commands/main.go
+++ b/test/cloudtest/pkg/commands/main.go
@@ -345,7 +345,7 @@ func (ctx *executionContext) assignTasks() {
 func (ctx *executionContext) performClusterUpdate(event operationEvent) {
 	ctx.Lock()
 	defer ctx.Unlock()
-	logrus.Infof("Instance for cluster %s is updated %v", event.clusterInstance.group.config.Name, event.clusterInstance)
+	logrus.Infof("Cluster instance %s is updated: %v", event.clusterInstance.id, event.clusterInstance)
 	if event.clusterInstance.taskCancel != nil && event.clusterInstance.state == clusterCrashed {
 		// We have task running on cluster
 		event.clusterInstance.taskCancel()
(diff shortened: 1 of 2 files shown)


test/cloudtest/pkg/commands/main.go
```diff
@@ -281,23 +280,21 @@ func (ctx *executionContext) performExecution() error {
...
-		case <-timectx.Done():
-			ctx.printStatistics()
-			timectx, cancelfunc = context.WithTimeout(context.Background(), 1*time.Minute)
-			defer cancelfunc()
 		case <-termChannel:
 			return errors.New("termination request is received")
 		case <-timeoutCtx.Done():
 			ctx.printStatistics()
 			return errors.Errorf("global timeout elapsed: %v seconds", ctx.cloudTestConfig.Timeout)
+		case <-time.After(60 * time.Second):
```
haiodo: This is not right, @alex-yust you do same mistake as I, since updates happen more often it will lead to update very-very often, and will no print messages, since will exit with event := <-ctx.operationChannel and not with print timeout.

alex-yust: Yeah, I understood that a bit later too ;)

alex-yust: - [ ] create dedicated timer for calling printStatistics

test/cloudtest/pkg/commands/main.go
```diff
@@ -268,8 +269,10 @@ func (ctx *executionContext) performExecution() error {
 	defer cancelFunc()
 
 	termChannel := tools.NewOSSignalChannel()
-	timectx, cancelfunc := context.WithTimeout(context.Background(), 30*time.Second)
-	defer cancelfunc()
+	defer ctx.printStatistics()
+
+	statTicker := time.NewTicker(60 * time.Second)
```
haiodo: Nice  👍

Issue #1898, opened 2019-11-17
Cover the case when MonitorNetNsInodeServer handled  broken connections


denis-tingajkin:
## Expected Behavior

MonitorNetNsInodeServer [should handle broken connections](https://github.com/networkservicemesh/networkservicemesh/blob/0cd820b807f76d1e152f6384a80054967ddc7d44/controlplane/pkg/nsmd/nsmd.go#L412)
 
## Current Behavior

Not covered by testing.

## Steps to Reproduce

1. Deploy nsmgr / forwarder
2. Deploy  dirty NSE
3. Deploy NSC
4. Verify connection
5. Delete NSE // NSE POD should be deleted first
6. Delete NSC
7. Make sure that broken connection handled and NSE custom resource deleted.

## Context

https://github.com/networkservicemesh/networkservicemesh/blob/0cd820b807f76d1e152f6384a80054967ddc7d44/controlplane/pkg/nsmd/nsmd.go#L412

denis-tingajkin:
test case updated

lobkovilya:
@a-bastrykina after the discussion with @haiodo:

1. `TestDeleteDirtyNSEWithClient` isn't valid: even having a client with connection, CRD for NSE should be eventually cleaned. So that test should be rewrote properly.
    Cleanup of CRD should be placed somewhere inside `healDstDown` in case of `waitNSE` returned false.

3. Need some research, what's going to happen in the following case:
    1. Start cross-connect monitor
    2. Delete NSE
    3. Wait for healing started (using logs)
    4. Delete NSC

    We expect to receive `Delete` event from cross-connect monitor, but there is a probability that it doesn't work correctly.

denis-tingajkin:
looks like solved

PR #1916, opened 2019-11-20
Broken connections handling fix

a-bastrykina:
## Description
Removed code in method `DeleteEndpointWithBrokenConnection`, which skips deleting endpoint when it has active client connection. 
It was expected that `MonitorNetNsInodeServer` handles that case, but this logic wasn't implemented. 
So when endpoint has no handlers to cleanup itself from the registry once it's killed, it will never be deleted from the registry. 
Also we had  `TestDeleteDirtyNSEWithClient` to test case, but it was invalid, so I fixed it too.  

## Motivation and Context
#1898 

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/controlplane/pkg/nsm/nsm_heal_processor.go
+++ b/controlplane/pkg/nsm/nsm_heal_processor.go
@@ -356,12 +356,6 @@ func (p *healProcessor) nseIsNewAndAvailable(ctx context.Context, endpointName s
 		return false
 	}
 
-	// Check local only if not waiting for specific NSE.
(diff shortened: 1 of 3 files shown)


nickolaev:
Can you please have a short description of what is happening here. There  are lots of details in the referenced issue, but having a simple summary will help. I see we have deleted code but why is this?

a-bastrykina:
@nickolaev Updated PR description

PR #1937, opened 2019-11-28
Use eval in for-each-module.sh

nickolaev:
## Description

Use the POSIX `eval` when invoking the command as specified
as an argument to ./scripts/for-each-module.sh. This ensures
better command handling and the ability to use the module name
in the command. 

Example:
```bash
./scripts/for-each-module.sh "eval 'echo  \${d#\.\/}'"
```

## Motivation and Context
To improve the capabilities of our modules tooling.

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/scripts/for-each-module.sh
+++ b/scripts/for-each-module.sh
@@ -1,10 +1,14 @@
 #!/bin/bash
 
-CMD=$*
+# A script to run through all the modules and execute the supplied command(s)
+# The name of the module can be accessed in the command sith _MODULE_ variable.
+# Example: for-each-module.sh echo \$_MODULE_
+
+CMD="$*"
 res=0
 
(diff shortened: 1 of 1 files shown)


alex-yust:
>ability to use the module name in the command.

Nice idea! How about defining a variable with module name to make things even prettier:
( d=$(dirname "$f") && cd "$d" && _MODULE_=$(basename "$d") && eval "$CMD"; ) || res=1

and now we can do:
./scripts/for-each-module.sh echo \$_MODULE_

nickolaev:
`basename` will return `api` for the `forwarder/api` module. Will stick to shell script string manipulation. Good idea, though.

PR #1938, opened 2019-11-28
Kind uses context now, drop KUBECONFIG

nickolaev:
## Description
Updated Kind is now using kubectl context to point to the right cluster. Drop all the oldschool `kubeconfig-path` references in the make machinery and in the docs.

## Motivation and Context
The old way of referring to the cluster is obsoleted and will be deprecated at some point. It is automatically set now, so no need to do anything for a single deployed kind cluster.

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/.cloudtest/kind.yaml
+++ b/.cloudtest/kind.yaml
@@ -16,7 +16,6 @@ providers:
         make kind-config
         make kind-start
       stop: make kind-stop
-      config: make kind-config-location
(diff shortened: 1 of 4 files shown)


fkautz:
Nice, I really like this! I have been wondering why kind did not support using contexts.

Issue #1950, opened 2019-12-02
Helm make targets not compatible with Helm 3


rktidwell:
## Expected Behavior

Helm make targets should succeed when running with helm 3 (as soon as it is officially released).

## Current Behavior

My distro (openSUSE Tumbleweed) is providing helm 3 packages. When running the NSM examples that require invoking helm make targets, they fail due to significant changes to the helm CLI.

## Steps to Reproduce

1. Install helm 3
2. Begin following the steps set out in https://github.com/networkservicemesh/examples/blob/master/README.md#run-the-examples .
3. On the second step you will invoke `'SPIRE_ENABLED=false INSECURE=true make helm-init helm-install-nsm'` . This fails with the following output:

```
make[1]: Entering directory '/home/ryan/workspace/go/src/networkservicemesh'
helm init --wait && ./scripts/helm-patch-tiller.sh
Error: unknown flag: --wait
make[1]: *** [.mk/helm.mk:23: helm-init] Error 1
make[1]: Leaving directory '/home/ryan/workspace/go/src/networkservicemesh'
make: *** [Makefile:53: helm-init] Error 2
```

4. Because `'helm init'` has been removed in v3 and won't be needed, I thought I would just work around this by invoking the  `'helm-install-nsm'` target. That fails too with the following output:

```
make[1]: Entering directory '/home/ryan/workspace/go/src/networkservicemesh'
# We specifically set admission-webhook variables here as it is a subchart
# there might be a way to set these as global and refer to them with .Values.global.org
# but that seems more intrusive than this hack. Consider changing to global if the charts
# get even more complicated
helm install --name=nsm \
--atomic --timeout 300 \
--set org="networkservicemesh",tag="master" \
--set forwardingPlane="vpp" \
--set insecure="true" \
--set global.JaegerTracing="true" \
--set spire.enabled="false",spire.org="networkservicemesh",spire.tag="master" \
--set admission-webhook.org="networkservicemesh",admission-webhook.tag="master" \
--set prefix-service.org="networkservicemesh",prefix-service.tag="master" \
--namespace="nsm-system" \
deployments/helm/nsm
Error: unknown flag: --name
make[1]: *** [.mk/helm.mk:28: helm-install-nsm] Error 1
make[1]: Leaving directory '/home/ryan/workspace/go/src/networkservicemesh'
make: *** [Makefile:53: helm-install-nsm] Error 2
```

[...]

PR #1956, opened 2019-12-04
Add support for Helm 3 in make targets

rktidwell:
This change adds support for Helm 3 while ensuring backward compatibility
with Helm 2. To accomplish this, affected make targets now invoke
wrapper scripts that determine the version of Helm available and invoke
helm accordingly.

Signed-off-by: Ryan Tidwell <rtidwell@suse.com>

## Description

* Introduce wrapper scripts for the 'helm init' step and the NSM install steps. These scripts check the version of the helm client available and proceed accordingly.
* Change .mk/helm.mk to invoke wrapper scripts so that makefiles are insulated from the differences in helm versions.

## Motivation and Context

With the recent availability of Helm 3 and support for Helm 2 disappearing in about a year, a little housekeeping to ensure compatibility with Helm 3 is in order. See #1950 

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/scripts/helm-nsm-install.sh
+++ b/scripts/helm-nsm-install.sh
@@ -1,0 +1,204 @@
+#!/bin/bash
+
+# A script for wrapping NSM installation via Helm. Both Helm 2 and Helm 3 are
+# supported.
+
+
+function with_helm_2 () {
+  # We specifically set admission-webhook variables here as it is a subchart
+  # there might be a way to set these as global and refer to them with .Values.global.org
+  # but that seems more intrusive than this hack. Consider changing to global if the charts
+  # get even more complicated
+  echo -n "Performing chart installation..."
+  helm install --name="$CHART" \
+  --atomic --timeout 300 \
+  --set org="$CONTAINER_REPO",tag="$CONTAINER_TAG" \
+  --set forwardingPlane="$FORWARDING_PLANE" \
+  --set insecure="$INSECURE" \
+  --set prometheus="${PROMETHEUS}" \
+  --set metricsCollectorEnabled="${METRICS_COLLECTOR_ENABLED}" \
+  --set global.JaegerTracing="true" \
+  --set spire.enabled="$SPIRE_ENABLED",spire.org="$CONTAINER_REPO",spire.tag="$CONTAINER_TAG" \
+  --set admission-webhook.org="$CONTAINER_REPO",admission-webhook.tag="$CONTAINER_TAG" \
+  --set prefix-service.org="$CONTAINER_REPO",prefix-service.tag="$CONTAINER_TAG" \
+  --namespace="$NSM_NAMESPACE" \
+  deployments/helm/"$CHART" &
+  PID=$!
+  sleep 2
+  while ps -p ${PID} > /dev/null;
+  do
+    printf "."
+    sleep 2
+  done
+  echo "Done"
+}
+
+function with_helm_3 () {
+  # Ensure we have the nsm namespace available
+  echo -n "Ensure namespace $NSM_NAMESPACE exists..."
+  kubectl apply -f k8s/conf/namespace-nsm.yaml &
+  PID=$!
+  sleep 2
+  while ps -p ${PID} > /dev/null;
+  do
+    printf "."
+    sleep 2
+  done
+  echo "Done"
+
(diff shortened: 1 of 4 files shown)


scripts/helm-nsm-install.sh
```diff
@@ -0,0 +1,142 @@
...
+  --set admission-webhook.org="$CONTAINER_REPO",admission-webhook.tag="$CONTAINER_TAG" \
+  --set prefix-service.org="$CONTAINER_REPO",prefix-service.tag="$CONTAINER_TAG" \
+  --namespace="$NSM_NAMESPACE" \
+  deployments/helm/"$CHART"
+
+}
+
+function with_helm_3 () {
+  # Ensure the nsm namespace has been created
+  kubectl config use-context kind-nsm
```
nickolaev: Please, do not refer to kind in this script. Using kind is one way to deploy NSM, but for development some may use Vagrant or others. Keep the script more generic.

rktidwell: Ack, we can probably do something different here.

rktidwell:
> Btw are there any changes to the helm charts needed?

I haven't stumbled across anything that needs attention yet, but I've only tried out a couple charts. There's more work to do on that front, and anything we might find could be addressed with other PR's.

nickolaev:
Is this one still a WiP? Seems like there is a bug that was fixed only i 3++ versions:

https://github.com/helm/helm/issues/6361

rktidwell:
> Is this one still a WiP? Seems like there is a bug that was fixed only i 3++ versions:
> 
> [helm/helm#6361](https://github.com/helm/helm/issues/6361)

I suppose we can take the WIP tag off this PR. I am just trying to understand the CI errors and whether this change is causing the failure or not. Assuming we use a patched version of Helm there shouldn't be a problem RE https://github.com/helm/helm/issues/6361. This  change is simply a wrapper around invocation of the client, so I think this issue is a bit orthogonal to what I'm attempting to address with this PR.

scripts/helm-nsm-install.sh
```diff
@@ -0,0 +1,170 @@
...
+    ;;
+    --container_tag)
+    CONTAINER_TAG="$2"
+    shift
+    shift
+    ;;
+    --insecure)
+    INSECURE="$2"
+    shift
+    shift
```
lobkovilya: If `INSECURE` env is not set, then installation scripts hangs forever because of the second `shift`. Please add `INSECURE?=false` to Makefiles (same way as for `CONTAINER_REPO` or `CONTAINER_TAG`)

scripts/helm-nsm-install.sh
```diff
@@ -0,0 +1,170 @@
...
+  --set forwardingPlane="$FORWARDING_PLANE" \
+  --set insecure="$INSECURE" \
+  --set global.JaegerTracing="true" \
+  --set spire.enabled="$SPIRE_ENABLED",spire.org="$CONTAINER_REPO",spire.tag="$CONTAINER_TAG" \
+  --set admission-webhook.org="$CONTAINER_REPO",admission-webhook.tag="$CONTAINER_TAG" \
+  --set prefix-service.org="$CONTAINER_REPO",prefix-service.tag="$CONTAINER_TAG" \
+  --namespace "$NSM_NAMESPACE" \
+  deployments/helm/"$CHART" &
+  sleep 2
+  while [ -d /proc/$PID ]
```
lobkovilya: That doesn't look like MacOS compatible solution. I'm not sure but I think it is important to us.  Could we consider some solution based on `ps`?

lobkovilya:
`make helm-delete-*` seems to be broken for Helm 3, since `--purge` flag became default for `helm delete`

scripts/helm-nsm-install.sh
rktidwell: Thanks for pointing this out, I'll fix that.

scripts/helm-nsm-install.sh
rktidwell: I'll look into it, I don't work on a mac so that didn't occur to me. It's a nice touch for folks who develop on macs.

nickolaev:
Thanks @rktidwell !

PR #1958, opened 2019-12-04
Cleanup old clusters with cloudtest run

artem-belov:
## Description
Desctroy clusters older then 4 hours when cloudtest run

## Motivation and Context
Should help to avoid leaking clusters

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/scripts/aws/destroy-kubernetes-cluster.go
+++ b/scripts/aws/destroy-kubernetes-cluster.go
@@ -21,7 +21,7 @@ import (
 //nolint:gochecknoglobals
 var deferError error
 
-func checkDeferError(err error) bool {
+func (ac *AWSCluster) checkDeferError(err error) bool {
 	if err != nil {
 		if aerr, ok := err.(awserr.Error); ok {
 			if aerr.Code() == "Throttling" {
@@ -49,7 +49,7 @@ func checkDeferError(err error) bool {
 	return false
 }
 
-func DeleteEksRole(iamClient *iam.IAM, eksRoleName *string) {
+func (ac *AWSCluster) deleteEksRole(iamClient *iam.IAM, eksRoleName *string) {
 	log.Printf("Deleting EKS service role \"%s\"...\n", *eksRoleName)
 
 	policyArn := "arn:aws:iam::aws:policy/AmazonEKSClusterPolicy"
@@ -57,30 +57,30 @@ func DeleteEksRole(iamClient *iam.IAM, eksRoleName *string) {
 		RoleName:  eksRoleName,
 		PolicyArn: &policyArn,
 	})
-	checkDeferError(err)
+	ac.checkDeferError(err)
 
 	policyArn = "arn:aws:iam::aws:policy/AmazonEKSServicePolicy"
 	_, err = iamClient.DetachRolePolicy(&iam.DetachRolePolicyInput{
 		RoleName:  eksRoleName,
 		PolicyArn: &policyArn,
 	})
-	checkDeferError(err)
+	ac.checkDeferError(err)
 
 	_, err = iamClient.DeleteRole(&iam.DeleteRoleInput{
 		RoleName: eksRoleName,
 	})
-	if checkDeferError(err) {
+	if ac.checkDeferError(err) {
 		return
 	}
 
 	log.Printf("Role \"%s\" successfully deleted!\n", *eksRoleName)
 }
 
-func DeleteEC2NetworkInterfaces(ec2Client *ec2.EC2, cfClient *cloudformation.CloudFormation, nodesStackName *string) {
+func (ac *AWSCluster) deleteEC2NetworkInterfaces(ec2Client *ec2.EC2, cfClient *cloudformation.CloudFormation, nodesStackName *string) {
 	cfResp, err := cfClient.DescribeStacks(&cloudformation.DescribeStacksInput{
 		StackName: nodesStackName,
 	})
-	if checkDeferError(err) {
+	if ac.checkDeferError(err) {
 		return
 	}
 
@@ -101,7 +101,7 @@ func DeleteEC2NetworkInterfaces(ec2Client *ec2.EC2, cfClient *cloudformation.Clo
 		},
 	})
 
-	if checkDeferError(err) {
+	if ac.checkDeferError(err) {
 		return
 	}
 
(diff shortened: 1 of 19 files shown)


scripts/aws/destroy-kubernetes-cluster.go
```diff
@@ -300,17 +300,27 @@ func deleteAllKubernetesClusters(saveDuration time.Duration) {
 			continue
 		}
 
+		if matched, err := regexp.MatchString(namePattern, stackName[7:]); err != nil || !matched {
```
haiodo: I'm not sure: stackName[7:] is good :-)

scripts/aws/destroy-kubernetes-cluster.go
```diff
@@ -292,25 +317,35 @@ func deleteAllKubernetesClusters(saveDuration time.Duration) {
...
 
 		wg.Add(1)
 
 		go func() {
 			defer wg.Done()
-			logrus.Infof("Deleting %s", stackName[7:])
-			deleteAWSKubernetesCluster(stackName[7:])
+			for att := 0; att < 3; att++ {
+				logrus.Infof("Deleting %s (created %v), attempt %d", stackName[len(awsClusterStackPrefix):], stack.CreationTime, att+1)
+				err := NewAWSCluster(stackName[7:]).DeleteAWSKubernetesCluster()
```
haiodo: Please use prefix len.

scripts/packet/destroy-old-clusters.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+
+pushd "$(dirname "${BASH_SOURCE[0]}")"/../../test/cloudtest/pkg/providers/packet/packet_cleanup || exit 1
+go run ./... -t "$PACKET_AUTH_TOKEN" -p "$PACKET_PROJECT_ID" -k y -c y
```
haiodo: Let's modify packet_cleanup tool to use ENV variable instead of passing value via command line. Thanks,

scripts/gke/destroy-old-clusters.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+usage() { echo "Usage: $0 [-t <hours>] [-p <string>]" 1>&2; exit 1; }
```
alex-yust: It would be nice to have both options explained a bit more

scripts/gke/destroy-old-clusters.sh
```diff
@@ -0,0 +1,39 @@
...
+numreg='^[0-9]+$'
+
+while getopts ":t:p:" o; do
+    case "${o}" in
+        p)
+            p=${OPTARG}
```
alex-yust: p identifier is Ok when its scope just a few lines of code, otherwise a more descriptive name is needed for readability sake.

scripts/aws/destroy-old-clusters.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
```
nickolaev: Wouldn't this be much simpler if written in go? What makes `/bin/bash` the right choice here?

scripts/azure/destroy-old-clusters.sh
```diff
@@ -0,0 +1,62 @@
+#!/bin/bash
```
nickolaev: Why not in go?

scripts/gke/destroy-old-clusters.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
```
nickolaev: go?

scripts/aws/destroy-old-clusters.sh
artem-belov: cloudtest use bash commands to control aws cloud for testing. This script just a Wrapper of golang programm to use it with cloudtest

scripts/gke/destroy-old-clusters.sh
artem-belov: cloudtest use bash commands to create,destroy and control gke cloud. So bash scripts is the simpliest way to find and destroy old clusters

scripts/azure/destroy-old-clusters.sh
artem-belov: cloudtest use bash commands to create,destroy and control azure cloud. So bash scripts is the simpliest way to find and destroy old clusters

nickolaev: These scripts will create more mainatners burden than htey deserve, please consider using `go`
I don't think that code like this line is something we should accept to maintain in 2020 and onwards:
```
 IFS=$'\t'; read -r -a cols <<< "${raws[$i-1]}"; unset IFS;
```

lobkovilya: We already have a bunch of scripts in bash, I don't think makes sense to rewrite all of them into go. 
Maybe we can rewrite that particular line in a more readable way? @nickolaev will it be ok to you?

scripts/azure/destroy-old-clusters.sh
```diff
@@ -0,0 +1,62 @@
...
+if [ -z "${time_passed}" ] || [ -z "${pattern}" ]; then
+    usage
+fi
+
+CLUSTERS=$(az aks list -g "nsm-ci" --query "[].{Name:name,Id:id,Group:resourceGroup}" -o tsv | grep "${pattern}")
+IFS=$'\n'; 
+# shellcheck disable=SC2206
+raws=($CLUSTERS); 
+unset IFS;
+for ((i=1;i<=${#raws[@]};i++)); do
```
lobkovilya: @artem-belov why not simply 
```bash
for cluster in $CLUSTERS; do
...
```

scripts/azure/destroy-old-clusters.sh
nickolaev: Having simple scripts may be OK, but these are too complex jsut to parse some arguments, You can parse the arguments in the go code itself. Why need these shell script magics.
A simple wrapper is OK, putting advanced logic is a potential source of problems in the future. All these clouds will change the way they operate in 6 months probably and then we (or whoever works on the project) will ave to figure out again how this code works. Go is a cleaner solution for all the advanced processing - arguments, algorithms, retries etc. Simple wrappers in shell are OK - similar to what #1973 does.

lobkovilya: That PR has quite simple logic, but it's written in complicated way. I believe we can review and simplify it, @alex-yust has thoughts about simplification

scripts/azure/destroy-old-clusters.sh
artem-belov: `for cluster in $CLUSTERS; do` will split it by rows, taps and spaces, I just need clusters list result to be splitted by rows.

scripts/azure/destroy-old-clusters.sh
artem-belov: @nickolaev I have simplified bash scripts. Could you please take a look at them.

scripts/azure/destroy-old-clusters.sh
lobkovilya: @artem-belov will it work correctly for `for cluster in $rows; do` ?

artem-belov: You are right, I have change script to foreach type. Thank you
```for cluster_info in "${rows[@]}"; do```

File: scripts/azure/destroy-old-clusters.sh [hunk 10]
```diff
@@ -0,0 +1,51 @@
...
+CLUSTERS=$(az aks list -g "nsm-ci" --query "[].{Name:name,Id:id,Group:resourceGroup}" -o tsv | grep "${pattern}")
+
+# shellcheck disable=SC2206
+IFS=$'\n' rows=($CLUSTERS); # Split result to string array by rows
+
+for cluster_info in "${rows[@]}"; do
+    # shellcheck disable=SC2206
+    IFS=$'\t' cols=(${cluster_info}); # Split each row to values array by \t separator ([0]=name, [1]=id, [2]=resourceGroup)
+    
+    last_activity=$(get_date "$(get_last_cluster_activity "${cols[1]}")")
```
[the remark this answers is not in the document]
artem-belov: last_activity is a last timestamp from cluster creation log. "age" is not correct word for that function

PR #1970, opened 2019-12-09
[WIP] Common endpoint service

a-bastrykina:
Signed-off-by: Alena Bastrykina <alena.bastrykina@xored.com>

## Description
Move endpoint_service to common package
## Motivation and Context
Now endpoint service is distinguished to the local and remote request cases. But the files `endpoint_service.go` in local and remote packages are almost duplicated, so we need to make unified endpoint service.

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


controlplane/pkg/common/request_builder.go
```diff
@@ -0,0 +1,101 @@
...
+package common
+
+import (
+	"github.com/networkservicemesh/networkservicemesh/controlplane/api/connection"
+	"github.com/networkservicemesh/networkservicemesh/controlplane/api/networkservice"
+	"github.com/networkservicemesh/networkservicemesh/controlplane/api/registry"
+	"github.com/networkservicemesh/networkservicemesh/controlplane/pkg/model"
+)
+
+type RequestBuilder interface {
```
lobkovilya: Are you thinking of adding unit tests for `RequestBuilder`?

a-bastrykina: Yes, I'm going to provide them in next commits

controlplane/pkg/common/request_builder.go
```diff
@@ -0,0 +1,101 @@
...
+	idGenerator func() string
+}
+
+type RemoteRequestBuilder struct {
+	nsmName     string
+	idGenerator func() string
+}
+
+func createLocalNSERequest(connectionId string, idGenerator func() string, nsmName string, localMechanisms []*connection.Mechanism, requestConn *connection.Connection) *networkservice.NetworkServiceRequest {
+	if connectionId == "" {
```
denis-tingajkin: Let's move the check for `connectionId` to method `Build`.
It will reduce args length for fucntion`createLocalNSERequest`.

controlplane/pkg/common/request_builder.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+type LocalRequestBuilder struct {
+	nsmName     string
+	idGenerator func() string
+}
+
+type RemoteRequestBuilder struct {
+	nsmName     string
+	idGenerator func() string
```
denis-tingajkin: I'm not sure about this function pointer. In all cases, we using for generation ` m.ConnectionID()`. I think we can get rid of `idGenerator func() string`

a-bastrykina: In this case, we need do store the whole model to the Builder, while Builder needs only its ID generation function.

denis-tingajkin:
@a-bastrykina also please fix linter issues. The job with linter has a green color but actually it has issues. Please check `lint` job output https://circleci.com/gh/networkservicemesh/networkservicemesh/117298?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link
I'll figure out why linter doesn't track problems...

controlplane/pkg/common/request_builder.go
a-bastrykina: Therefore we need to duplicate `connectionId` check twice, because `createLocalNSERequest` is called twice

controlplane/pkg/common/request_builder.go
denis-tingajkin: ok, I got it. Yes, you are right we need this function pointer.

a-bastrykina:
@lobkovilya Added tests, please check

controlplane/pkg/tests/request_builder_test.go
```diff
@@ -0,0 +1,169 @@
...
+		NetworkServiceName:        "network_service",
+	},
+}
+
+var testConnection = &connection.Connection{
+	Labels:         map[string]string{"key": "value"},
+	NetworkService: "network_service",
+}
+
+func TestBuildLocalNSERequestFromLocalEndpointService(t *testing.T) {
```
lobkovilya: Please consider renaming to `TestRequestBuilder_LocalReqestFromLocalEndpoint`

a-bastrykina: Done.

a-bastrykina:
Related issue is #2001

controlplane/pkg/common/request_builder.go
```diff
@@ -0,0 +1,108 @@
...
+			return m.ConnectionID()
+		},
+	}
+}
+
+// NewRemoteRequestBuilder creates new request builder for the endpoint service in remote service server
+func NewRemoteRequestBuilder(m model.Model) *RemoteRequestBuilder {
+	return &RemoteRequestBuilder{
+		nsmName: m.GetNsm().GetName(),
+		idGenerator: func() string {
```
denis-tingajkin: Can we use here `idGenerator: m.ConnecitonID` ?

controlplane/pkg/tests/endpoint_service_test.go
```diff
@@ -0,0 +1,250 @@
...
+	// Flag to check if Close() was called on client
+	closeCalled := false
+
+	nseClientMock.EXPECT().Close(gomock.Any(), gomock.Any()).DoAndReturn(func(_ context.Context, _ *connection.Connection) error {
+		closeCalled = true
+		return nil
+	})
+
+	// Mock NSE manager
+	nseManagerMock := mock.NewMockNetworkServiceEndpointManager(ctrl)
```
denis-tingajkin: Why we can't use for this test already existing stubs?
https://github.com/networkservicemesh/networkservicemesh/blob/1479afb615aca94ae3de155249d91258c21722ab/controlplane/pkg/nsm/nsm_heal_processor_test.go#L388
https://github.com/networkservicemesh/networkservicemesh/blob/1479afb615aca94ae3de155249d91258c21722ab/controlplane/pkg/nsm/nsm_heal_processor_test.go#L399

a-bastrykina: The realizations of `NetworkServiceEndpointManager`'s methods  in `nseManagerMock` and `nseManagerStub` differ, in my tests I need more simple logic for this methods, and some of them are not even called. 
For NSEClient, I need  `Close()` and `Reuquest()` methods, which are not implemented in `nseClientStub`. Of cource, I could add implementation, but I prefer to use mock opportunity. With usage of mocks, we will be able to get rid even of these `nseClientStub` `nseManagerStub` stubs too, and I think it will make code more brief and readable.

controlplane/pkg/common/request_builder.go
a-bastrykina: Fixed

controlplane/pkg/tests/endpoint_service_test.go
denis-tingajkin: >ith the usage of mocks, we will be able to get rid even of these nseClientStub nseManagerStub stubs too, 

@a-bastrykina sounds good to me

controlplane/pkg/tests/request_builder_test.go
```diff
@@ -0,0 +1,178 @@
...
+	return &model.Forwarder{
+		RemoteMechanisms: []*connection.Mechanism{
+			{
+				Type: vxlan.MECHANISM,
+			},
+		},
+	}
+}
+
+func getLocalTestEndpoint() *registry.NSERegistration {
```
alex-yust: 'get' means that object already exists and the function 'fetches' it. However, a new structure is generated for every call thus it seems more accurate to name the function {new|create}LocalTestEndpoint.

denis-tingajkin:
@a-bastrykina also I have some questions about mocking for nsm tests:
1. Should we add gen.go file for generation mocs? It could help to update tests if some interfaces will be modified.
2. Should we add into the documentation a few words about mock usage? It could help with writing unit tests.

a-bastrykina:
@denis-tingajkin 
1. That is a good idea.
2. Maybe, in documentation, we should mention the interfaces, which already have generated mocks, and recommend to use them.

PR #1989, opened 2019-12-13
Initial commit for Path API changes

edwarnicke:
See:
https://docs.google.com/presentation/d/1QU5FEq7QloLqEjJs-MMMWvcgPzkz6j-OYk-9k2gDTjc/edit#slide=id.g73e6edae28_0_0
https://drive.google.com/file/d/12AhnjrZnqwT5w8rP7Gv91xjKMgbQjPan/view?usp=sharing

This is the minimal patch that can introduce the new API.

As Path is replacing:

repeated string network_service_managers = 6;

in the Connection Message, this PR simply replaces how that was being used with Path.
It does not introduce the new style healing.

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


controlplane/api/connection/connection_helpers.go
```diff
@@ -28,16 +28,16 @@ func (c *Connection) IsRemote() bool {
 		return false
 	}
 	// If we have two or more, it is remote
-	return len(c.NetworkServiceManagers) > 1
+	return len(c.GetPath().GetPathSegments()) > 1
```
haiodo: It is not right with segments, since if will be more segments because of forwarder.

controlplane/api/connection/connection.proto
```diff
@@ -34,13 +35,24 @@ enum State {
   DOWN = 1;
 }
 
+message PathSegment {
+  string name      = 1;
+  string id        = 2;
+  string token = 3;
```
haiodo: @edwarnicke How is generating this token, is ever NSE could add some data encoded with it? For example if NSE is passing to another network service, and it would be good to have endpoint-id to be encoded here to be able to restore full sequence on same NSEs instead of new ones.

PR #1993, opened 2019-12-16
Don't force docker login on push makefile target

tiswanso:
Allow usage of `docker-push` target in already logged in environment.

## Description

Don't make me set my docker password in a env var to use the `docker-push` target.

e.g. if my shell is already logged into my registry let this work:  
```CONTAINER_REPO=tiswanso CONTAINER_TAG=myfix make docker-push  ```

## Motivation and Context

Ease of development workflows.

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### CODE REVIEW

File: .mk/build.mk [hunk 1]
```diff
@@ -96,7 +96,9 @@ docker-push: $(addsuffix -push,$(addprefix docker-,$(images)));
 
 .PHONY: docker-login
 docker-login:
-	@echo "${DOCKER_PASSWORD}" | docker login -u "${DOCKER_USERNAME}" --password-stdin
+	@if [ -n "${DOCKER_PASSWORD}" ]; then \
+		echo "${DOCKER_PASSWORD}" | docker login -u "${DOCKER_USERNAME}" --password-stdin ;\
+	fi
```
alex-yust: For diagnostic purpose, it might be good to add 'else' clause with echoing something like 'docker-login: bypassing authorization'. Otherwise it might take some time to debug a situation when DOCKER_PASSWORD is not set by mistake.

tiswanso: updated

Issue #1997, opened 2019-12-16
Unable to connect when requesting a specific endpoint owned by a different nsmgr


tiswanso:
## Expected Behavior

When a client performs a NetworkService request with a specific endpoint and nsmgr tuple, it should be connected.

## Current Behavior

The `controlplane/pkg/nsm/nse_manager.go` `func (nsem *nseManager) GetEndpoint(...` returns an error that it `Could not find endpoint with name: %s at local registry` because it always assumes the exact endpoint request is only for endpoints owned by itself (the local nsmgr).

## Steps to Reproduce

The issue is seen with virtual L3 example PoC using the latest implementation of the unified local/remote NetworkService request API:  https://github.com/tiswanso/examples/tree/on_unified_api/examples/vl3_basic

It finds all the endpoints of a specific network service and tries to connect to them via filling in the NetworkService request with the exact endpointName and network_service_manager.

See the specific code:
https://github.com/tiswanso/examples/blob/on_unified_api/examples/vl3_basic/vl3-nse/cmd/vl3_connect.go#L188-L199

## Context

* Kubernetes Version:
* etc.

nickolaev:
Wait. That is plain wrong. You don't let NSM Clients deal with he list of endpoints. That is not what the purpose of NSM is. You have a specific labels and let NSMgr deal with sorting out which Endpoint you get a connection to. Please reconsider you vl3 example implementation.

tiswanso:
Duplicating comment from PR here to keep issue from going stale... 

@nickolaev @lobkovilya -- per the meeting before the holidays, this should be a valid scenario that we don't have NSMgr's GetEndpoint error out in. Per the conversation, we want to be able to support a class of clients that know what exact endpoints they want to connect to. Do we still have disagreement here?

PR #1998, opened 2019-12-16
nsmgr interdomain: getEndpoint needs to check if specific endpoint is remote

tiswanso:
- if client is asking for exact endpoint it can't assume the endpoint is
  owned by itself--e.g. check the network_service_manager name.

Signed-off-by: Tim Swanson <tiswanso@cisco.com>

Fixes: #1997

## Description

Check whether the endpoint info in the request is for the current nsmgr via checking the network_service_manager name.  If it's not the same, perform a find and forward the request along.

*NOTE*:  this diff currently doesn't check the find result to match up with requested endpoint and name.  That will be added (possibly in a follow-on PR or this one if folks agree).

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/controlplane/pkg/nsm/nse_manager.go
+++ b/controlplane/pkg/nsm/nse_manager.go
@@ -31,13 +31,17 @@ func (nsem *nseManager) GetEndpoint(ctx context.Context, requestConnection *conn
 	span.LogObject("ignores", ignoreEndpoints)
 	// Handle case we are remote NSM and asked for particular endpoint to connect to.
 	targetEndpoint := requestConnection.GetNetworkServiceEndpointName()
+	myNsemName := nsem.model.GetNsm().GetName()
+	targetNsemName := requestConnection.GetDestinationNetworkServiceManagerName()
 	span.LogObject("targetEndpoint", targetEndpoint)
 	if len(targetEndpoint) > 0 {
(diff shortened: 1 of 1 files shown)


tiswanso:
I agree that we should consider carefully how the endpoint selection works and who is allowed to perform it.  However, this method just doesn't work if somehow it is called with a selected endpoint that is not on this node.  Conversely, the code base (& this method) is currently allowing the caller (a NSC/NSE or connection replay?) to select endpoints that are available in this nsmgr instance.

lobkovilya:
Method `GetEndpoint` could be called in 2 cases:
1. We are local NSMgr and we got request from NSC:
    in that case `targetEndpoint` is empty and we have to go to k8s-registry (using `discoveryClient`) and discover endpoints

2. We are remote NSMgr and we got request from NSMgr that works on another node:
    in that case `targetEndpoint` should contain endpoint that we have in model, otherwise we should fail. 

So NSC/NSE never choose Endpoint themselves, their local NSMgr must do that for them

tiswanso:
@nickolaev @lobkovilya  -- per the meeting before the holidays, this should be a valid scenario that we don't have NSMgr's GetEndpoint error out in.  Per the conversation, we want to be able to support a class of clients that know what exact endpoints they want to connect to.  Do we still have disagreement here?

Issue #1994, opened 2019-12-16
Interdomain:  fix proxy-nsmgr helm chart & expose nsmgr nsmd port


tiswanso:
## Expected Behavior

It should be possible to deploy NSM in multiple clusters with consistent properties to get the Interdomain feature to function.  

## Current Behavior

The current helm charts have multiple problems:
1. proxy-nsmgr's nsmd-k8s is not be able to discover networkservices (due to svcacct mapping)
1. unable to deploy proxy-nsmgr in insecure mode
1. unable to map proxy-nsmd-k8s (pNSR) to local port that is actually able to be exposed across clusters (defaults to 80).
1. unable to expose access to network service managers outside of the nodes due to not having a hostport mapping. 

### proxy-nsmgr service account problem
The helm chart doesn't specify the service-account to use for the proxy-nsmgr pods so they don't have cluster role binding access to a role that allows them to networkservice objects.

The following shows up in the `proxy-nsmgr` `proxy-nsmd-k8s` container's log:
```
time="2019-12-13T22:36:03Z" level=error msg="unable to list NSs for cache initialization: %v: networkservices.networkservicemesh.io is forbidden: User \"system:serviceaccount:nsm-system:default\" cannot list resource \"networkservices\" in API group \"networkservicemesh.io\" in the namespace \"default\""
```

### Proxy-nsmgr not able to be installed with insecure mode

The proxy-nsmgr's are unable to come up and get connections to any other NSM components due to `--set insecure="true"` not being consumed by the helm chart.

### Exposing pNSR local port

The proxy-nsmgr port mappings indicate that the hostport `5005` should map to containerPort `80`.  The code that tries to access the remote cluster's pNSR uses the same port num env var to also start the local pNSR grpc server--PROXY_NSMD_K8S_REMOTE_PORT.  So, if you need to access the remote cluster's pNSR at port 5005, your local pNSR has to be started on port 5005.  By default it tries to expose port 5005 (hostPort 5005) and listen on port 80.  So, there's no way to tell it to access port 5005 and still listen locally on port 80.

### Expose access to network service managers outside of the nodes

[...]

PR #2012, opened 2019-12-19
Interdomain helm tests and fixes

artem-belov:

## Description
Fixing issues in helm templates
Prepare interdomain tests for helm charts
Split icmp-responder nse and nsc

## Motivation and Context
NSM does not have interdomain helm ci tests
Issues #1994

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


tiswanso:
Please get the helm chart changes in under PR #1994.  It'd be good to get those in ASAP since they're totally busted for interdomain.

deployments/helm/proxy-nsmgr/templates/proxy-nsmgr.tpl
```diff
@@ -13,24 +13,38 @@ spec:
...
+              value: "false"
+{{- end }}
         - name: proxy-nsmd-k8s
           image: {{ .Values.registry }}/{{ .Values.org }}/proxy-nsmd-k8s:{{ .Values.tag }}
           imagePullPolicy: {{ .Values.pullPolicy }}
           ports:
-            - containerPort: 80
-              hostPort: 5005
+            - containerPort: 5005
+              hostPort: 80
```
tiswanso: taking hostPort 80 seems crazy to me.

artem-belov: Hello @tiswanso 
80 port was chosen in Interdomain Spec #714 and you can always change it with environment variable `PROXY_NSMD_K8S_REMOTE_PORT`
Also it very helpful for us because 80 port is opened by default in all clouds

tiswanso: I realize this is to get the tests to pass and the rationale is to try and implement the tests' interdomain deployment so that proxy-nsmgr's NSR is at the domain's port 80.  But in the tests the domain is the public IP of the node with proxy-nsmgr and not really a service exposed in a typical way via DNS and a loadbalancer which is more the spirit of what's in the Interdomain spec.  

My point is that, it's a bad idea to assume NSM owns the nodes by grabbing well known ports __on the node__.  By taking a node's port 80, the installation is very likely to conflict with something in a real k8s deployment.  In pub clouds port 80 opened externally because it is typically mapped to an API gateway of some sort (e.g. nginx/k8s-ingress/istio-gateway).  

So, IMO, this looks really hacky/crazy to any operator taking a look at NSM helm charts.  There's no real reason to expose the service like this other than what you're doing to make it work in CI tests.

tiswanso:
Could you explain the flow and how it tests interdomain?  If I'm not misunderstanding, the script `helm-test-charts.sh` was modified to take multiple lists of helm charts as params (space separated strings) and map them to clusters' kubeconfigs env vars.  Then it installs the charts mapped to the clusters and runs "make k8s-check" which ends up running nsc_ping_all.sh to see if the client can ping.

Is there a way to see the output of `helm-test-charts.sh` and any other cloudtest artifacts (STORE_POD_LOGS_DIR content)?  I wasn't able to find them on the job artifacts on circleci but maybe I wasn't looking in the right place.

artem-belov:
@tiswanso Hello.
You are right. Script is running "nsm proxy-nsmgr endpoint" charts on first cluster. Then it set **NETWORK_SERVICE** env var with proper network service name (with ip address of first cluster) and starts "nsm proxy-nsmgr client" on the second cluster. 
Logs you can find in **test-integration** ci job artifacts. For example for last ci run the path is: _".tests/
 cloud_test/azure-2_gke-1/001-Example-helm-interdomain-run.log"_

Issue #1728, opened 2019-10-07
Cleanup CI-specific docker images from DockerHub registry


alex-yust:
## Overview

Project's CI creates unique image tag for every build, which means images built during CI are never  (theoretically) de-tagged and cleaned up from DockerHub registry. Those images take pretty amount of space and create clutter in registry.

It seems right to cleanup the images by default.

Although there is no docker command to remove an image from a remote registry, it is possible
via Docker Registry HTTP API V2 https://devopsheaven.com/docker/dockerhub/2018/04/09/delete-docker-image-tag-dockerhub.html

edwarnicke:
Good idea :)

alex-yust:
done

PR #2016, opened 2019-12-20
Cleanup CI generated images from DockerHub

alex-yust:
#1728 

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -165,6 +165,13 @@ jobs:
             popd
             ./cloudtest
           no_output_timeout: 40m
+      - run:
+          name: Cleanup CI images from DockerHub unless it's master branch
+          command: |
+            [ "${CIRCLE_BRANCH}" == "master" ] && exit
+            echo "Cleaning up CI images from DockerHub.."
+            CONTAINER_TAG="ci_${CIRCLE_WORKFLOW_ID}_${CIRCLE_SHA1:8:8}"
+            REPO=networkservicemeshci TAGS="$CONTAINER_TAG" IMAGES=$(make docker-list) .circleci/dockerhub-image-remove
       - store_test_results:
           path: .tests/cloud_test/results/
       - store_artifacts:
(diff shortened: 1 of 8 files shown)


.circleci/wait-all-jobs-complete.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+function running_jobs_count() {
+    curl -s -u "4b586c69a308a01acfd126b225d2247cbe11c2d2:" \
```
denis-tingajkin: What means 4b586c69a308a01acfd126b225d2247cbe11c2d2?

.circleci/config.yml
```diff
@@ -251,6 +261,10 @@ workflows:
       - build-containers:
           requires:
             - test-sanity-check
+# cleanup
+      - cleanup-images:
+          requires:
+            - test-sanity-check
```
denis-tingajkin: Why we can't just wait here integration-test job?

alex-yust: 1) in such case the job will be run only if testing passes, which is not that we want
2) "docker-push-tag" job needs the images too

.circleci/wait-all-jobs-complete.sh
alex-yust: CircleCI API requires an access token to request info.
It is supposed to be CIRCLECI_TOKEN environment variable set in CircleCI config.

@edwarnicke can we have such a variable? The token from the code here is enough (or even prefered), since only reading capability is necessary.

.circleci/config.yml
denis-tingajkin: Is `circle ci` has not solutions for problems 1. 2 ?

alex-yust: I haven't found any way. 
What we need is to run the cleanup job after all of the (build, test, push) jobs are finished (if started), regardless of their exit statuses.

.circleci/wait-all-jobs-complete.sh
nickolaev: This is usually a bad practice. Even publishing it here sounds like a really bad security issue. Please remove.

alex-yust: done

PR #2020, opened 2019-12-24
WIP: Cleanup old CI generated images from DockerHub

alex-yust:
#1728 

The implementation that doesn't use separate worker to detect when CI workflow finishes.
Each CI run removes at most 10 tags of CI images older that given number of days, if there are any.
Currently there are about 2000 old tags for each container in CI repos.

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -228,6 +228,19 @@ jobs:
           command: |
             .circleci/helm/test-charts.sh
 
+  cleanup-images:
+    docker:
+      - image: circleci/golang:1.13-stretch
+    steps:
+      - checkout
+      - run:
+          name: Cleanup old CI images from DockerHub
+          command: |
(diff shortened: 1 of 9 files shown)


.circleci/config.yml
```diff
@@ -237,6 +228,19 @@ jobs:
           command: |
             .circleci/helm/test-charts.sh
 
+  cleanup-images:
+    docker:
+      - image: circleci/golang:1.13-stretch
+    steps:
+      - checkout
```
haiodo: Why we need this at all? Should every build to cleanup images produced? I think we need to run it once and at finish of build.

alex-yust: Please see #2016 - it has drawbacks performance and security wise

Issue #2013, opened 2019-12-19
Reconsider some testing bash sctipts in golang


artem-belov:
## Overview

@nickolaev suggested to move some existing testing bash scripts in golang, such as `helm-test-charts.sh` and cloud creating tools

Lets use this "Issue" for duscussion

artem-belov:
@edwarnicke @haiodo @lobkovilya @Denis @alex-yust @a-bastrykina

lobkovilya:
Personally I don't like the idea to eliminate shell tests.
If we had some UI, we would have automatic ui tests. Deploying of helm charts in some way our UI, because that what users contact with. Users contact with helm charts usually using bash. That's why I believe it's really important to us to have tests that reproduce user's steps as accurately as possible.

nickolaev:
Before #1247 our testing was a mess of shell scripts and unmaintainable CircleCI yaml:
https://github.com/networkservicemesh/networkservicemesh/pull/1247/files#diff-1d37e48f9ceff6d8030570cd36286a61

Fortunately this was removed with the excellent cloud-test framework that @haiodo introduced.
Now we are slowly returning to these testing practices. Yes, the people that use NSM will default to the helmcharts, that is where we have a nice guide
https://networkservicemesh.io/docs/setup/run/


```shell
helm repo add nsm https://helm.nsm.dev/ # Add the latest release nsm helm repo
helm install nsm/nsm # Install the nsm infrastructure in your Kubernetes Cluster
```

But I really fail to see how a shell script like this one helps any user to understand that:
https://github.com/networkservicemesh/networkservicemesh/blob/master/scripts/helm-test-charts.sh

```shell
#!/usr/bin/env bash

error(){ echo "error: $*" >&2; exit 1; }

[ -n "$*" ] && CHARTS="$*"
[ -n "$CHARTS" ] || error "chart name(s) expected as positional arguments or CHARTS variable"
[ -n "$NSM_NAMESPACE" ] || error "NSM_NAMESPACE must be set"

: "${CHECK_CMD:=make k8s-check}"

reverse() {
  { for e in "$@"; do echo "$e"; done } | tac | tr '\n' ' '
}

echo "Testing HELM chart deployment, chart(s): ${CHARTS}, namespace: ${NSM_NAMESPACE}"

for ch in $CHARTS; do
  dp="deployments/helm/$ch"
  [ -d "$dp" ] || error "Deployment $dp not found"
done

[...]

PR #2025, opened 2020-01-02
Simplification of shell-tests

lobkovilya:
Signed-off-by: Ilya Lobkov <ilya.lobkov@xored.com>

## Motivation and Context

Based on discussion in https://github.com/networkservicemesh/networkservicemesh/issues/2013
## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


lobkovilya:
I deleted `scripts/helm-test-charts.sh` with some logic and I added `test/shell/*_test.sh` files that contain only real steps to reproduce some use-case. 
That's exactly what I described here https://github.com/networkservicemesh/networkservicemesh/issues/2013#issuecomment-567471469

.mk/k8s.mk
```diff
@@ -143,10 +143,22 @@ k8s-forward:
 	@$(kubectl) port-forward $$($(kubectl) get pods | grep $(pod) | cut -d \  -f1) $(port1):$(port2)
 
 .PHONY: k8s-check
-k8s-check:
+k8s-check: k8s-icmp-check k8s-vpn-check
+
+.PHONY: k8s-icmp-check
```
denis-tingajkin: Should we also update [example doc]( https://github.com/networkservicemesh/networkservicemesh/blob/7f7a7730a5d2ea0dce04c7819647458a92c35195/docs/vagrant/guide-vagrant.md#verify-deploying-the-examples) to use `k8s-icmp-check` / `k8s-vpn-check` ?

lobkovilya: I didn't change the behaviour of `make k8s-check` it still runs 2 checks as it was before. Our bash-checkers also require some simplification, I intended to do it in further PRs

denis-tingajkin: in the doc for verifying `icmp example`need to execute `make k8s-check`. Script `test/shell/icmp_test.sh` uses for verifying  `make k8s-icmp-check`. I think examples from docs and examples from CI should be equal.

Issue #2007, opened 2019-12-17
NSMD doesn't recover connection in case of interdomain


denis-tingajkin:
## Steps to reproduce

1) Rebase on https://github.com/networkservicemesh/networkservicemesh/pull/1861
2) Run tests:
2.1) TestInterdomainNSEHealLocalToRemote
2.2) TestInterdomainNSMHealRemoteDieNSMD_NSE // will faill. NSMD doesn't recover a connection.

## Expected Behavior

NSMD should recover a connection

## Current Behavior

NSMD doesn't recover a connection

## Logs
https://118168-127937836-gh.circle-artifacts.com/0/home/circleci/project/.tests/cloud_test/aws-2_azure-2/002-TestInterdomainNSMHealRemoteDieNSMD_NSE-run.log

NSMD logs
[nsmgr-ip-192-168-130-59.us-east-2.compute.internal-nsmd.log](https://github.com/networkservicemesh/networkservicemesh/files/3973874/nsmgr-ip-192-168-130-59.us-east-2.compute.internal-nsmd.log)

nickolaev:
Issue fixed with #2030. Closing.

PR #2030, opened 2020-01-06
Fix: "NSMD doesn't recover connection in case of interdomain"

denis-tingajkin:

## Motivation and Context
https://github.com/networkservicemesh/networkservicemesh/issues/2007

## How Has This Been Tested?

- [X] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


controlplane/pkg/nsmd/nsmd_crossconnect_client_test.go
```diff
@@ -0,0 +1,66 @@
...
+	waitConnectionCount(c.remotePeers["a"], 1)
+	assert.Expect(len(c.remotePeers["a"].connections)).ShouldNot(gomega.BeZero())
+	cancel()
+	waitConnectionCount(c.remotePeers["a"], 0)
+	assert.Expect(len(c.remotePeers["a"].connections)).Should(gomega.BeZero())
+}
+
+func waitConnectionCount(rm *remotePeerDescriptor, count int) {
+	lastTime := time.Now()
+	for time.Now().Before(lastTime.Add(time.Second / 2)) {
```
alex-yust: more readable IMO would be: 
time.Now().Since(startTime) < time.Second / 2

controlplane/pkg/nsmd/nsmd_crossconnect_client.go
```diff
@@ -148,7 +147,7 @@ func (client *NsmMonitorCrossConnectClient) startPeerMonitor(clientConnection *m
 		remotePeer.connections[clientConnection.ConnectionID] = clientConnection
 		return
 	}
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, cancel := context.WithCancel(ctx)
```
haiodo: This is not right, since ctx is client Request context if we came here for first time.

controlplane/pkg/nsmd/nsmd_crossconnect_client_test.go
```diff
@@ -0,0 +1,66 @@
...
+		if len(rm.connections) == count {
```
lobkovilya: I think you need to lock before checking the length of `connections`

controlplane/pkg/nsmd/nsmd_crossconnect_client.go
```diff
@@ -452,14 +454,16 @@ func (client *NsmMonitorCrossConnectClient) remotePeerConnectionMonitor(ctx cont
...
 		span.LogError(err)
-		// Same as DST down case, we need to wait for same NSE and updated NSMD.
 		connections := client.xconManager.GetClientConnectionByRemote(remotePeer)
 		for _, cc := range connections {
 			ccSpan := common.SpanHelperFromConnection(ctx, cc, "remotePeerConnectionMonitor.update")
 			defer ccSpan.Finish()
 			ccSpan.LogObject("clientConnection", cc)
 			client.xconManager.DestinationDown(ccSpan.Context(), cc, true)
 		}
+		client.remotePeerLock.Lock()
```
haiodo: It seems we need to update this map properly and not here, please check why it was not empty.

denis-tingajkin: yes, you are right. The problem was happening because monitoring goroutine sometimes closes (for example stream returned error) and the listener `AddCliendtConnection` not checks that the goroutine already finished.

Scenario:
1. Peer monitoring goroutine finishes by received error 
2. Model receives new client connection
3. Method `AddCliendtConnection` looks into a map of peers and see that peer already exists and exits.

This PR  fixes this scenario.

controlplane/pkg/nsmd/remote_peer_descriptor.go
```diff
@@ -0,0 +1,91 @@
...
+	RemoveConnection(connection *model.ClientConnection)
+	Cancel()
+	Context() context.Context
+	Reset()
+	IsCanceled() bool
+	HasConnection() bool
+	RemoteNsm() *registry.NetworkServiceManager
+}
+
+type remotePeerDescriptor struct {
```
lobkovilya: Probably makes sense to make that structure thread safe

denis-tingajkin: this structure using only in locked sections. I do not think that we need to make this structure thread-safe.

lobkovilya: At least here you pass it inside go routine https://github.com/networkservicemesh/networkservicemesh/pull/2030/files#diff-cd75ec725c492a08d9a9a6154b4e8418R143

denis-tingajkin: there missing lock sections. Thanks, fixed.

controlplane/pkg/nsmd/nsmd_crossconnect_client.go
```diff
@@ -428,39 +420,45 @@ func (client *NsmMonitorCrossConnectClient) handleXconEvent(event monitor.Event,
...
-		connections := client.xconManager.GetClientConnectionByRemote(remotePeer)
+		connections := client.xconManager.GetClientConnectionByRemote(remoteNsm)
 		for _, cc := range connections {
 			ccSpan := common.SpanHelperFromConnection(ctx, cc, "remotePeerConnectionMonitor.update")
 			defer ccSpan.Finish()
 			ccSpan.LogObject("clientConnection", cc)
 			client.xconManager.DestinationDown(ccSpan.Context(), cc, true)
 		}
 	}
+	client.remotePeerLock.Lock()
```
lobkovilya: `remotePeerConnectionMonitor` in charge of only one remote peer, why it locks the whole map of peers?

denis-tingajkin: Peer canceling should be synchronized with Adding/Removing client connection

lobkovilya: That's why makes sense to add locks inside each RemotePeerDescriptor

denis-tingajkin: thanks, fixed.

controlplane/pkg/nsmd/remote_peer_descriptor.go
denis-tingajkin: finally fixed, thanks

PR #2036, opened 2020-01-08
Replace registration.sh with go-based side-car

lobkovilya:
## Motivation and Context

Go side-car is more convenient way to register SPIFFE entries. Also side-car will be useful for configuring spire-servers across several clouds. 
## How Has This Been Tested?

- [X] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


side-cars/cmd/nsm-spire/main.go
```diff
@@ -0,0 +1,111 @@
...
+	var err error
+	c := tools.NewOSSignalChannel()
+
+	var serverConn *grpc.ClientConn
+	for {
+		serverConn, err = tools.DialUnixInsecure(serverSocketPath)
+		if err == nil {
+			break
+		}
+		logrus.Errorf("failed to dial server: %v", err)
```
haiodo: I think a short delay should be good idea to be added here.

nickolaev:
I am afraid we are proliferating side-cars too much. Shall we have a unified one?

side-cars/cmd/nsm-spire/main.go
lobkovilya: `tools.DialUnixInsecure` is blocking with timeout 15 sec by default

lobkovilya:
@nickolaev that's interesting point, what would you like to see in unified side-car?

haiodo:
I think yes we could have unified one, for monitor/spire/dns stuff all together.

Issue #2086, opened 2020-01-30
testing: cluster initialization failure is ignored


alex-yust:
## Expected Behavior

initialization should be tried once again, cluster to be marked as crashed if failed.

## Current Behavior

cluster is considered working, causing all tests to fail

## Failure Information (for bugs)

See gke-3:
- startup: 
   https://120466-127937836-gh.circle-artifacts.com/0/home/circleci/project/.tests/cloud_test/gke-3/002-start.log
- prepare: 
  https://120466-127937836-gh.circle-artifacts.com/0/home/circleci/project/.tests/cloud_test/gke-3/003-prepare.log
- CI build: 
https://circleci.com/gh/networkservicemesh/networkservicemesh/120466#artifacts/containers/0

## Context

There are errors in 'start' and 'prepare' scripts, both are ignored

alex-yust:
fixed

PR #2088, opened 2020-01-30
fix: cluster initialization failure ignored

alex-yust:
fix for #2086 

Additionally, a waiting is added for unavailable services to become available, which prevents helm initialization errors on GKE, like here: 
https://120466-127937836-gh.circle-artifacts.com/0/home/circleci/project/.tests/cloud_test/gke-3/003-prepare.log

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

### DIFF
--- a/scripts/prepare-cluster.sh
+++ b/scripts/prepare-cluster.sh
@@ -1,15 +1,5 @@
 #!/bin/bash
 
-set -x
-
-function prepare() {
-  echo "Preparing cluster..."
-
-  make helm-init
-  make helm-install-config
-  make spire-install
-}
-
 # workaround for https://github.com/helm/helm/issues/6361:
 function delete_unavailable_apiservice() {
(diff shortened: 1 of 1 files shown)


scripts/prepare-cluster.sh
```diff
@@ -1,15 +1,25 @@
 #!/bin/bash
 
-set -x
-
 function prepare() {
   echo "Preparing cluster..."
+  make -i helm-init helm-install-config spire-install
```
lobkovilya: Why `-i`?

alex-yust: It's an attempt to add error detection in executing first two rules while preserving old workflow: all targets are executed regardless of errors in previous rules. 
In fact, it's a piece that was put before 'waiting for services' code, '-i' might be unnecessary now. Let try to remove it..

Issue #2093, opened 2020-02-04
Helm charts install command should not require --dep-up option


alex-yust:
Currently, nsm chart depend on config chart which contain CRDs used to implement k8s-config make target used for testing.

Also, docs/guide-helm.md and Helm CI tests should be rewritten to use bare Helm commands instead of 'make' targets.

nickolaev:
I think it is more or less a matter to call `make k8s-config` before the `make helm-install-nsm` and remove the `file://../config` from `requirements.yaml`.

alex-yust:
done

PR #2094, opened 2020-02-04
Helm charts install command should not require --dep-up option

alex-yust:
#2093 

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


alex-yust:
> Please do not move the config inside the core NSM deployment charts. Setting up the cluster shall remain in `make k8s-config` and in its own `config` chart.

but.. the only way to not use '--dep-up' option is to have config as subchart on 'nsm' .

alex-yust:
> Please do not move the config chart to the nsm one. We want clean NSM deployment not bound to the pre-configuration. For example, I would expect that the OpenShift integration would have some different approach to the overall deployment.

@nickolaev Please note that the pre-configuration was always a part of nsm chart, directly (embedded CRDs) or via dependency chart (recently). 
Do I understand correctly that besides the PR objective you want also to remove pre-configuration step from nsm chart to open the possibility to integrate OpenShift in future?

nickolaev:
@alex-yust right, OK let's merge this and we can follow up to move the configs out of the NSM chart if we decided to.

PR #2127, opened 2020-03-13
Field for arbirary k,v data in the message IPContext

uablrek:
## Description

Add in message IPContext;
```
map<string, string> extra_context = 11; /* a k,v map for transfering arbitrary information from an NSE to an NSC or dataplane */
```

## Motivation and Context

Often some extra data needs to be exchanged. This PR makes that possible without altering the protocol every time.

## How Has This Been Tested?
- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


haiodo:
Hi @uablrek,

I personally think IPContext is not a good candidate for any custom information, better to have a additional context object with this information.

uablrek:
@haiodo I have no opinion myself in this matter. I just created this PR whith this feature that I tought was agreed with @edwarnicke.

edwarnicke:
@haiodo I think the information that @uablrek has in mind is in fact IPContext... just not of the sort we've envisioned so far.  I think this could be done either here (and in EthernetContext, and DNSContext, etc) or we could do it once at the ConnectionContext level.  I'm inclined to suggest additional_context as a name at the ConnectionContext level.  @uablrek thoughts?

uablrek:
Sounds reasonable to have it in ConnectionContext since it is supposed to be general parameters. I will make the change.

edwarnicke:
@uablrek Looks great... could you also push a PR to https://github.com/networkservicemesh/api

PR #2132, opened 2020-03-17
Add podman support for local builds and deployments

rktidwell:
This change adds support for podman. This is useful in environments
where the container runtime is CRI-O instead of Docker. Support is
achieved by enabling Makefile's and build scripts to test for the
docker CLI and fall back to podman if docker is not available, relying
on CLI argument compatibility between the two tools.

See https://github.com/networkservicemesh/examples/issues/87

Signed-off-by: Ryan Tidwell <rtidwell@suse.com>

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:


.mk/build.mk
```diff
@@ -23,8 +23,10 @@ VPP_AGENT=ligato/vpp-agent:v3.0.1
 CGO_ENABLED=0
 GOOS=linux
 DOCKER=./build
-GO_BUILD_ENV = CGO_ENABLED=$(CGO_ENABLED) GOOS=$(GOOS) 
+GO_BUILD_ENV = CGO_ENABLED=$(CGO_ENABLED) GOOS=$(GOOS)
 GO_BUILD = ${GO_BUILD_ENV} go build -ldflags "-extldflags '-static' -X  main.version=$(VERSION)"
+# Prefer docker, but fall back to podman for docker-compatible tooling depending on the container runtime available
+RUNTIME = $(shell command -v docker || command -v podman)
```
haiodo: I would prefer a better name, but in general looks good. Thanks.

Issue #2169, opened 2020-07-10
Interdomain NSM: client and endpoint on the same cluster node fails


terror96:
## Expected Behavior

Client and endpoint can reside on the same cluster node in the case of interdomain NSM.

## Current Behavior

Client fails to init.

## Failure Information (for bugs)

nsm-vpp-forwarder-lspbp: level=error msg="VPPApiError: Source and destination are the same (-82)" loc="descriptor/interface_crud.go(95)" logger=vpp-ifplugin.if-descriptor

nsmd: level=error msg="NSM:(9.1.1) Forwarder request failed: rpc error: code = FailedPrecondition desc = KeyErrors: [config/vpp/v2/interfaces/SRC-b (CREATE): VPPApiError: Source and destination are the same (-82)] retry: 4" operation=ProgramAttempt-4

## Steps to Reproduce

Create two k8s clusters (cluster 1 and cluster 2) with one worker node in each, and deploy nsmrs in cluster 1. Deploy nsm and proxy-nsmgr to both clusters. Launch both endpoint and client in cluster 1, and make sure that they are spawned on the same worker node.

Please note. We are using OpenStack VMs, not kind etc.

## Failure Logs

The above messages are extracted from 'nsm-vpp-forwarder-lspbp' and 'nsmd' in cluster 1.

denis-tingajkin:
@terror96 Am I understand correctly that the client wants "networkservice@client.domain"?

denis-tingajkin:
It is a good catch :) 

@edwarnicke Should we fix this issue for monorpo?

edwarnicke:
@denis-tingajkin Please do :)

@terror96 Thanks for reporting the issue :)

terror96:
> @terror96 Am I understand correctly that the client wants "${networkservice name}@${client domain}"?

The client is configured to request endpoint at ${networkservice name}@${NSMRS}.

denis-tingajkin:
@terror96 The problem should be fixed with #2170, can you please check? :)

terror96:
> @terror96 The problem should be fixed with #2170, can you please check? :)

Sure :) Yes, I can confirm that it now works also in our OS setup. Good job, thanks.

denis-tingajkin:
Fixed by #2170

tiswanso:
[...]

PR #2170, opened 2020-07-20
fix: Interdomain NSM: client and endpoint on the same cluster node fails

denis-tingajkin:

## Description
This PR fixes bug 2169

## Motivation and Context
https://github.com/networkservicemesh/networkservicemesh/issues/2169

## How Has This Been Tested?

- [X] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/test/kubetest/pods/proxynsmd.go
+++ b/test/kubetest/pods/proxynsmd.go
@@ -72,6 +72,16 @@ func ProxyNSMgrPodWithConfig(name string, node *v1.Node, config *NSMgrPodConfig)
 							ContainerPort: 5005,
 						},
 					},
+					Env: []v1.EnvVar{
+						{
+							Name: "NODE_NAME",
+							ValueFrom: &v1.EnvVarSource{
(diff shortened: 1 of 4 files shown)


test/kubetest/pods/proxynsmd.go
```diff
@@ -72,6 +72,16 @@ func ProxyNSMgrPodWithConfig(name string, node *v1.Node, config *NSMgrPodConfig)
 							ContainerPort: 5005,
 						},
 					},
+					Env: []v1.EnvVar{
+						{
+							Name: "NODE_NAME",
```
denis-tingajkin, line 77: Note: We already have this env for helm: https://github.com/networkservicemesh/networkservicemesh/blob/master/deployments/helm/proxy-nsmgr/templates/proxy-nsmgr.tpl#L38. So this patch should be work for helm cases.

haiodo: Nice.

PR #2193, opened 2020-09-24
Multiple services per one endpoint

xzfc:
## Description

Before to this PR, `NsmEndpoint` performed only a single `RegisterNSE` gRPC call during the execution of the `Start()` method. Considering that multiple `NsmEndpoint` couldn't share a single client socket, it was not possible to register multiple NSEs per single container.

This PR extends `NsmEndpoint` constructor to accept options. Multiple registrations could be specified using the new option `WithRegistrations`.

This change should be mostly backward compatible with the old code since only the optional variadic argument is added.

## Motivation and Context

#2186

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/sdk/endpoint/endpoint.go
+++ b/sdk/endpoint/endpoint.go
@@ -21,6 +21,7 @@ import (
 	"io"
 	"net"
 
+	"github.com/hashicorp/go-multierror"
 	"github.com/pkg/errors"
 
 	"github.com/networkservicemesh/networkservicemesh/pkg/tools/jaeger"
@@ -44,13 +45,28 @@ type NsmEndpoint interface {
 	Delete() error
 }
 
+// Registration stores parameters that NsmEndpoint passes to the RegisterNSE
+// gRPC call.
+type Registration struct {
+	Name   string
+	Labels map[string]string
+}
+
+// Option is an option that can be given to a NsmEndpoint on construction.
+type Option func(*nsmEndpoint)
+
 type nsmEndpoint struct {
 	*common.NsmConnection
 	service        networkservice.NetworkServiceServer
 	grpcServer     *grpc.Server
 	registryClient registry.NetworkServiceRegistryClient
-	endpointName   string
+	registrations  []registration
 	tracerCloser   io.Closer
+}
+
+type registration struct {
+	Registration
+	registeredName string
 }
 
 func (nsme *nsmEndpoint) setupNSEServerConnection() (net.Listener, error) {
@@ -84,7 +100,6 @@ func (nsme *nsmEndpoint) Start() error {
 	unified.RegisterNetworkServiceServer(nsme.grpcServer, nsme)
 
 	listener, err := nsme.setupNSEServerConnection()
-
 	if err != nil {
 		logrus.Errorf("Unable to setup NSE")
 		return err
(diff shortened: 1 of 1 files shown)


sdk/endpoint/endpoint.go
```diff
@@ -44,15 +44,27 @@ type NsmEndpoint interface {
 	Delete() error
 }
 
+// NsmEndpointRegistration stores parameters that NsmEndpoint passes to the
+// RegisterNSE gRPC call.
+type NsmEndpointRegistration struct {
```
denis-tingajkin, line 49: In general, it is a nice solution for me:) 
```suggestion
type EndpointRegistration struct {
```

sdk/endpoint/switch_builder.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/pkg/errors"
+
+	"github.com/networkservicemesh/networkservicemesh/controlplane/api/connection"
+	"github.com/networkservicemesh/networkservicemesh/controlplane/api/networkservice"
+)
+
+// SwitchEndpointBuilder is a helper to construct SwitchEndpoint instance
+// which matches incoming requests based on Connection.NetworkService and
+// Connection.Labels.
+type SwitchEndpointBuilder struct {
```
denis-tingajkin, line 13: I do not think that actually we need this entity because it could look complex for the user

sdk/endpoint/switch.go
```diff
@@ -0,0 +1,61 @@
...
+// incoming requests into one of its items specified by user-provided matcher
+// function.
+type SwitchEndpoint struct {
+	// endpointMap    map[string][]SwitchEndpointItem
+	items   []networkservice.NetworkServiceServer
+	matcher ConnectionEndpointMatcher
+}
+
+// NewSwitchEndpoint returns a new SwitchEndpoint.
+func NewSwitchEndpoint(items []networkservice.NetworkServiceServer, matcher ConnectionEndpointMatcher) *SwitchEndpoint {
```
denis-tingajkin, line 27: I would be suggesting to simplify API to something like:

```suggestion
func NewSwitchEndpoint(items ...networkservice.NetworkServiceServer *SwitchEndpoint {
return func NewSwitchEndpointWithCustomMatchers(items..., DefaultMatcher1(), DefaultMatcher2(), ...)
```

sdk/endpoint/switch.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/golang/protobuf/ptypes/empty"
+
+	"github.com/networkservicemesh/networkservicemesh/controlplane/api/connection"
+	"github.com/networkservicemesh/networkservicemesh/controlplane/api/networkservice"
+	"github.com/networkservicemesh/networkservicemesh/pkg/tools/spanhelper"
+)
+
+// ConnectionEndpointMatcher is a function that accepts a connection and returns
+// the corresponding network service server.
+type ConnectionEndpointMatcher func(*connection.Connection) (networkservice.NetworkServiceServer, error)
```
denis-tingajkin, line 15: I do not think that the custom func should return network service.NetworkServiceServer.

```suggestion
type ConnectionEndpointMatcher func(*connection.Connection, config Configuration) bool
```

denis-tingajkin (changes requested):

sdk/endpoint/endpoint.go
xzfc: The linter doesn't like repetition (`endpoint.EndpointRegistration`), so I'll name it `endpoint.Registration` then.

denis-tingajkin: endpoint.Registration sounds better :)

xzfc:
For the record:

Initially, we planned to implement a general purpose switch endpoint (which could be found in [this commit](https://github.com/xzfc/networkservicemesh/commit/2c8245eb201a725dee2e63ca7b0adde96472b808)). But then we decided not to include it into the SDK due to the low demand for this feature and how easy it is to implement a specialized custom endpoint in the user code.

sdk/endpoint/endpoint.go
```diff
@@ -180,8 +208,26 @@ func (nsme *nsmEndpoint) Close(ctx context.Context, incomingConnection *connecti
 	return &empty.Empty{}, nil
 }
 
-// NewNSMEndpoint creates a new NSM endpoint
+// NewNSMEndpoint creates a new NSM endpoint with a single registration.
 func NewNSMEndpoint(ctx context.Context, configuration *common.NSConfiguration, service networkservice.NetworkServiceServer) (NsmEndpoint, error) {
+	registration := []Registration{MakeNsmEndpointRegistration(configuration)}
+	return NewNSMEndpointWithRegistrations(ctx, configuration, service, registration)
+}
```
denis-tingajkin, line 215: NOTE: you can just use this construction
```suggestion
func NewNSMEndpoint(ctx context.Context, configuration *common.NSConfiguration, service networkservice.NetworkServiceServer, registrations...Registartion) (NsmEndpoint, error) {
	if len(registrations) == 0 {
		registrations := []Registration{MakeNsmEndpointRegistration(configuration)}
	}
	return NewNSMEndpointWithRegistrations(ctx, configuration, service, registration) //TODO inline
}
```

denis-tingajkin (approved):
In general looks good. 
This PR allows to endpoints register multiple network services.

sdk/endpoint/endpoint.go
xzfc: I think that having an (additional) function that takes an array is more clear than a variadic function with an implicit `len == 0` case.

denis-tingajkin: Please take a look at the go `option` pattern. I think actually we have a structure that has a default value of registrations `[]Registration{MakeNsmEndpointRegistration(configuration)}`

sdk/endpoint/endpoint.go
```diff
@@ -44,15 +44,35 @@ type NsmEndpoint interface {
...
 	grpcServer     *grpc.Server
 	registryClient registry.NetworkServiceRegistryClient
-	endpointName   string
+	registrations  []registration
 	tracerCloser   io.Closer
 }
 
+type nsmEndpointOptions struct {
+	registrations []Registration
+}
```
denis-tingajkin, line 69:
```suggestion

```

sdk/endpoint/endpoint.go
```diff
@@ -44,15 +44,35 @@ type NsmEndpoint interface {
...
+// Registration stores parameters that NsmEndpoint passes to the RegisterNSE
+// gRPC call.
+type Registration struct {
+	Name   string
+	Labels map[string]string
+}
+
+// NSMEndpointOption is an option that can be given to a NsmEndpoint on
+// construction.
+type NSMEndpointOption func(*nsmEndpointOptions)
```
denis-tingajkin, line 56:
```suggestion
type Option func(*nsmEndpoint)
```

sdk/endpoint/endpoint.go
```diff
@@ -196,10 +241,34 @@ func NewNSMEndpoint(ctx context.Context, configuration *common.NSConfiguration,
...
 		service:       service,
+		registrations: registrations,
 	}
 
 	return endpoint, nil
 }
+
+// WithRegistrations returns NSMEndpointOption for specifying a list of
+// registrations. When not specified, a single registration is performed.
+func WithRegistrations(registrations []Registration) NSMEndpointOption {
```
denis-tingajkin, line 262:
```suggestion
func WithRegistrations(registrations...Registration) NSMEndpointOption {
```

sdk/endpoint/endpoint.go
```diff
@@ -97,54 +111,71 @@ func (nsme *nsmEndpoint) Start() error {
...
+	return err
+}
+
+func (nsme *nsmEndpoint) Delete() error {
+	var err error
+	for i := range nsme.registrations {
+		err1 := nsme.deregister(&nsme.registrations[i])
+		if err == nil && err1 != nil {
+			err = err1
+		}
```
denis-tingajkin, line 177: This check could ignore some errors...
Please consider using `errors.Wrapf` to put all errors into result.

Also, let's do not use names like `err1`. I would suggest `unregisterErr`

sdk/endpoint/endpoint.go
```diff
@@ -97,54 +111,71 @@ func (nsme *nsmEndpoint) Start() error {
...
+	span.LogObject("endpoint-name", r.registeredName)
 	span.Logger().Infof("NSE registered: %v", registeredNSE)
 	span.Logger().Infof("NSE: channel has been successfully advertised, waiting for connection from NSM...")
-
-	return nil
 }
 
-func (nsme *nsmEndpoint) Delete() error {
-	span := spanhelper.FromContext(context.Background(), fmt.Sprintf("Endpoint-%v-Delete", nsme.Configuration.EndpointNetworkService))
+func (nsme *nsmEndpoint) deregister(r *registration) error {
```
denis-tingajkin, line 156: [minor]
```suggestion
func (nsme *nsmEndpoint) unregister(r *registration) error {
```

denis-tingajkin: It can add some consistent with the registry API https://github.com/networkservicemesh/api/blob/master/pkg/api/registry/registry.proto#L40

sdk/endpoint/endpoint.go
```diff
@@ -199,7 +232,36 @@ func NewNSMEndpoint(ctx context.Context, configuration *common.NSConfiguration,
 	endpoint := &nsmEndpoint{
 		NsmConnection: nsmConnection,
 		service:       service,
+		registrations: nil,
+	}
+
+	// Default options
+	WithRegistrations(MakeRegistration(configuration))(endpoint)
```
denis-tingajkin, line 239:
```suggestion
		registrations: []*Registration{MakeRegistration(configuration)},
	}
```

sdk/endpoint/endpoint.go
xzfc: Used `go-multierror` as in `client.go`.

sdk/endpoint/endpoint.go
xzfc: It should be a bit more verbose since it should be an array of nested registration objects, but OK.

PR #2201, opened 2020-10-02
Update Kind setup instructions

electrocucaracha:
The instructions provided by [this guide](https://github.com/networkservicemesh/networkservicemesh/blob/master/docs/guide-build.md#install) seems to be out of dated.

## Description
- KinD has [dropped](https://github.com/kubernetes-sigs/kind/commit/3abbcbbb99cec020a052c97eeaf427229b7514ac) `v1alpha3`
- `SPIRE_ENABLED` and `--spire_enabled` aren't validated in the `helm-nsm-install.sh` script.
- `NSM_NAMESPACE` uses `nsm-system` as default value which is not created.
-  `SPIRE_ENABLED` is enabled by default

## Motivation and Context

These changes allow the creation and installation of a KinD cluster.

## How Has This Been Tested?

- [x] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/docs/guide-build.md
+++ b/docs/guide-build.md
@@ -32,7 +32,7 @@ And if you are using the Kind machinery to run your K8s cluster (described a bit
 make k8s-save
 ```
 
-because ```make k8s-save``` will build your containers and save them in `scripts/vagrant/images` where they can be loaded by the Kins K8s cluster.
+because ```make k8s-save``` will build your containers and save them in `scripts/vagrant/images` where they can be loaded by the Kind K8s cluster.
 
 You can also selectively rebuild any component, say the `nsmd`, with ```make k8s-nsmd-save```
 
--- a/scripts/kind.yaml
+++ b/scripts/kind.yaml
@@ -1,6 +1,6 @@
 ---
 kind: Cluster
-apiVersion: kind.sigs.k8s.io/v1alpha3
+apiVersion: kind.x-k8s.io/v1alpha4
 nodes:
   - role: control-plane
   - role: worker


denis-tingajkin (approved):

.mk/k8s.mk
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-NSM_NAMESPACE?=nsm-system
+NSM_NAMESPACE?=default
```
edwarnicke, line 15: Curious why the namespace change here?

.mk/k8s.mk
```diff
@@ -37,7 +37,7 @@ include .mk/null.mk
 
 include .mk/gke.mk
 
-SPIRE_ENABLED?=true
+SPIRE_ENABLED?=false
```
edwarnicke, line 40: Curious why we are defaulting to SPIRE_ENABLED=false ?

docs/guide-build.md
```diff
@@ -32,7 +32,7 @@ And if you are using the Kind machinery to run your K8s cluster (described a bit
 make k8s-save
 ```
 
-because ```make k8s-save``` will build your containers and save them in `scripts/vagrant/images` where they can be loaded by the Kins K8s cluster.
+because ```make k8s-save``` will build your containers and save them in `scripts/vagrant/images` where they can be loaded by the Kind K8s cluster.
```
edwarnicke, line 35: Thanks for fixing the typo :)

.mk/k8s.mk
electrocucaracha: I didn't know that a `spire-install` make action exists so I think it's better to change the documentation mentioning this action.

.mk/k8s.mk
edwarnicke: Is it not created by the helm chart?

.mk/k8s.mk
edwarnicke: I'm starting to feel like perhaps our issues are in some gaps between our Makefiles and the helm charts.  @denis-tingajkin  @haiodo  you've looked at this more recently than I have, thoughts?

denis-tingajkin: spire is enabled by default, I think we should keep this behaviour

.mk/k8s.mk
denis-tingajkin: In helm charts, we are using the built-in object `Release.Namespace`. And here I think we can still use `NSM_NAMESPACE?=nsm-system`. NOTE: you can set NSM_NAMESPACE=default and invoke any `make` target :)

scripts/kind.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 kind: Cluster
-apiVersion: kind.sigs.k8s.io/v1alpha3
+apiVersion: kind.x-k8s.io/v1alpha4
```
denis-tingajkin, line 3: @edwarnicke I've tested locally nsm works correctly with this version.

edwarnicke: @Do you mean kind works correctly with v1alpha4 or v1alpha3?

denis-tingajkin: I meant this version `kind.x-k8s.io/v1alpha4`

edwarnicke:
@electrocucaracha Would you be willing to try a version of this PR that updates the API version to v1alpha4 but doesn't change the other defaults wrt spire etc?

electrocucaracha:
> @electrocucaracha Would you be willing to try a version of this PR that updates the API version to v1alpha4 but doesn't change the other defaults wrt spire etc?

@edwarnicke, makes sense I'm still working with @denis-tingajkin trying to understand the actions required by the setup

denis-tingajkin:
@electrocucaracha can you please rebase branch to the latest master?

Issue #2202, opened 2020-10-03
proxy-nsmd-k8s always filters for local NSEs


tiswanso:
## Expected Behavior

FindNetworkService RPC calls should return all endpoints (NSEs) available in a domain for a NetworkService, even if the call is for a remote (interdomain) NSR.  

## Current Behavior

If there's local endpoints (NSEs), proxy-nsmd-k8s will only return those local endpoints (NSEs). 

## Steps to Reproduce

1. Deploy interdomain NSM with an interdomain NSR
2. Deploy a network service with NSEs in all domains
3. Trigger FindNetworkService() for the interdomain NSR
4. The NetworkServiceResponse will only contain the local domain's endpoints.

## Context

The fix for #2169  (PR #2170) introduced this filter.

A potential fix that works in my setup is to wrap the localNSE filter in an proxy-nsmd-k8s env var setting:
```
diff --git a/k8s/pkg/proxyregistryserver/discovery.go b/k8s/pkg/proxyregistryserver/discovery.go
index 3072aa86..7f44e1c5 100644
--- a/k8s/pkg/proxyregistryserver/discovery.go
+++ b/k8s/pkg/proxyregistryserver/discovery.go
@@ -4,6 +4,7 @@ import (
        "context"
        "fmt"
        "os"
+       "strconv"
        "strings"

        "github.com/pkg/errors"
@@ -71,8 +72,10 @@ func (d *discoveryService) FindNetworkService(ctx context.Context, request *regi
                        return nil, dErr
                }
                for _, nsm := range response.NetworkServiceManagers {
-                       if url, urlErr := d.currentDomainNSMgrURL(ctx, d.clusterInfoService, nsm.Url); urlErr == nil && nsm.Url == url {
-                               return d.handleLocalFindCase(response, url), nil
+                       if isLocalNseFilter, _ := strconv.ParseBool(os.Getenv("PROXYNSR_LOCALNSE_FILTER")); isLocalNseFilter {
+                               if url, urlErr := d.currentDomainNSMgrURL(ctx, d.clusterInfoService, nsm.Url); urlErr == nil && nsm.Url == url {
+                                       return d.handleLocalFindCase(response, url), nil
+                               }
                        }
                        nsm.Name = fmt.Sprintf("%s@%s", nsm.Name, nsm.Url)
                        nsmURL := os.Getenv(ProxyNsmdAPIAddressEnv)
```

edwarnicke:
[...]

PR #2203, opened 2020-10-05
fix: proxy-nsmd-k8s always filters for local NSEs

denis-tingajkin:

## Motivation and Context
https://github.com/networkservicemesh/networkservicemesh/issues/2202

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/k8s/pkg/proxyregistryserver/discovery.go
+++ b/k8s/pkg/proxyregistryserver/discovery.go
@@ -72,7 +72,8 @@ func (d *discoveryService) FindNetworkService(ctx context.Context, request *regi
 		}
 		for _, nsm := range response.NetworkServiceManagers {
 			if url, urlErr := d.currentDomainNSMgrURL(ctx, d.clusterInfoService, nsm.Url); urlErr == nil && nsm.Url == url {
-				return d.handleLocalFindCase(response, url), nil
+				d.handleLocalFindCase(response, nsm, url)
+				continue
 			}
 			nsm.Name = fmt.Sprintf("%s@%s", nsm.Name, nsm.Url)
 			nsmURL := os.Getenv(ProxyNsmdAPIAddressEnv)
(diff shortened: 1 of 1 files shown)


edwarnicke:
@tiswanso Does this look like what you have in mind?

k8s/pkg/proxyregistryserver/discovery.go
```diff
@@ -70,9 +70,14 @@ func (d *discoveryService) FindNetworkService(ctx context.Context, request *regi
...
 			return nil, dErr
 		}
-		for _, nsm := range response.NetworkServiceManagers {
+		managers := make(map[string]*registry.NetworkServiceManager)
+		for key, nsm := range response.NetworkServiceManagers {
+			managers[key] = nsm
 			if url, urlErr := d.currentDomainNSMgrURL(ctx, d.clusterInfoService, nsm.Url); urlErr == nil && nsm.Url == url {
-				return d.handleLocalFindCase(response, url), nil
+				d.handleLocalFindCase(response, nsm, url)
+				delete(managers, key)
```
tiswanso, line 78: I think you could just move the add (line 75) to after the if statement and remove this delete, right?

tiswanso:
> @tiswanso Does this look like what you have in mind?

@edwarnicke  -- if I understand it correctly, it looks good.  Looks like it just localizes the network service manager entries (for local NSEs), in which case, it preserves the full list of NSEs.  The local nsmgr info is an improvement.

denis-tingajkin:
@tiswanso yes, you are correct. This PR just add possible to localize local nsmgrs/nses. This behavior is more correct then just replace all nsmgrs/nses to local if it is actually local case https://github.com/networkservicemesh/networkservicemesh/issues/2169 and it is fixing https://github.com/networkservicemesh/networkservicemesh/issues/2202

k8s/pkg/proxyregistryserver/discovery.go
denis-tingajkin: Thanks, fixed.

tiswanso:
fyi, this fix works for me in my setup.  Thanks, @denis-tingajkin !

Issue #2212, opened 2020-10-29
Admission-webhook is using the wrong container registry, org and tag for CoreDNS


LionelJouin:
## Expected Behavior
When deploying NSM with locally built and stored images, neither NSM pods nor endpoints/clients should use the official NSM repo, only the local repo which is configured in the values.yaml file.

## Current Behavior
We used locally built and stored images for the NSC.
The NSC used the following images for the init-container:
- nsm-dns-init
- nsm-init
- nsm-monitor
- coredns
The first 3 images were pulled from the local repo as expected, but coredns image was pulled from docker.io/networkservicemesh/coredns:master, ignoring the registry, org and tag defined in the deployments/nsm/charts/admission-webhook/values.yaml.

## Failure Information (for bugs)
In the _k8s/cmd/admission-webhook/patches.go_ file for all the other images the `getRepo()` and `getTag()` functions are called, but for the coredns image, the "networkservicemesh/coredns:master" is hardcoded (https://github.com/networkservicemesh/networkservicemesh/blob/master/k8s/cmd/admission-webhook/patches.go#L49).

## Steps to Reproduce
1. Build all images and store them in own repository.
2. Deploy NSM based on these images.
3. Deploy an NSC.
4. In the output of 'kubectl describe pod <NSC_pod>' look at the events and the images pulled for the init container.

## Context
Ubuntu 18.04
Docker 19.03
Kubernetes 1.19
Xcluster, 2 nodes
NSCs with initContainers
Forwarder kernel
NSM master from Oct 12

## Failure Logs
```
$ kubectl describe pod ntarget-5d65b6b9c5-68q4h
Name:         ntarget-5d65b6b9c5-68q4h
Namespace:    default
Priority:     0
Node:         vm-002/192.168.1.2
Start Time:   Wed, 28 Oct 2020 09:11:04 +0100
Labels:       networkservicemesh.io/app=ntarget
              pod-template-hash=5d65b6b9c5
Annotations:  <none>
Status:       Pending
IP:           11.0.1.8
IPs:
  IP:           11.0.1.8
  IP:           1100:0:0:1::8
Controlled By:  ReplicaSet/ntarget-5d65b6b9c5
Init Containers:
  nsm-dns-init:
    Container ID:  cri-o://7f52136cac0353f7ca587546157a5c3f8dc0660ac798ff729a68252a91973b95
    Image:         <own_repository>/nsm-dns-init:base
[...]

PR #2213, opened 2020-10-29
fix: Admission-webhook is using the wrong container registry, org and tag for CoreDNS

denis-tingajkin:

## Motivation and Context
https://github.com/networkservicemesh/networkservicemesh/issues/2212

## How Has This Been Tested?

- [ ] Tested locally
- [ ] Have not tested

## Types of changes

## Checklist:

### DIFF
--- a/k8s/cmd/admission-webhook/env/env.go
+++ b/k8s/cmd/admission-webhook/env/env.go
@@ -1,8 +1,0 @@
-package env
-
-import "github.com/networkservicemesh/networkservicemesh/utils"
-
-const (
-	//DNSSearchDomainsPatchEnv represents a list of strings. Uses for customization of DNS patch.
-	DNSSearchDomainsPatchEnv utils.EnvVar = "DNS_SEARCH_DOMAINS"
-)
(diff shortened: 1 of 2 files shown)


LionelJouin:
Using this version, I am now able to change the registry and the repository but I am still not able to change the tag.

Also, I am able to set the registry and the repository only if I put them together in the environment variable `REPO` of the admission-webhook pod.

denis-tingajkin:
@LionelJouin Note: we are using CoreDNS with the master version from a separate repo by default. So we cannot set for CoreDNS image some specific version. As a solution I see a variant to add some additional env for CoreDNS image but it looks not good for me. 
What are the problems you see with using a custom repo and master version?

denis-tingajkin:
@LionelJouin Yes, and it looks correct for me because of CoreDNS container **is not part of this repository.**

denis-tingajkin:
Here failing build because on CI we are using networkservicemeshci repository that has not CoreDNS image. So it looks like this direction is invalid.

